PR #3, opened 2015-01-13
RANGER-203: Policy implementation to support permissions & policyCondition

gautamborad:
Added the new UI layout for permission / policy condition


Hoo199212:
Different resources of a component use the same set of resource types. Is the design granularity a bit coarse

PR #4, opened 2015-11-04
RANGER-699 API for deny/allow from external tool

yonzhang:


security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -1729,4 +1731,149 @@ private RangerPolicyEngine getPolicyEngine(String serviceName) {
 
 		return ret;
 	}
+	
+	protected RangerPolicy policyFlip(RangerPolicy policy) {
```
mneethiraj, line 1735: Consider the following impl:
public RangerPolicy applyPolicy(RangerPolicy policy) {
  RangerPolicy ret = null;

  RangerPolicyResourceSignature resourceSignture = new RangerPolicyResourceSignature(policy);
  List<RangerPolicy>                     existingPolicies    = svcStore.getPoliciesByResourceSignature(policy.getService(), resourceSignature.getSignature(), true);

  if(CollectionUtils.isEmpty(existingPolicies)) {
     ret = svcStore.createPolicy(policy);
  } else {
    RangerPolicy existingPolicy = existingPolicies.get(0); // there should be only one policy for the given resources
    // TODO: update existingPolicy, per the contents of policy. Refer to the pseudo code in RANGER-699
   ret = svcStore.updatePolicy(existingPolicy);
  }

  return ret;
}

PR #16, opened 2017-05-24
Extra debug logging in UserSync Group Search

rvesse:
When debugging issues with group search it can be hard to understand why
groups are not being found because the specific LDAP search filters used
i.e. The filter templates with placeholders filled are not logged.
This adds extra debug logging so when log level is DEBUG the exact LDAP
search filter to be used will be logged.  Users can then test this
filter directly against their LDAP server to determine why it is
incorrect and then adjust their configuration appropriately a


aloklal99:
@rvesse Thanks for submitting the patch and for the detailed description.
- Please use the [apache review board](https://reviews.apache.org/) to submit your code changes for review and commit.
- Also [create a jira](https://issues.apache.org/jira/browse/RANGER) to go along with the review and mention it in the review board review.

rvesse:
https://reviews.apache.org/r/59801/
https://issues.apache.org/jira/browse/RANGER-1634

FYI Pull requests would be much easier to deal with than Review Board, I work on other Apache projects where we use PRs exclusively

PR #35, opened 2019-04-05
RANGER-2397

junseung-hwang:
https://jira.apache.org/jira/browse/RANGER-2397


aloklal99:
@junseung-hwang,

- Please use the [apache review board](https://reviews.apache.org/) to submit your code changes for review and commit.

Thanks for submitting the patch!

mneethiraj:
@aloklal99  - after recent Apache-wide changes, accessing review board requires an Apache ID - which is available only for committers. Git pull seems to be the best alternate.

Until we find a way to merge pull requests, I would suggest committers to download .patch file from pull requests (by simply adding .patch to the URL - like https://github.com/apache/ranger/pull/35.patch).

aloklal99:
@mneethiraj, My bad, thanks for clarifying.  Let me also cross post this correction on the prior patch request.

kumaab:
Commit: https://github.com/apache/ranger/commit/18d0f1a

PR #39, opened 2019-08-01
Decrypt the password coming from x_service_config_map

JituS:
When the connection is established for presto service, the password is stored in x_service_config_map MySQL table.
Autocomplete will still not work if the password is not decrypted.
Added the step to decrypt the password.

### DIFF
--- a/plugin-presto/src/main/java/org/apache/ranger/services/presto/client/PrestoClient.java
+++ b/plugin-presto/src/main/java/org/apache/ranger/services/presto/client/PrestoClient.java
@@ -79,8 +79,13 @@ public class PrestoClient extends BaseClient implements Closeable {
     String url = prop.getProperty("jdbc.url");
 
     Properties prestoProperties = new Properties();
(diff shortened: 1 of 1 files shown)


tooptoop4:
@JituS what autocomplete does it fix? I tried this patch but autocomplete of catalog/schema/table names still did not work when adding new policies
@bolkedebruin any idea?

bolkedebruin:
Autocomplete just works here without this patch so I suspect misconfiguration

JituS:
@bolkedebruin Have you tried it with presto? I faced the issue related to auto complete while configuring the policies for presto tables. Adding that changes actually fixed the issue.
I don't have a stack trace now as it was a long way back. But you can refer `HiveClient.java:716` class where we are decrypting the password before creating the connection.

bolkedebruin:
@JituS I wrote the support for Presto in Ranger so yes I did. But maybe I am missing the special case here?

PR #42, opened 2019-11-05
Adding support for ElasticSearch and Hive 2.3.x compatibility

acharneski:
…h components


agents-audit/src/main/java/org/apache/ranger/audit/destination/ElasticSearchAuditDestination.java
```diff
@@ -0,0 +1,188 @@
...
+    public boolean log(Collection<AuditEventBase> events) {
+        boolean ret = false;
+		try {
+			logStatusIfRequired();
+			addTotalCount(events.size());
+
+            if (client == null) {
+				connect();
+				if (client == null) {
+					// Solr is still not initialized. So need return error
```
boscodurai: We need to replace with Elastic Search

agents-audit/src/main/java/org/apache/ranger/audit/destination/ElasticSearchAuditDestination.java
```diff
@@ -0,0 +1,188 @@
...
+                    try {
+                        me = client = new RestHighLevelClient(
+                                RestClient.builder(
+                                        MiscUtil.toArray(urls, ",").stream()
+                                                .map(x -> new HttpHost(x, port, protocol))
+                                                .<HttpHost>toArray(i -> new HttpHost[i])
+                                ));
+                        ;
+                    } catch (Throwable t) {
+                        LOG.fatal("Can't connect to Solr server. urls=" + urls, t);
```
boscodurai: Based on the logic, every time log is called and if the connection is not established, connect() is called. And if Elastic Search is not available, then we will continuously keep logging fatal. Since plugins run within process of the service, it will be bombarded with fatal logs. I feel, we should limit the number of logging. Probably keeping track of how many times we log per interval. E.g. we can have static variable which says last logged time and have a configurable what should be interval to log. If the interval > (currentTime - lastLogTime ), then log... Just a suggestion

pom.xml
```diff
@@ -995,6 +995,11 @@
                 <artifactId>enunciate-core-annotations</artifactId>
                 <version>${enunciate.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.glassfish</groupId>
+                <artifactId>javax.el</artifactId>
+                <version>3.0.1-b11</version>
```
boscodurai: We should externalize the version

security-admin/src/main/java/org/apache/ranger/elasticsearch/ElasticSearchAccessAuditsService.java
```diff
@@ -0,0 +1,390 @@
...
+					"Error connecting to search engine",
+					MessageEnums.ERROR_SYSTEM);
+		}
+		List<VXAccessAudit> xAccessAuditList = new ArrayList<VXAccessAudit>();
+		Map<String, Object> paramList = searchCriteria.getParamList();
+		updateUserExclusion(paramList);
+
+		SearchResponse response;
+		try {
+			response = elasticSearchUtil.searchResources(searchCriteria, searchFields, sortFields, client);
```
boscodurai: Seems except  for this, everything seems to be common between Solr and Elastic. Should consider abstracting a class with common methods? E.g. updateUserExclusion(), getExcludeUsersList(), getServiceUserList(), etc. This will be easy to maintain going forward.

security-admin/src/main/java/org/apache/ranger/elasticsearch/ElasticSearchMgr.java
```diff
@@ -0,0 +1,90 @@
...
+					try {
+						me = client = new RestHighLevelClient(
+								RestClient.builder(
+										MiscUtil.toArray(urls, ",").stream()
+												.map(x -> new HttpHost(x, port, protocol))
+												.<HttpHost>toArray(i -> new HttpHost[i])
+								));
+						;
+					} catch (Throwable t) {
+						logger.fatal("Can't connect to Solr server. urls=" + urls, t);
```
boscodurai: Replace Solr with Elastic search

security-admin/src/main/java/org/apache/ranger/elasticsearch/ElasticSearchMgr.java
```diff
@@ -0,0 +1,90 @@
...
+		return value;
+	}
+
+	synchronized void connect() {
+		RestHighLevelClient me = client;
+		if (me == null) {
+			synchronized (ElasticSearchAuditDestination.class) {
+				me = client;
+				if (client == null) {
+					String urls = PropertiesUtil.getProperty("ranger.audit.solr.url");
```
boscodurai: Should this be "elasticsearch"?

acharneski:
Apologies about the premature PR. Please close this if you wish, I'll reopen when complete. I had intended to make a PR within my own fork to track the change before release. I will review the comments, though, thank you for taking a look!

ranger-util/src/test/LICENSE.txt
```diff
@@ -0,0 +1,323 @@
+                                 Apache License
```
acharneski, line 1: Maven complained that this directory does not exist. This file can enforce that while not raising an alert with the rats build plugin.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerSafenetKeySecure.java
```diff
@@ -69,7 +70,7 @@ public RangerSafenetKeySecure(Configuration conf) throws Exception {
 			int javaVersion = getJavaVersion();
 			/*Minimum java requirement for Ranger KMS is Java 8 and Maximum java supported by Ranger KMS is Java 11*/
 			if(javaVersion == 8){
-				provider = new sun.security.pkcs11.SunPKCS11(pkcs11CfgFilePath);
+                provider = sun.security.pkcs11.SunPKCS11.class.getConstructor(String.class).newInstance(pkcs11CfgFilePath);
```
acharneski, line 72: Using reflection here prevents compilation errors when using Java 11

pom.xml
```diff
@@ -69,7 +69,7 @@
         </mailingList>
     </mailingLists>
     <properties>
-        <maven.version.required>3.6.2</maven.version.required>
+        <maven.version.required>3.6.1</maven.version.required>
```
acharneski, line 72: Currently the latest available version on Mac/Brew. Seems to work fine.

acharneski:
This PR is ready for review and feedback. Some of the items are merely to get the build to work in my environments. 

My main concern is that I have been unable to test this functionality in an integrated environment - Currently our organization is only using the Hive plugin and my analysis of the code suggests that only "grant" and "revoke" permission operations (in RangerBasePlugin) make use of RangerAuthContextListener, which seems to end up in the audit log. Our test environment does show such events, but I am fairly certain this does not actually use the Audit DB (with the new Elasticsearch support). I have prepared some test code that uses a local Elasticsearch to read and write events (ElasticSearchAccessAuditsServiceTest) but have been unable to test actual audit events in our integration test environment.

Do you have any suggestions on how to properly test this? Thank you very much for your guidance and feedback!

acharneski:
Replaced with https://reviews.apache.org/r/71921/

PR #44, opened 2019-11-13
LDAP user sync bug

xihuanbanku:
Change date format to 24 hours.  For that using 12 hours format will missing records in LDAP which create between 13:00-24:00.


tooptoop4:
is fullsync affected or just deltasync?

xihuanbanku:
> is fullsync affected or just deltasync?

just deltasync

PR #49, opened 2019-12-30
Adding support for prestodb plugin on ranger

vasveena:
Ranger has extended support for PrestoSQL via https://issues.apache.org/jira/browse/RANGER-2395. Plugin: https://github.com/apache/ranger/tree/master/plugin-presto.

Created a plugin to extend support to prestodb as well. The new plugin will exist along with PrestoSQL and the user can choose plugin based on presto distribution on their system.


qwalker:
I am expecting prestodb support too, hop it's merged soon.

PR #54, opened 2020-03-05
Update CustomSSLSocketFactory.java

YYTVicky:


YYTVicky:
Hi, 
   we found the TLS is not secure for build connection (https://www.packetlabs.net/tls-1-1-no-longer-secure/), would you mind changing it to TLSv1.3 for security concern?

PR #55, opened 2020-03-05
Ranger REST API missing info when GET user by name

dargooo:
https://issues.apache.org/jira/browse/RANGER-2737


mneethiraj:
@dargooo  - thank you for the fix. Merged this fix in master and ranger-2.4 branches.

PR #56, opened 2020-03-06
Update PasswordUtils.java

YYTVicky:


YYTVicky:
Hi,
   the for PBEParameterSpec, the iteration count should be large then 1000 for security reasons, so we recommend you use a number larger than 1000 when setting DEFAULT_ITERATION_COUNT = 17

Thanks and hope to get any feedback from your side!

PR #57, opened 2020-03-07
[RANGER-2747] Ranger integration with presto works on "Release 317" but fails on versions later

ebyhr:
Update Presto version to 318 from 310

JIRA https://issues.apache.org/jira/browse/RANGER-2747


brucemen711:
Hi @ebyhr, will this path work with presto 320 or 325 ?

PR #61, opened 2020-04-10
RANGER-2789: GET API service/xusers/users turns very slow 

rickyma:
RANGER-2789: GET API service/xusers/users turns very slow when there are more than 1000 users.
https://issues.apache.org/jira/browse/RANGER-2789

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/db/XXGroupUserDao.java
+++ b/security-admin/src/main/java/org/apache/ranger/db/XXGroupUserDao.java
@@ -66,6 +66,17 @@ public class XXGroupUserDao extends BaseDao<XXGroupUser> {
 		return null;
 	}
 
+	public List<XXGroupUser> findAllXXGroupUser() {
+		try {
+			return getEntityManager()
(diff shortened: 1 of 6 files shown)


rickyma:
Copied from the issue link: 
We have more than 1000 users in production environment.

When calling API 'service/xusers/users?pageSize=1000&startIndex=0' using HTTP GET method, Ranger Admin takes 10+ minutes to response.

I've already uploaded the patch. By applying this patch, the response time has decreased from 10+ minutes to around 30 seconds in our production environment.

The problem of this issue is that the original code will query the database for every one of the 1000+ users:
https://github.com/apache/ranger/blob/master/security-admin/src/main/java/org/apache/ranger/service/XUserServiceBase.java#L84

The main ideas of the uploaded patch are:
1.Find all users in the database at one time.
2.Use HashMap as cache to store all the database info in memory.
3.Just use the cache during any follow-up db operation.

rickyma:
Patch committed : https://github.com/apache/ranger/commit/f45054d1b933b5e65cb19ace014d3b2cb0d78132

PR #64, opened 2020-04-30
keystore load plaintext password

YYTVicky:


Mahir92:
Hi! To get a better understanding of what @YYTVicky  referred to, please take a look at the code snippet in the following url - https://pastebin.ubuntu.com/p/sWBW4zfSzR/ .

This is just an example code that shows how to load the password, instead of using plaintext.

PR #66, opened 2020-06-11
RANGER-2856: A policy should be deleted if it has no policyItems

rickyma:
Condition: A policy contains only one policyItem and the policyItem only sets one user and no groups.

Action: Delete the user in the policyItem using API: 'http://ip:6080/service/xusers/users/{id}?forceDelete=true'

Result: The policy still exists, but it has no policyItems. A policy with no policyItems is completely meaningless and it should be deleted.


rickyma:
Patch committed：https://github.com/apache/ranger/commit/400beaa7604e4c83d051dc6e83b8bc109e5d8c53

PR #67, opened 2020-06-29
RANGER-2885: Add missing PermType Java codes for new versions of Kafka

rickyma:
There are codes missing for:

[RANGER-2117](https://issues.apache.org/jira/browse/RANGER-2117)
[RANGER-2692](https://issues.apache.org/jira/browse/RANGER-2692)

When calling 'http://10.1.0.44:6080/service/public/api/policy?repositoryName=tdw_kafka\&repositoryType=kafka\&isEnabled=false', it will return 'Unknown' value in permMapList.permList field, which is not appropriate.


rickyma:
Patch committed : https://github.com/apache/ranger/commit/a15ec3c1028b01b292e257a1a50d83fcd8d91149

PR #74, opened 2020-09-10
RANGER-5117: Fix ssl config param in installprop2xml.properties

mjshariati98:
According to this incorrect mapping: if set 'AUTH_SSL_ENABLED' to false in 'install.properties', then setup.sh will produce incorrect 'ranger-ugsync-site.xml' config that causes usersync process not working because actually 'ranger.usersync.enabled' set to false.


pritishsahoo (commented):
It's the same issue with 2.0.0 version also.

pritishsahoo (approved):

pritishsahoo (approved):
Usersync process works with this change. ranger.usersync.enabled property is correctly set to true with this change.

vijayinani:
The issue still persists. I went crazy as my Ranger LDAP integration was not working. Finally solved after ghosting the key value pair.

sbmrvv:
The issue still persists in Ranger 2.4.0.

@vijayinani I went crazy as well 

Finally solved with `ranger.usersync.enabled` to set `true`

vijayinani:
> The issue still persists in Ranger 2.4.0.
> 
> @vijayinani I went crazy as well
> 
> Finally solved with `ranger.usersync.enabled` to set `true`

Yes, I also did the same. This should be fixed. Give me access and I will fix it! :)

PR #79, opened 2020-10-07
RANGER-3027: Improve response time for GET API service/xusers/users

rickyma:


v-9980:
This is useful for me when upgrading from version 2.1.0 to 2.3.0(RANGER-2789 has been reverted). 
In the Ranger plugin, only policies and roles are synchronized, and there is no mapping for users and groups. In order to enable group authorization in the policies, and to ensure that only user-specific policies are passed in the plugin, we maintain a cache of user-to-group mapping in the plugin. The synchronization frequency of this cache is consistent with the policy synchronization time. We use the service/xusers/users API to fetch the user-to-group relationship. When there are fewer users, Ranger-admin runs smoothly. However, when there are a large number of users(2k), Ranger-admin experiences a significant number of blocked threads.

thanks, RickyMa

image
image

rickyma:
I'm glad that it can be of help to you, although it is still waiting to be merged. 
It's been 3 years since I first opened this merge request.   :(

@mneethiraj @kumaab Please have a look at this. Thanks.

PR #81, opened 2020-10-16
Release ranger 2.1.0 cdh 6.3.1 hylink

gm19900510:


zhangxinjian123:
The cdh6.1.1 integrated ranger 2.1.0 also changes the relevant code according to the instructions. Hive can perform permission control, but the permission control of hdfs does not take effect. What is the matter?

PR #82, opened 2020-10-24
RANGER-3058: [ranger-hive] create table fails when ViewDFS(client side HDFS mounting fs) mount points are targeting to Ozone/S3 FS.

umamaheswararao:
https://issues.apache.org/jira/browse/RANGER-3058


kumaab:
Submitted via: https://reviews.apache.org/r/72989

PR #84, opened 2020-12-21
Fix typo in RangerKafkaAuditHandler logs

jrevillard:


kumaab:
Thanks @jrevillard, the change is present via this commit: https://github.com/apache/ranger/commit/da68108

PR #89, opened 2021-01-28
Spaced are authorized in LDAP value strings setup.sh script doesnt account for spaces

dprophet:
Example in install.properties file:
xa_ldap_userDNpattern=OU=Enabled Accounts,DC=mydc,DC=mycompany,DC=com

The space in the LDAP string causes 3rd and 4th arguments to the updatePropertyToFilePy method in the startup.sh to be wrong. This then breaks the generated ews/webapp/WEB-INF/classes/conf/ranger-admin-site.xml file.


dprophet:
Here is the JIRA issue link
[https://issues.apache.org/jira/browse/RANGER-3166](url)

PR #90, opened 2021-02-27
RANGER-2962: {USER} Transformation in Hive Plugin's Row level Filters

mudit-97:


mneethiraj:
Suggestion in RANGER-2962 addresses the usecase, hence this patch is not necessary. Closing this PR.

PR #93, opened 2021-03-29
FIX: Update Docker Compose Ranger Depends On

gabdants:
Since the docker compose version is at '3', when the build starts, it cannot build the images because of the depends_on `condition: service_started`.


mneethiraj:
Please reopen if the issue exists in the lastest release, 2.8.0.

PR #94, opened 2021-03-31
RANGER-3227: use GenericType for List<Object>

zhoukangcn:
see: https://issues.apache.org/jira/browse/RANGER-3227


kumaab:
Present as part of https://github.com/apache/ranger/commit/e01d39c954fa196c321f7f91ef15a518b34bde49

PR #97, opened 2021-04-15
RANGER-3244 Update test related dependencies and Maven plugins

martin-g:
Extract Maven properties for all Maven plugin versions.


martin-g:
https://reviews.apache.org/r/73282/
https://issues.apache.org/jira/browse/RANGER-3244

PR #96, opened 2021-04-15
RANGER-3243 Fix build due to hive exec and JDK 8 specific class

martin-g:
- org.apache.ranger.authorization.presto.authorizer.RangerSystemAccessControlFactory was not able to import com.google.common.base.Throwables.throwIfUnchecked because it was coming from an old copy of Guava's Throwables shaded in hive-exec.
By replacing hive-exec with orc-core in agents-audit module all depending modules use their preferred version of Guava (26.0-jre)
-  Use reflection to load SunPKCS11 class because it is not available in JDK 9+ and breaks the build on newer JDK versions

PR #98, opened 2021-04-16
RANGER-3245 Install JDK from Ubuntu repository. Add job for ARM64 

martin-g:
https://issues.apache.org/jira/browse/RANGER-3245

Oracle 8 is not available at TravisCI anymore. Use OpenJDK 1.8 and 11 instead.

**Note**: this PR depends on https://github.com/apache/ranger/pull/96. Without it the build on JDK 11 will still fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,18 +15,21 @@
 
 
 sudo: false
+dist: bionic
 language: java
+
 cache:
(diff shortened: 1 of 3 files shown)


martin-g:
Any feedback on the suggested changes ?

PR #99, opened 2021-04-27
RANGER-3255 : Lost nessary pre-check for Empty input argument

Xinshiyou:
exception fixed : https://issues.apache.org/jira/browse/RANGER-3255


mneethiraj:
This is addressed in PR #779.

PR #104, opened 2021-06-10
RANGER-3313 Remove org.apache.ranger.audit.entity.AuthzAuditEventDbOb…

chia7712:
RANGER-3184 had removed support for audit to RDBMS. The org.apache.ranger.audit.entity.AuthzAuditEventDbObj in the class element can be removed also.


mneethiraj:
@chia7712 - thank you for the patch. Instead of updating persistence.xml, I suggest removing the file since it is no more referenced after RANGER-3184 (which removed support for writing audit logs to RDBMS).

PR #105, opened 2021-06-10
RANGER-3315 Fix travis CI - leverage docker to run CI

chia7712:
issue: https://issues.apache.org/jira/browse/RANGER-3315

We ought to reuse `build_ranger_using_docker.sh` to simplify ranger CI (all required dependencies can be included by container). Also, our travis CI config is broken because of following reasons.
1. Currently, only JDK 8 is supported (https://issues.apache.org/jira/browse/RANGER-3243)
2. the oracle JDK is NOT supported (https://github.com/sormuras/bach/issues/56)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,33 +13,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 sudo: false
+arch: amd64
+os: linux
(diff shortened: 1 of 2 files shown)


chia7712:
This PR includes #103 to make build_ranger_using_docker.sh works with travis

PR #114, opened 2021-08-24
RANGER-3231 Ranger should use kafka Authorizer from KIP-504

chia7712:
issue: https://issues.apache.org/jira/browse/RANGER-3231


OneCricketeer:
Link #133

PR #115, opened 2021-09-01
Add Trino v360  support

utk-spartan:


huanghp004:
hi everyone,I didn't find Trino in version 2.1.0

ThoSap:
Hi, will this be part of the upcoming 2.2.0 release?

rameeshm:
Hi ThoSap, I see that this is still in review stage and won't be possible to include in the release.

ThoSap:
Looking at the past release cycle, does that mean we have to wait another year for Trino support?

butaozhang:
+1 Look forward to merging ranger-rino plugin feat.

rgelsi:
+1 Ranger is the only thing that currently prevents us from migrating to Trino.

rameeshm:
@ThoSap @rgelsi We shall plan doing a minor release may be in 3 months which can include Ranger-3182.
Please comment on the jira and lets work on it to bring it to closure.

PR #119, opened 2021-10-07
RANGER-3469: Fix Off-By-One Error in XUser Syncing

belugabehr:


mneethiraj:
This issue was addressed in [RANGER-3920](https://issues.apache.org/jira/browse/RANGER-3920).

PR #131, opened 2022-01-31
setup error if optional environment variable RANGER_ADMIN_LOG4J_CONF_…

jonashartwig:
dict.get returns None if the key not exists, where as dict[key] fails the application.


mneethiraj:
The setup script expects log releated configurations, , RANGER_ADMIN_LOG_DIR and RANGER_ADMIN_LOGBACK_CONF_FILE, to be present in install.properties - in multiple code blocks. If these are to be optional, all references must be updated. If this fix is critical, can you please open a new PR with updates to all references?

PR #135, opened 2022-03-06
RANGER-3010. Change the function of addUsersAndGroups & removeUsersAn…

ted12138:
…dGroups REST API

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/PublicAPIsv2.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/PublicAPIsv2.java
@@ -745,18 +745,20 @@ public class PublicAPIsv2 {
     	This API is used to add users and groups with/without GRANT privileges to this Role. It follows add-or-update semantics
  	 */
 	@PUT
-	@Path("/api/roles/{id}/addUsersAndGroups")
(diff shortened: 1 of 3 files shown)


security-admin/src/main/java/org/apache/ranger/rest/PublicAPIsv2.java
```diff
@@ -745,18 +745,20 @@ public RangerRole getRole(@PathParam("id") Long id, @Context HttpServletRequest
     	This API is used to add users and groups with/without GRANT privileges to this Role. It follows add-or-update semantics
  	 */
 	@PUT
-	@Path("/api/roles/{id}/addUsersAndGroups")
-	public RangerRole addUsersAndGroups(@PathParam("id") Long roleId, List<String> users, List<String> groups, Boolean isAdmin, @Context HttpServletRequest request) {
-		return roleREST.addUsersAndGroups(roleId, users, groups, isAdmin);
```
mneethiraj, line 750: @ted12138  - thank you for the patch. The changes look good.

One suggestion: to be consistent with use of` {id}` and `{name}` in APIs, I would suggest to retain `{id}` in this API and introduce another API for using `{name}` - like: `/api/roles/name/{name}/addUserAndGroups`.

ted12138: OK, I got your point, I will solve it later.

firebook:
So when will you merge this request? @mneethiraj @ted12138

PR #136, opened 2022-03-18
RANGER-3671: Fix chainedPlugin construct

CanGuan:
Details:
https://issues.apache.org/jira/browse/RANGER-3671

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -1081,10 +1081,17 @@ public class RangerBasePlugin {
 				continue;
 			}
 
+			String serviceType = pluginConfig.get(chainedServicePropPrefix + "." + chainedService + ".service.type");
+			if (StringUtils.isBlank(serviceType)) {
+				LOG.error("Ignoring chained service " + chainedService + ": not service type specified");
+
+				continue;
+			}
+
 			try {
 				@SuppressWarnings("unchecked")
 				Class<RangerChainedPlugin> pluginClass   = (Class<RangerChainedPlugin>) Class.forName(className);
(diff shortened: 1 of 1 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1084,7 +1084,7 @@ private RangerAdminClient getAdminClient() throws Exception {
 			try {
 				@SuppressWarnings("unchecked")
 				Class<RangerChainedPlugin> pluginClass   = (Class<RangerChainedPlugin>) Class.forName(className);
-				RangerChainedPlugin        chainedPlugin = pluginClass.getConstructor(RangerBasePlugin.class, String.class).newInstance(this, chainedService);
+				RangerChainedPlugin        chainedPlugin = pluginClass.getConstructor(RangerBasePlugin.class, String.class, String.class).newInstance(this, pluginConfig.getServiceType(), chainedService);
```
mneethiraj: serviceType should that of chainedService; pluginConfig.getServiceType() would be the serviceType of current plugin.

mneethiraj: @CanGuan  - the expectation is that the class that implements RangerChainedPlugin has a constructor that takes 2 arguments (RangerBasePlugin parent, String serviceName), and that constructor calls RangerBasePlugin constructor with 3 arguments (RangerBasePlugin parent, String serviceType, String serviceName).

CanGuan: @mneethiraj Please have a look. The service type for chained plugin is loaded from parent plugin config.

kumaab:
The jira has been resolved, closing the PR.

PR #147, opened 2022-05-18
RANGER-3763:The max limit of the requested entities is not configurable in tagsync

Hoo199212:
This solves the configuration problem of the max limit of the requested entities


tagsync/src/main/java/org/apache/ranger/tagsync/process/TagSyncConfig.java
```diff
@@ -120,7 +121,7 @@ public class TagSyncConfig extends Configuration {
 	private static final int     DEFAULT_TAGSYNC_SINK_MAX_BATCH_SIZE = 1;
 	private static final String  TAGSYNC_SINK_MAX_BATCH_SIZE_PROP    = "ranger.tagsync.dest.ranger.max.batch.size";
 
-
+	private static final String TAGSYNC_ATLAS_REST_SOURCE_REQUESTED_ENTITIES_LIMIT_MAX  = "ranger.tagsync.source.atlasrest.requested.entities.limit.max";
```
mneethiraj, line 124: I suggest to rename the property to: ranger.tagsync.source.atlasrest.entities.batch.size

mneethiraj (approved):

tagsync/src/main/java/org/apache/ranger/tagsync/process/TagSyncConfig.java
AstraHoo: Thanks for your help

mneethiraj:
Fix is now committed in master branch. Thanks @Hoo199212  for the patch.

PR #154, opened 2022-06-22
RANGER-3790: Ranger tagsync module should not depend on kafka server

patrik-marton:
This commit removes the unused kafka core dependency from the assembly xml of the tagsync module, so it will not be added to the distribution.


mneethiraj:
This PR is already merged in master branch - https://github.com/apache/ranger/commit/76349ecaea86f36bd5908151f2ec50943b67ab21.

PR #162, opened 2022-07-19
Ranger 3833

datahu0702:
Fix the error in the default schema of audit log, which makes it impossible to query by 'application'.


mneethiraj (approved):

security-admin/contrib/solr_for_audit_setup/conf/managed-schema
```diff
@@ -61,7 +61,7 @@
   <field name="_version_" type="long" indexed="false" stored="true"/>
   <field name="access" type="key_lower_case" multiValued="false"/>
   <field name="action" type="key_lower_case" multiValued="false"/>
-  <field name="agent" type="key_lower_case" multiValued="false"/>
+  <field name="agentId" type="key_lower_case" multiValued="false"/>
```
mneethiraj: I see that access audit logs sent to Elasticsearch, in `ElasticSearchAuditDestination.java`, do have field name as `agent`; so this chance should not be needed.

mneethiraj:
If the issue still exists, please reopen with details. As noted in my comment earlier, the field name `agent` is the correct one; hence this fix seems not necessary.

PR #164, opened 2022-08-01
RANGER-3751: Fixing Elasticsearch/Solr Audit log Datetime format

tarekabouzeid:
This PR is for fixing date for audits appearing as 1970 when elasticsearch is used as audits store.


ognjen-it:
It's fixed so this PR could be closed. #314

PR #168, opened 2022-08-12
[Trino Plugin] fix the test failure

alvinwo:
I got test failure when test the trino plugin:
```
Caused by: java.lang.IllegalStateException: Detected both log4j-over-slf4j.jar AND bound slf4j-log4j12.jar on the class path, preempting StackOverflowError. See also http://www.slf4j.org/codes.html#log4jDelegationLoop for more details.
	at org.slf4j.impl.Log4jLoggerFactory.<clinit>(Log4jLoggerFactory.java:54)
	... 28 more
```
Screen Shot 2022-08-12 at 18 00 35

The failure is that the two jars `log4j-over-slf4j.jar` AND `slf4j-log4j12.jar` are detected in the classpath: https://stackoverflow.com/questions/20117720/detected-both-log4j-over-slf4j-jar-and-slf4j-log4j12-jar-on-the-class-path-pree
Screen Shot 2022-08-12 at 17 51 05

So I removed the test dependency `log4j-over-slf4j` to run the test.

### DIFF
--- a/plugin-trino/pom.xml
+++ b/plugin-trino/pom.xml
@@ -67,12 +67,6 @@
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>log4j-over-slf4j</artifactId>
-            <version>${slf4j.version}</version>
-	    <scope>test</scope>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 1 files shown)


kumaab:
Trino plugin implementation has been moved to the trino repo.

PR #173, opened 2022-08-29
Installation Fails setup.sh : Newer MySQL requires to set the flag useSSL false when not using SSL

justusrk:
Running setup.sh fails because of not explicitly setting useSSL=false. 

This is done properly in db_setup.py but not in dba_setup.py.  This patch fixes the problem and runs the setup successfully for newer MySQL versions.


ramackri:
As part of this PR:https://github.com/apache/ranger/pull/173
This issue got addressed and it seems needs to be fixed for newer MySQL versions.
@mneethiraj

ramackri (approved):

PR #184, opened 2022-11-16
Remove os.path.join causing incorrect windows path

bdsoha:
@mneethiraj **The exact same issue occurred with `apache-atlas` and was already merged https://github.com/apache/atlas/pull/203**

When running on *Windows* the `os.path.join` using the operating systems `os.sep=\` which causing the full url to be incorrect.

Using `urllib.parse` might be a little bit slower, but is less error-prone.

### DIFF
--- a/intg/src/main/python/apache_ranger/client/ranger_client.py
+++ b/intg/src/main/python/apache_ranger/client/ranger_client.py
@@ -19,7 +19,6 @@
 
 import json
 import logging
-import os
 from apache_ranger.exceptions                 import RangerServiceException
 from apache_ranger.model.ranger_base          import RangerBase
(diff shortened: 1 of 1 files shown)


bdsoha:
@mneethiraj Any change we can get this merged? Thanks in advance.

mneethiraj:
@bdsoha - thanks for the fix. I filed [RANGER-3969](https://issues.apache.org/jira/browse/RANGER-3969) to track this issue; will merge this fix shortly.

bdsoha:
@mneethiraj Together with a package release on PyPi?

mneethiraj:
@bdsoha -  Apache Ranger Python client 0.0.7 was released few minutes back with this fix.

PR #186, opened 2022-11-21
RANGER-3968:The plugin cannot download the latest policy

Hoo199212:
Add transaction timeout settings

### DIFF
--- a/security-admin/src/main/resources/conf.dist/ranger-admin-default-site.xml
+++ b/security-admin/src/main/resources/conf.dist/ranger-admin-default-site.xml
@@ -681,4 +681,8 @@
 		<name>ranger.allow.kerberos.auth.login.browser</name>
 		<value>false</value>
 	</property>
+	<property>
+		<name>ranger.policy.retriever.transaction.timeout</name>
+		<value>30</value>
+	</property>
 </configuration>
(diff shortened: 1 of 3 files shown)


Hoo199212:
@bhavikpatel9977 please review, thank you so much!

mneethiraj:
@Hoo199212  - what happens in environments having large number of policies, where loading can take more than 30 seconds? Wouldn't this patch result in loading to fail every time?

Hoo199212:
> 
In our production environment, the cache update thread has been stuck, the plugin can not get the latest permission information for a long time (perhaps more than 1 day)，But once the server is restarted, it will be normal immediately. The timeout allows the thread to release the lock. The transaction time can be set, if the data volume is large and takes a long time to load, we can set a reasonable value

security-admin/src/main/resources/conf.dist/ranger-admin-default-site.xml
```diff
@@ -681,4 +681,8 @@
 		<name>ranger.allow.kerberos.auth.login.browser</name>
 		<value>false</value>
 	</property>
+	<property>
+		<name>ranger.policy.retriever.transaction.timeout</name>
```
pan3793, line 685: What's the time unit? Could you please follow the existing configuration key naming style?
`<name>ranger.resource.lookup.timeout.value.in.ms</name>`

PR #191, opened 2022-11-25
RANGER-3985: use table creation rule for trino

jonashartwig:
The ranger rules to create tables in Trino currently check schema level to create.

If this is set, anyone can create any table/view. There is no way to limit the naming of tables.

However e.g. drop, alter rights are granted on table level. So user might create any table, but not remove them.

To allow a more strict implementation view/table creation should verify table name as well.

In that case the previous behaviour can be created by adding a rule to allow create on catalog/schema/*.


ognjen-it:
Hi,

what's going on with this PR? Is there any blocker to merge it?

kumaab:
Closing the PR, Trino plugin implementation has moved to the Trino repo.

PR #198, opened 2022-12-16
update lucene version to fix audit log throw no suchfield LUCENE_8_5_…

deadlove:
…1 issue


deadlove:
using elasticsearch as audit store, when go to audit page  will throw ElasticsearchException class init error, because of the lucene version problem, after upgrade the lucene version, this issue resloved. But lucene-spatial has no higher version than 8.4.1 so we keep this as before.

esselius:
I just hit this as well

floyola:
Is the issue solved with just changing the lucene version?

PR #207, opened 2023-01-12
RANGER-4029 Ranger cannot build with HBase 2.5.x versions because pre…

chrajeshbabu:
…Balance coprocessor hook syntax changed in HBASE-26147


Jetiaime:
Has it been completely modified? It seems that it has only been partially modified.

mneethiraj:
Ranger currently builds with HBase 2.6.0. If this is still an issue to be addressed, please reopen and rebase.

PR #210, opened 2023-02-07
RANGER-4071: Support for LDAP/AD usernames and group names with special chars

kumaab:

Ranger usersync regards the forward slash ('/') as a separator character in the regex so it ignores it if you actually want to use it in your groupname. e.g. RRR/group_name_suffix (/ can be potentially used as a domain separator in AD groups)

Using a different character like + in the regex works to get something like RRR+group_name_suffix but not RRR/group_name_suffix.

## How was this patch tested?

Added unit tests for the new feature.
mvn clean package on ugsync module.

Tested the changes on a cluster with these additional usersync configs with LDAP sync source:

name: ranger.usersync.mapping.regex.separator
value: #

name: ranger.usersync.mapping.username.regex
value: s#^(.*)#DE/$1#g

### DIFF
--- a/ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
+++ b/ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
@@ -1335,16 +1335,18 @@ public class UserGroupSyncConfig  {
 		return isUserSyncNameValidationEnabled;
 	}
 
-	public String getRegexSeparator() throws RuntimeException {
+	public String getRegexSeparator() {
+		String ret = DEFAULT_MAPPING_SEPARATOR;
 		String val = prop.getProperty(SYNC_MAPPING_SEPARATOR);
 		if(StringUtils.isNotEmpty(val)) {
-			if (val.length() == 1)
-				return val;
-			else{
-				throw new RuntimeException("More than one character found in RegEx Separator");
+			if (val.length() == 1) {
+				ret = val;
+			} else {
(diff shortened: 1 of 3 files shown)


ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
```diff
@@ -1332,6 +1335,18 @@ public boolean isUserSyncNameValidationEnabled() {
...
+	public String getRegexSeparator() throws RuntimeException {
+		String val = prop.getProperty(SYNC_MAPPING_SEPARATOR);
+		if(StringUtils.isNotEmpty(val)) {
+			if (val.length() == 1)
+				return val;
+			else{
+				throw new RuntimeException("More than one character found in RegEx Separator");
+			}
+		}
+		return DEFAULT_MAPPING_SEPARATOR;
```
rameeshm: We follow this pattern with in the Ranger Code for method returns, I would recommend to do the same
String ret = DEFAULT_MAPPING_SEPARATOR;
String val = prop.getProperty(SYNC_MAPPING_SEPARATOR);
if (StringUtilts.isNotEmpty(val))  {
      if (val.length() == 1) {
             ret = val;
      } else {
             throw new RuntimeException("More than one character found in RegEx Separator");
        }
 }
  return ret;
}

kumaab: Sure, will update.

spolavarpau1 (approved):

PR #211, opened 2023-02-08
RANGER-2713: Remove audit fields from XXPolicyRef objects

andrewluotechnologies:
XXPolicyRef objects have fields such as create time, update time, added by user ID, updated by user ID, but there fields are entirely useless since they are all copied from the XXPolicy object. In addition, while improving performance for creation of policies with large numbers of users, we discovered that a lot of time was being spent in JPA converting these Date objects especially. After removing these fields we saw a significant performance improvement (a secondary benefit is less database space usage).


mneethiraj:
@andrewluotechnologies  - thank you for the patch. It is merged in master and ranger-2.8 branches via #796.

PR #214, opened 2023-02-08
RANGER-4074: Fix x_user_module_perm entry creation with NULL module Id

kumaab:

assignPermissionToUser in XUserMgr assigns permissions for a module (which does not exist) to users with Auditor role. The PR fixes the bug.

## How was this patch tested?

Testing underway


spolavarpau1 (approved):

kumaab:
Tracked here #222

PR #218, opened 2023-02-20
Hdds 6986

djordje-mijatovic:


plugin-ozone/pom.xml
```diff
@@ -91,7 +91,7 @@ limitations under the License.
         <dependency>
             <groupId>org.apache.ozone</groupId>
             <artifactId>ozone-common</artifactId>
-	        <version>${ozone.version}</version>
+	        <version>1.4.0-SNAPSHOT</version>
```
djordje-mijatovic, line 94: I had to hardcode the version in order to test it locally as some changes are also done on the Ozone side. How are we going to handle it?

djordje-mijatovic:
We have added the json and xml files in order to have the Ranger configured differently in the test envireoument. For example, in the file ozone_test.json we have some preconfigured policies needed for the tests like this one for example:

```
"service": "test",
"name": "Snapshot",
"policyType": 0,
"policyPriority": 0,
"description": "Policy for user Snapshot",
"isAuditEnabled": true,
"resources": {
	"volume": {
		"values": [
			"volume"
		],
		"isExcludes": false,
		"isRecursive": false
	},
	"bucket": {
		"values": [
			"bucket"
		],
		"isExcludes": false,
		"isRecursive": false
	},
	"key": {
		"values": [
			".snapshot/snapshotUser/*"
		],
		"isExcludes": false,
		"isRecursive": true
	}
}
```
		
When we write more tests in the future it might be possible we will need more policies to be configured and this is the right place to do it.

plugin-ozone/src/main/java/org/apache/ranger/authorization/ozone/authorizer/RangerOzoneAuthorizer.java
```diff
@@ -151,6 +160,18 @@ public boolean checkAccess(IOzoneObj ozoneObject, RequestContext context) {
 		}
 
 		try {
+			if (snapShotName != null) {
+				rangerResource.setValue(KEY_RESOURCE_KEY,".snapshot/" + snapShotName + "/" + ozoneObj.getKeyName());
+				RangerAccessResult result = rangerPlugin
+						.isAccessAllowed(rangerRequest);
+				if (result == null) {
+					LOG.error("Ranger Plugin returned null. Returning false");
+					return false;
```
prashantpogde, line 169: If we return false from here for snapshotted path, we will miss out on emitting logs. Check the corresponding code path for non-snapshotted path below.

plugin-ozone/src/main/java/org/apache/ranger/authorization/ozone/authorizer/RangerOzoneAuthorizer.java
```diff
@@ -59,25 +59,34 @@ public class RangerOzoneAuthorizer implements IAccessAuthorizer {
...
 		}
+		if (context == null) {
+			LOG.error("Context object is null!!");
+			return returnValue;
+		}
 		OzoneObj ozoneObj = (OzoneObj) ozoneObject;
 		UserGroupInformation ugi = context.getClientUgi();
 		ACLType operation = context.getAclRights();
 		String resource = ozoneObj.getPath();
+		String snapShotName = context.getSnapshotName();
```
prashantpogde, line 89: We need one more field to specify snapshot scope e.g. whether it is volume or a bucket.

PR #220, opened 2023-02-21
Fix type of authWindowStartTime parameter in XXAuthSessionDao.getRecentAuthFailureCountByLoginId

andrewluotechnologies:


Wzh1997:
image
image


Why does this error occur when this method is called? There is an extra comma, which causes a SQL syntax error and thus the API call fails.

PR #231, opened 2023-03-07
Simplifies the dependency of `gethostname4j` for the ranger client

Yikf:
Currently, `ranger-plugin-commons` module relies on `gethostname4j`, `jna ` and `jna-platform` to get the native hostname, which is for `AuthzAuditEvent` only.

`gethostname4j`, `jna ` and `jna-platform` is too onerous, it doesn't seem worthwhile to maintain the three jars to get the hostname, and `gethostname4j`, `jna ` and `jna-platform` relies on the native lib, which prevents users from doing shade.


Yikf:
Could you please take a look if you find a moments, thanks @kumaab

PR #223, opened 2023-03-02
RANGER-4043: Fix 'getent group' command when enumerate groups is enabled

fateh288:
## What changes were proposed in this pull request?

'getent group' command incorrectly used with 2 parameters which caused command to return null instead of group info. Corrected the command usage

Original command: `String[] cmd = new String[] {"bash", "-c", command + " '" + group + "'"};`
Results in: bash -c getent group test_group1 'test_group1' - incorrect

New changed command: `String[] cmd = new String[] {"bash", "-c", command };`
Results in: bash -c getent group test_group1 - correct

## How was this patch tested?
ranger.usersync.group.enumerategroup property set to "root,wheel,daemon"
Obtained logs suggest getent command successfully got executed:
```
2023-03-01 21:11:58,625 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: Executing: [bash, -c, getent group root]
2023-03-01 21:11:58,630 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: bash -c getent group root for group root returned root:x:0:
2023-03-01 21:11:58,630 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: Executing: [bash, -c, getent group wheel]
2023-03-01 21:11:58,639 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: bash -c getent group wheel for group wheel returned wheel:x:10:jenkins
2023-03-01 21:11:58,639 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: Executing: [bash, -c, getent group daemon]
2023-03-01 21:11:58,647 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: bash -c getent group daemon for group daemon returned daemon:x:2:
2023-03-01 21:11:58,647 DEBUG org.apache.ranger.unixusersync.process.UnixUserGroupBuilder: Done adding extra groups 
```
Verified the obtained log results by executing commands such as "bash -c getent group daemon" which returned consistent results as logs.

PR #233, opened 2023-03-09
Ranger 4043: fixing getent group command

fateh288:
getent group command required one more change as suggested. 
https://github.com/apache/ranger/pull/223


ugsync/src/main/java/org/apache/ranger/unixusersync/process/UnixUserGroupBuilder.java
```diff
@@ -538,7 +538,7 @@ private void buildUnixGroupList(String allGroupsCmd, String groupCmd, boolean us
 
 			for (String group : groups) {
 				String command = String.format(groupCmd, group);
-				String[] cmd = new String[]{"bash", "-c", command + " '" + group + "'"};
+				String[] cmd = new String[]{"bash", "-c", command};
```
kumaab: This change is already present, please revise review.

PR #243, opened 2023-04-10
fix: mismatch the value of access_type and action

Wenkang-xu:
When I try to use the ranger audit logs to analysis the access info, the value of Access Type and Permission on the ranger web page confuses me. After readed the source code, I found the problem that mismatch the value of Access Type and Permission.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/audit/RangerDefaultAuditHandler.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/audit/RangerDefaultAuditHandler.java
@@ -124,10 +124,10 @@ public class RangerDefaultAuditHandler implements RangerAccessResultProcessor {
 			ret.setRequestData(request.getRequestData());
 			ret.setEventTime(request.getAccessTime() != null ? request.getAccessTime() : new Date());
 			ret.setUser(request.getUser());
-			ret.setAction(request.getAccessType());
+			ret.setAction(request.getAction());
 			ret.setAccessResult((short) (result.getIsAllowed() ? 1 : 0));
 			ret.setPolicyId(result.getPolicyId());
-			ret.setAccessType(request.getAction());
+			ret.setAccessType(request.getAccessType());
 			ret.setClientIP(request.getClientIPAddress());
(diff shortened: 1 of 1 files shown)


Wenkang-xu:
image
image
Look at the picture above, the first picture shows some hive audit logs, the second picture shows some prestodb audit logs.

### explain the problem
We know Permission on the web page is the Operation Type [[this.setAction(hiveOpTypeName);](https://github.com/apache/ranger/blob/b9d24af9f93053b70a5c692be13aa212b3e0fb26/hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAccessRequest.java)] which doesn't contain "select"([See this](https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/plugin/HiveOperationType.java)). For example: HiveAccessType.SELECT can reference many Operation Type, like QUERY, SHOWPARTITIONS etc. You can find in [this code](https://github.com/apache/ranger/blob/7901d88e2aeac1d515581a0e987e7d4318da78b1/hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java). 
Therefore we can clearly judge that mismatch the value between Access Type and Action.

### Test my code
After I changed the code which is my submit and repackaged it, the web page show autit logs as I expected.
image

image

ramackri (approved):
@mneethiraj 
It seems this needs to be merged into master 
https://github.com/apache/ranger/blob/master/agents-common/src/main/java/org/apache/ranger/plugin/audit/RangerDefaultAuditHandler.java#L128

mneethiraj:
@ramackri, @Wenkang-xu  - wouldn't swapping setAccessType() and setAction() implementation result in backward compatibility issue in audit logs being generated by all plugins that exist currently?

ramackri:
> @ramackri, @Wenkang-xu - wouldn't swapping setAccessType() and setAction() implementation result in backward compatibility issue in audit logs being generated by all plugins that exist currently?

Yes @mneethiraj i agree with your point

PR #253, opened 2023-05-09
RANGER-4230: Add REST APIs to force delete external users & groups

kumaab:
Addition of 2 new REST APIs to delete external users and groups.
The proposed approach uses existing methods for deletion of users & groups via 2 REST API calls(1 for users and 1 for groups), all the processing of deletes happen in Java with immediate transaction commit for every user/group.

Stats(time taken):
- Users:
  - Deletion of **50k** users takes: **33 mins 28 seconds** @ **40 ms** per delete operation (This is still expensive, more on this later)
- Groups:
  - Deletion of **100k** groups takes: **2 mins 51 seconds** @ **1.7 ms** per delete operation.

Current benchmark: **5.3** seconds for every delete operation. ([info](https://issues.apache.org/jira/browse/RANGER-3794))
Improvement(via this PR): **133X** (for ex: to delete 50k users via current python scripts it would take 3 days and 2 hrs approx.)

To further improve the running time of user deletes, db schema changes would be required as one possible approach.
On average per delete operation 35 ms/40 ms (~ 88%) of the time goes in updating 64 tables which have FK constraints on _**x_portal_user**_ (more info [here](https://github.com/apache/ranger/blob/aea071ca7c029e45e08beb7392d50c582f0e0c68/security-admin/src/main/java/org/apache/ranger/service/XPortalUserService.java#L195))

To avoid this "ON DELETE CASCADE/NULL" needs to be enabled for the respective db tables, which may improve the performance of the API significantly. This will be taken up in another jira to target best achievable running time for large no. of users/groups deletion.

## How was this patch tested?

Tested using docker.

Tested the latest patch with the following REST API calls:
* service/xusers/delete/external/groups?name=group_usertest_3982_b
* service/xusers/delete/external/groups?isVisible=0
* service/xusers/delete/external/groups?isVisible=0&groupSource=0
* service/xusers/delete/external/groups?syncSource=File
* service/xusers/delete/external/groups?syncSource=None

* service/xusers/delete/external/users?name=usertest_41305
* service/xusers/delete/external/users?isVisible=0&userSource=0
* service/xusers/delete/external/users?syncSource=None
* service/xusers/delete/external/users?emailAddress=xyz5@gmail.com

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
@@ -33,6 +33,7 @@ import java.util.Set;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.hadoop.util.Time;
 import org.apache.ranger.biz.ServiceDBStore.REMOVE_REF_TYPE;
 import org.apache.ranger.common.*;
 import org.apache.ranger.common.db.RangerTransactionSynchronizationAdapter;
@@ -2197,6 +2198,33 @@ public class XUserMgr extends XUserMgrBase {
 		}
 	}
 
+	public void bulkDeleteGroups(){
+		// get all external groups
+		long timeNow = Time.now();
+		List<Long> groupsIds = daoManager.getXXGroup().findByGroupSource(1);
+		logger.info("External groups fetched in memory!");
+		logger.info(String.format("Time taken to fetch users in memory = %d", (Time.now() - timeNow)));
+		logger.info("No of external groups is " + groupsIds.size());
+		long curTime = Time.now();
+		for(Long userId: groupsIds){
+			TransactionTemplate txTemplate = new TransactionTemplate(txManager);
+			txTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+			try {
+				txTemplate.execute(new TransactionCallback<Object>() {
+					@Override
+					public Object doInTransaction(TransactionStatus status) {
+						deleteXGroup(userId, true);
+						return null;
+					}
+				});
+			} catch (Throwable ex) {
+				logger.error("bulkDeleteGroups(): Failed to delete DB for group id: " + userId + " ", ex);
+				throw new RuntimeException(ex);
+			}
+		}
+		logger.info(String.format("Force Deletion of %d groups took %d milliseconds", groupsIds.size(), (Time.now() - curTime)));
+	}
+
 	private void blockIfZoneGroup(Long grpId) {
 		List<XXSecurityZoneRefGroup> zoneRefGrpList = daoManager.getXXSecurityZoneRefGroup().findByGroupId(grpId);
 		if (CollectionUtils.isNotEmpty(zoneRefGrpList)) {
@@ -2398,6 +2426,32 @@ public class XUserMgr extends XUserMgrBase {
 				xaBizUtil.createTrxLog(trxLogList);
 			}
 		}
+	}
+
+	public void bulkDeleteUsers(){
+		// get all external users
+		long timeNow = Time.now();
+		List<Long> userIds = daoManager.getXXUser().findByUserSource(1);
+		logger.info(String.format("%d External users fetched in memory!", userIds.size()));
+		logger.info(String.format("Time taken to fetch users in memory = %d", (Time.now() - timeNow)));
+		long curTime = Time.now();
+		for(Long userId: userIds){
+			TransactionTemplate txTemplate = new TransactionTemplate(txManager);
+			txTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+			try {
+				txTemplate.execute(new TransactionCallback<Object>() {
+					@Override
+					public Object doInTransaction(TransactionStatus status) {
+						deleteXUser(userId, true);
+						return null;
+					}
+				});
+			} catch (Throwable ex) {
+				logger.error("bulkDeleteUsers(): Failed to delete DB for user id: " + userId + " ", ex);
+				throw new RuntimeException(ex);
+			}
+		}
+		logger.info(String.format("Force Deletion of %d users took %d milliseconds", userIds.size(), (Time.now() - curTime)));
 	}
 
 	private void blockIfZoneUser(Long id) {
--- a/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
@@ -2198,31 +2198,29 @@ public class XUserMgr extends XUserMgrBase {
 		}
 	}
 
-	public void bulkDeleteGroups(){
-		// get all external groups
-		long timeNow = Time.now();
-		List<Long> groupsIds = daoManager.getXXGroup().findByGroupSource(1);
-		logger.info("External groups fetched in memory!");
-		logger.info(String.format("Time taken to fetch users in memory = %d", (Time.now() - timeNow)));
-		logger.info("No of external groups is " + groupsIds.size());
-		long curTime = Time.now();
-		for(Long userId: groupsIds){
+	public void forceDeleteGroups(List<VXGroup> groups){
+		long startTime = Time.now();
+		for(VXGroup group: groups){
+			Long groupId = group.getId();
 			TransactionTemplate txTemplate = new TransactionTemplate(txManager);
 			txTemplate.setPropagationBehavior(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
 			try {
 				txTemplate.execute(new TransactionCallback<Object>() {
 					@Override
 					public Object doInTransaction(TransactionStatus status) {
-						deleteXGroup(userId, true);
+						deleteXGroup(groupId, true);
 						return null;
 					}
 				});
 			} catch (Throwable ex) {
-				logger.error("bulkDeleteGroups(): Failed to delete DB for group id: " + userId + " ", ex);
+				logger.error("forceDeleteGroups(): Failed to delete group id: " + groupId + " ", ex);
 				throw new RuntimeException(ex);
 			}
 		}
-		logger.info(String.format("Force Deletion of %d groups took %d milliseconds", groupsIds.size(), (Time.now() - curTime)));
+		if (groups.size() > 0) {
+			logger.info(String.format("Force Deletion of %d groups took %d milliseconds",
+					groups.size(), (Time.now() - startTime)));
+		}
 	}
 
 	private void blockIfZoneGroup(Long grpId) {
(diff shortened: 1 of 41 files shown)


security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -863,6 +868,49 @@ public void deleteXUserByUserName(@PathParam("userName") String userName,
...
+	public Response deleteXUsers() {
+		if (!usersDeletionInProgress){
+			synchronized (XUserREST.class) {
+				usersDeletionInProgress = true;
+				xUserMgr.bulkDeleteUsers();
+				usersDeletionInProgress = false;
+			}
+			return Response.ok("External users deleted successfully").build();
+		}
+		return Response.ok("Users Deletion in progress, check ranger admin logs!").build();
```
mneethiraj: Instead of returning 'ok', consider returning an error code like 409 (conflict).

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -863,6 +868,49 @@ public void deleteXUserByUserName(@PathParam("userName") String userName,
...
+	public Response deleteXGroups() {
+		if (!groupsDeletionInProgress) {
+			synchronized (XUserREST.class) {
+				groupsDeletionInProgress = true;
+				xUserMgr.bulkDeleteGroups();
+				groupsDeletionInProgress = false;
+			}
+			return Response.ok("External groups deleted successfully").build();
+		}
+		return Response.ok("Groups Deletion in progress, check ranger admin logs!").build();
```
mneethiraj: Instead of returning 'ok', consider returning an error code like 409 (conflict).

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -165,6 +166,10 @@ public class XUserREST {
 	
 	static final Logger logger = LoggerFactory.getLogger(XUserMgr.class);
 
+	static volatile boolean usersDeletionInProgress = false;
```
mneethiraj: Use of static variables wouldn't work in HA deployments, as subsequent request might be going to a different instance. Why is it critical to avoid multiple simultaneous calls?

security-admin/src/main/resources/META-INF/jpa_named_queries.xml
```diff
@@ -161,6 +161,20 @@
 		</query>
 	</named-query>
 
+	<named-query name="XXUser.findByUserSource">
```
mneethiraj: XXUser.findByUserSource => XXUser.getUserIdsForSource

security-admin/src/main/resources/META-INF/jpa_named_queries.xml
```diff
@@ -161,6 +161,20 @@
...
+		<query>
+			SELECT user.id FROM XXUser user, XXPortalUser portalUser
+			WHERE user.name = portalUser.loginId and portalUser.userSource = :userSource
+		</query>
+	</named-query>
+
+	<named-query name="XXGroup.findByGroupSource">
```
mneethiraj: XXGroup.findByGroupSource => XXGroup.getGroupIdsForSource

security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
```diff
@@ -2197,6 +2198,33 @@ public void deleteXGroup(Long id, boolean force) {
 		}
 	}
 
+	public void bulkDeleteGroups(){
```
mneethiraj: bulkDeleteGroups() => forceDeleteAllExternalGroups()

security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
```diff
@@ -2400,6 +2428,32 @@ public synchronized void deleteXUser(Long id, boolean force) {
 		}
 	}
 
+	public void bulkDeleteUsers(){
```
mneethiraj: bulkDeleteUsers() => forceDeleteAllExternalUsers()

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -863,6 +868,49 @@ public void deleteXUserByUserName(@PathParam("userName") String userName,
...
+	/**
+	 * Proceed with caution: Force deletes users in bulk from the ranger db,
+	 * essentially serves as a cleanup Op for external users.
+	 * <tt>Delete</tt> happens one at a time with immediate commit on the transaction.
+	 */
+	@DELETE
+	@Path("/delete/external/users")
+	@PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
+	@Produces({ "application/json" })
+	public Response deleteXUsers() {
```
mneethiraj: Instead of an API specifically to delete all external users, consider a generic API that let's the call specify the users to delete via query-params - similar to in searchXUsers():

```
@DELETE
@Path("/users")
@PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
@Produces({ "application/json" })
public Response deleteUsers(@Context HttpServletRequest request) {
  SearchCriteria searchCriteria = searchUtil.extractCommonCriterias(request, xUserService.sortFields);

  // TODO: update searchCriteria with query-params 

  Long<Long> userIds = xUserMgr.deleteUsers(searchCriteria);

  return Response.ok(userIds.size() + " external users deleted successfully").build();
}

```

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -863,6 +868,49 @@ public void deleteXUserByUserName(@PathParam("userName") String userName,
...
+	/**
+	 * Proceed with caution: Force deletes groups in bulk from the ranger db,
+	 * essentially serves as a cleanup Op for external groups.
+	 * <tt>Delete</tt> happens one at a time with immediate commit on the transaction.
+	 */
+	@DELETE
+	@Path("/delete/external/groups")
+	@PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
+	@Produces({ "application/json" })
+	public Response deleteXGroups() {
```
mneethiraj: Instead of an API specifically to delete all external users, consider a generic API that let's the call specify the users to delete via query-params - similar to in searchXUsers():

```
@DELETE
@Path("/groups")
@PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
@Produces({ "application/json" })
public Response deleteGroups(@Context HttpServletRequest request) {
  SearchCriteria searchCriteria = searchUtil.extractCommonCriterias(request, xGroupUserService.sortFields);

  // TODO: update searchCriteria with query-params 

  Long<Long> groupIds = xUserMgr.deleteGroups(searchCriteria);

  return Response.ok(groupIds.size() + " external groups deleted successfully").build();
}
```

kumaab: thanks for the suggestion, changes included in the latest patch.

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
kumaab: thanks for the suggestion, changes included in the latest patch.

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
kumaab: Sure, makes sense. Sometimes these calls might take long time to get a response, the idea was to get a response indicating the progress(like 30% deletion complete or deletion in progress check back later) in the interim via another simultaneous call.

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -863,6 +864,61 @@ public void deleteXUserByUserName(@PathParam("userName") String userName,
 		xUserMgr.deleteXUser(vxUser.getId(), forceDelete);
 	}
 
+
+	/**
+	 * Proceed with <tt>caution</tt>: Force deletes users from the ranger db,
+	 * <tt>Delete</tt> happens one at a time with immediate commit on the transaction.
+	 */
+	@DELETE
+	@Path("/delete/external/users")
```
mneethiraj, line 876: I suggest replacing `external` in the URI with `force', since this API is not specific to external users i.e., userSource query-param.

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -863,6 +864,61 @@ public void deleteXUserByUserName(@PathParam("userName") String userName,
...
+		xUserMgr.forceDeleteUsers(userList.getList());
+		return Response.ok(userList.getListSize() + " users deleted successfully").build();
+	}
+
+	/**
+	 * Proceed with <tt>caution</tt>: Force deletes groups from the ranger db,
+	 * <tt>Delete</tt> happens one at a time with immediate commit on the transaction.
+	 */
+	@DELETE
+	@Path("/delete/external/groups")
```
mneethiraj, line 918: I suggest replacing `external` in the URI with `force', since this API is not specific to external group i.e., groupSource query-param.

security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
```diff
@@ -2400,6 +2426,31 @@ public synchronized void deleteXUser(Long id, boolean force) {
...
+				txTemplate.execute(new TransactionCallback<Object>() {
+					@Override
+					public Object doInTransaction(TransactionStatus status) {
+						deleteXUser(userId, true);
+						return null;
+					}
+				});
+			} catch (Throwable ex) {
+				logger.error("forceDeleteUsers(): Failed to delete user id: " + userId + " ", ex);
+				throw new RuntimeException(ex);
```
mneethiraj: This exception would cause delete operation to be aborted on the first failure, right? The response to the client wouldn't show that few users have been deleted. I would suggest the following changes:
- instead of aborting on the first failure, this method should attempt to delete subsequent users as well
- the return from this method should include list of users successfully deleted, and the list of users who couldn't be deleted

Similar updates in forceDeleteGroups() as well.

kumaab:
Addressed potential concerns of large memory consumption when bringing in the users and groups matching the search criteria, instead only bringing the ids in memory to be consumed later for deletion.

CC: @spolavarapu1, @mneethiraj

kumaab:
Tested the latest PR again with the same test cases as before (100k users & 50k groups), this incorporates changes from RANGER-4240 (from master), no significant difference in timings were observed(from the original timings - mentioned in description).

Before deletion:
ranger=> select count(*) from x_group_users;
 count  
--------
 199996
(1 row)
ranger=> select count(*) from x_group;
 count  
--------
 100002
(1 row)
ranger=> select count(*) from x_user;
 count 
-------
 50009
(1 row)

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
kumaab: The API is now specific to external users, updated the PR sometime back.

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
kumaab: The API is now specific to external groups, updated the PR sometime back.

security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
kumaab: Sure, I'll revise the PR to continue with deletes despite failures. 

I believe getting the usernames for all users deleted/failed deletes is an overhead since this would result in another db call(with new jpa query), and ignoring the rare occurrence of transaction exceptions(in which we might not get the logs from deleteUser()), info regarding every user is already logged as part of deleteUser() where we print the username of the user getting deleted. For debugging purposes in case of transaction failures, username of the user is still retrievable using the userid in the catch block.

kumaab:
Tested python client with the following test cases:
`from apache_ranger.client.ranger_client import *` 
 `ranger = RangerClientPrivate('http://localhost:6080', ('admin', 'rangerR0cks!'))`


* ranger.force_delete_external_groups("isVisible=1&syncSource=XYZ"))
* ranger.force_delete_external_users("isVisible=0&userRole=ROLE_ADMIN_AUDITOR")) # takes both params to filter
* ranger.force_delete_external_groups("isVisible=0&ihdfiohjf=453")) # only takes the first param
* ranger.force_delete_external_users("isVisible=0"))
* ranger.force_delete_external_users("name=usertest_142"))
* ranger.force_delete_external_groups("syncSource=ABC"))
* ranger.force_delete_external_groups("syncSource=None"))
* ranger.force_delete_external_users()) # deletes all external users
* ranger.force_delete_external_groups()) # deletes all external groups

spolavarpau1 (approved):

PR #256, opened 2023-05-15
RANGER-4237: Add update_credentials.py to update usernames/passwords

kumaab:

Aims to replace existing scripts with update_credentials.py

Fixes the following:
* changepasswordutil.py exposes passwords while taking input, use getpass instead.
* unused imports
* RMI call to java may hang indefinitely in case the command doesn't complete, add a timeout(60s) instead.
* Redundant logging constructs
* Fix while loops for parsing user input
* Use argparse & getpass to parse input
* Avoid sys.exit(1) in function calls to be more reusable.
* migrate to python3
* Consolidate functionalities to the CredManager class.

## How was this patch tested?

Tested username and password updates on ubuntu successfully using docker.


mneethiraj:
@kumaab - existing utilities changepasswordutil.py and changeusernameutil.py allow parameters to be passed via command-line, which makes it easier to invoke these utilities from scripts. The new script seems to support only interactive input of parameters. I suggest to retain support for command-line parameters.

Also, removal of existing utilities would likely break deployment tools (like Ambari?). Consider updating existing scripts, instead of replacing them with a new one.

kumaab:
@mneethiraj 

Sure, will update review to support command-line for new usernames and passwords as well.

From @pradeepagrawal8184's comment in https://reviews.apache.org/r/73937/, it looks like ambari scripts are not used anymore/ ambari doesn't support python3. The new script also follows PEP-8 guidelines for naming conventions.

PR #258, opened 2023-05-20
RANGER-4247: auditPolicyEvaluators should be set before logging it

eubnara:

Initialize `auditPolicyEvaluators` before calling `init()`. It is necessary to log `auditPolicyEvaluators` when debug logging mode is enabled.

## How was this patch tested?

Manual tests on on-premise Hadoop cluster.


mneethiraj:
Merged in master and ranger-2.8 branches.

PR #273, opened 2023-07-13
RANGER-4314: Delete nonexistent user return error status code

coldestlin:

https://issues.apache.org/jira/projects/RANGER/issues/RANGER-4314?filter=allissues&orderby=created+DESC

check return user is null before continue the code process, in order to make response code as 400 instead of 404

## How was this patch tested?

manual tests


ramackri (approved):
@mneethiraj  I feel this fix make sense to port into master 
VXGroup vXGroup = xGroupService.populateViewBean(xXGroup); --> this check is not needed when xXGroup is null

ramackri (commented):
resolve the conflicts

mneethiraj (approved):

ramackri:
@mneethiraj It seems this PR can be merged

PR #280, opened 2023-08-30
RANGER-4385: Fix role filtering logic in RolePredicateUtil

IOrzo:

Fix role filtering logic in RolePredicateUtil

## How was this patch tested?

1. Create a new user with role USER in Ranger Admin
2. Create a role like this table

|Role Name|Users|Groups|Roles|
|-|-|-|-|
|role_second|test|--|role_first|

3. use this account to query through the role name parameter role_first


ramackri (approved):
cc >> @mneethiraj

PR #295, opened 2023-12-21
RANGER-4623:Change the http client of the knox plug-in to avoid conne…

upczsh:
…ctivity failure caused by SSL links.

The ranger plug-in client does not need SSL handshake when calling knox, so it is replaced with apache http client to solve the SSL handshake problem and adjust the knox interface.
The ranger knox test connection error is : 
Unable to retrieve any files using given parameters, You can still save the repository and start creating policies, but you would not be able to use autocomplete for resource names. Check ranger_admin.log for more info.

org.apache.ranger.plugin.client.HadoopException: Exception on REST call to KnoxUrl : https://xxx:8443/gateway/admin/api/v1/topologies..
javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target.
PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target.

## How was this patch tested?
Click ranger knox service connectivity test, and pull down to view policy files, etc.
You can see it : https://issues.apache.org/jira/projects/RANGER/issues/RANGER-4623?filter=allissues

### DIFF
--- a/knox-agent/pom.xml
+++ b/knox-agent/pom.xml
@@ -102,6 +102,16 @@
             <version>${protobuf-java.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>${commons.lang3.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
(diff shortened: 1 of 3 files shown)


upczsh:
Requesting review for this PR.
@kumaab @bhavikpatel9977 @mneethiraj

PR #308, opened 2024-03-28
RANGER-4762:Prevent duplicate values for resource while validating policy

fateh288:
Previously: Via REST, policy creation possible when duplicate values specified for a resource as a list e.g. for a resource, multiple dbs can be mentioned like db:[test_db1,test_db1] and this policy is considered different from a policy already created for the resource [test_db1]
The same scenario is already prevented by Ranger UI as it prevents duplicate values for a resource (in a given policy)

After this patch: Error is thrown for duplicate resource values like [test_db1,test_db1] - {"statusCode":1,"msgDesc":"(0) Validation failure: error code[3056], reason[Values for the resource=database contained a duplicate value=test_db1. Ensure all values for a resource are unique], field[resource-values], subfield[database], type[semantically incorrect] "}

## How was this patch tested?

Unit tests added for the scenario
Manually tested via REST end points for policy creation that duplicate resource values for a resource throws above error.


kulkabhay (approved):

fateh288:
The changes have been merged : https://github.com/apache/ranger/commit/ffda775366ac8ac8a6869991dbab5e12d307a423

PR #310, opened 2024-04-17
RANGER-4776: SolrAuditDestination uses local SSLContext instead of se…

turcsanyip:
…tting the system-wide default

In order to avoid setting the system-wide `SSLContext` in `SolrAuditDestination` (which may affect other components running in the same JVM negatively), use Solr's `HttpClientUtil.setSocketFactoryRegistryProvider()` to pass the `SSLContext` to the Solr client locally.

## How was this patch tested?

We use the Ranger Plugin in Apache NiFi.
Before the change, Azure components in NiFi failed with "unable to find valid certification path to requested target" error.
After the change, the Ranger - Solr connection still works and the Azure components can also access the needed certificate in the system default truststore (cacerts).


mneethiraj:
@rameeshm  - can you please review this update?

PR #315, opened 2024-06-12
RANGER-4818: Fix updated users with role assignments from undergoing role reset to ROLE_USER

kumaab:

Allow role resets for users not part of the group based role assignment rules in the last page of the paged requests from usersync to ranger admin. 

## How was this patch tested?

In Progress


kumaab:
New Request open here: https://github.com/apache/ranger/pull/320

PR #332, opened 2024-07-11
RANGER-4858: Remove usage of htrace-core4

kokosing:
Remove usage of htrace-core4

This project is retired. Since it was not updated for a long time it
contains plenty of old dependencies (shaded) that are marked by security
scanners as source of vulnerabilities.


kokosing:
@mneethiraj it looks like htrace is having a shaded dependency for jackson which is outdated and have plenty of CVEs. I am not sure how it was used in Ranger from my short investigation it looks like is either not used or there is missing infrastructure part in tests. Since htrace is no longer supported I think it should be removed from Ranger and potentially if needed I would recommend using opentelemetry. 

Would you like to help me to contribute this change? Also our scanners flagged other CVEs related issues, if this contribution is successful I would be very happy to update other libraries.

mneethiraj (approved):

mneethiraj:
[~kokosing] - thank you for the contribution. The patch is merged in master and ranger-2.5 branches.

kokosing:
Thank you!

PR #346, opened 2024-07-20
RANGER-4307 Allow the synchronization time interval of LDAP users to …

13668215435:
Allow the synchronization time interval of LDAP users to exceed the one hour limit, and users can customize the time interval


13668215435:
I think it can be modified in the following way：reuse the config param "ranger.usersync.sleeptimeinmillisbetweensynccycle" in the config file，rather than just using him as a criterion.

PR #269, opened 2023-07-03
[RANGER-4307] Allow users customize the time interval

zhutong6688:
Modified the UserGroupSyncConfig.java file and added the configuration of user forced custom time synchronization. If enabled, LDAP users can be synchronized according to the user defined time.

## What changes were proposed in this pull request?

RANGER-4307: Added configuration option for ldap: ranger. usersync. ldap. force. sleep time in millisbetweensynccycle. enabled. If enabled (set to true), users will be synchronized according to the time interval set by the user. According to the actual production and usage situation, this demand is often urgently requested by users. this also gives the user more autonomy.

## How was this patch tested?

It was tested in the company's Big data cluster test environment, and later synchronized to the production environment. After rigorous testing, this modification has fully achieved its purpose. The test screenshot is as follows:
image

PR #350, opened 2024-07-20
Ranger-4307 Allow the synchronization time interval of LDAP users to exceed the one hour limit, and users can customize the time interval

13668215435:
I think it can be modified in the following way：reuse the config param "ranger.usersync.sleeptimeinmillisbetweensynccycle" in the config file，rather than just using him as a criterion.


kumaab:
This is merged in master: https://github.com/apache/ranger/commit/854a113 via #269

PR #351, opened 2024-07-20
Exclude org.codehaus.jackson

kokosing:
Exclude org.codehaus.jackson

This libraries are old (2013) and have plenty of CVEs. They were
migrated to org.fasterxml.jackson.

### DIFF
--- a/ranger-hive-plugin-shim/pom.xml
+++ b/ranger-hive-plugin-shim/pom.xml
@@ -52,6 +52,10 @@
                     <groupId>org.apache.commons</groupId>
                     <artifactId>commons-compress</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.codehaus.jackson</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


kokosing:
FYI @mneethiraj 

See https://github.com/advisories/GHSA-c27h-mcmw-48hv and https://mvnrepository.com/artifact/org.codehaus.jackson/jackson-mapper-asl

kokosing:
@mneethiraj why shim modules requires so many dependencies? I see that in hive plugin as well as in hbase. The more dependencies we put the higher risk for CVEs. How can I check what is actually used? I removed some of them and I see that project (module) compiles just fine, is this good enough for testing?

mneethiraj:
> why shim modules requires so many dependencies

@kokosing  - you are right, shim modules only have dependency on libraries that include classes referenced in authorization interface (like HiveAuthorizer, HiveAuthorizerFactory), and ranger-plugin-classloader library.

About this patch, given org.codehaus.jackson libraries are not included in Hive plugin packaging, is it necessary to exclude them from pom.xml file?

kokosing:
> About this patch, given org.codehaus.jackson libraries are not included in Hive plugin packaging, is it necessary to exclude them from pom.xml file?

I used `mvn dependency:tree` and I excluded it until the moment I no longer saw this dependency.

> shim modules only have dependency on libraries that include classes referenced in authorization interface (like HiveAuthorizer, HiveAuthorizerFactory), and ranger-plugin-classloader library.

Is it safe to remove these dependencies and assume that if project compiles we are good?

kumaab:
Please rebase the PR to run all checks, 1 check is missing.

kokosing:
Done

mneethiraj:
@kokosing  - the patch is merged in master and ranger-2.5 branches. Thank you!

kokosing:
Thank you!

PR #354, opened 2024-07-25
Update solr to 9.4.1

kokosing:
Update solr to 9.4.1

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -200,7 +200,7 @@
         <servlet.api.version>2.5</servlet.api.version>
         <slf4j-api.version>2.0.13</slf4j-api.version>
         <slf4j.version>2.0.13</slf4j.version>
-        <solr.version>8.11.2</solr.version>
+        <solr.version>9.4.1</solr.version>
         <spring-ldap-core.version>2.4.1</spring-ldap-core.version>
         <springframework.security.version>5.7.12</springframework.security.version>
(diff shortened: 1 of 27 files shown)


pom.xml
```diff
@@ -200,7 +200,7 @@
         <servlet.api.version>2.5</servlet.api.version>
         <slf4j-api.version>2.0.13</slf4j-api.version>
         <slf4j.version>2.0.13</slf4j.version>
-        <solr.version>8.11.2</solr.version>
```
mneethiraj, line 203: Build fails with Solr version updated to 9.4.1.  Solr 8.11.3 is the last supported version in JDK8. Higher versions require JDK11.

mneethiraj:
@kokosing  - Solr version update to 8.11.3 is merged in master and ranger-2.5 branches, via RANGER-4870. I suggest tracking the config cleanup in a separate PR. Thank you!

kokosing:
Thank you for merging and feedback!

kokosing:
See https://github.com/apache/ranger/pull/358

pom.xml
kokosing: > Build fails with Solr version updated to 9.4.1. Solr 8.11.3 is the last supported version in JDK8. Higher versions require JDK11.

I noticed that and I couldn't figure out why. In IDE it was working just fine, but not in CLI.

PR #361, opened 2024-07-26
Exclude old protobuf transitive dependency

kokosing:
Exclude old protobuf transitive dependency

Exclude versions 2.5.0 and 3.15.0


kokosing:
It won't work. Protobuf is needed.

PR #366, opened 2024-07-27
Exclude avro dependency

kokosing:
Exclude avro dependency

Avro is not used by Ranger and it ships CVEs to the project.

### DIFF
--- a/hbase-agent/pom.xml
+++ b/hbase-agent/pom.xml
@@ -81,6 +81,10 @@
                     <groupId>org.apache.zookeeper</groupId>
                     <artifactId>zookeeper</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>org.apache.avro</groupId>
+                    <artifactId>avro</artifactId>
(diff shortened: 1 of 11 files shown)


kokosing:
FYI @mneethiraj @ksobolew

ksobolew:
I can see some references to Avro in `plugin-schema-registry` and assembly for `kms`, but indeed, no direct dependencies. I'm not qualified to tell if this means that Avro is being used for something.

kokosing:
@mneethiraj Any chance to land this before release-ing Ranger 2.5.0?

mneethiraj:
@kokosing  - this update will likely not be to be included in Ranger 2.5.0 release; release-candidate #3 is already out for voting. I suggest getting this merged in master branch for 3.0.0 release; if a maintenance release is necessary, 2.5.1 release can be considered.

mneethiraj:
Patch merged via RANGER-4884.

kokosing:
Thank you!

PR #369, opened 2024-07-29
RANGER-4876: Plugin Analytics 

fateh288:
Accumulate count of isAccessAllowed and log periodically

The new class is meant to keep running count of how many times a plugin is called by a service and logs then periodically (every 300 seconds). Count is reset after logging. 

## How was this patch tested?

Tested in HDFS plugin, log with analytics prints every 300 seconds.


agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -86,6 +87,7 @@ public RangerBasePlugin(String serviceType, String serviceName, String appId) {
 	}
 
 	public RangerBasePlugin(RangerPluginConfig pluginConfig) {
+		this.pluginAnalytics = new RangerPluginAnalytics();
```
vyommani, line 90: RangerBasePlugin has 6  constructor, "pluginAnalytics "  instance variable is not getting initialised from all 6 code paths.  Please do let me know if i am looking the wrong code base.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -494,6 +497,9 @@ public void cleanup() {
 		if (policyEngine != null) {
 			((RangerPolicyEngineImpl) policyEngine).releaseResources(true);
 		}
+		if (pluginAnalytics!=null) {
```
vyommani: do we need null check ?.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerPluginAnalytics.java
```diff
@@ -0,0 +1,75 @@
...
+    if (statusHandler == null){
+      LOG.info("statusHandler is null, stop not required for asyncLoggerPluginAnalytics");
+      return;
+    }
+    LOG.info("Attempting to cancel asyncLoggerPluginAnalytics");
+    try {
+      scheduler.shutdown();
+      statusHandler.cancel(true);
+    }
+    catch (Exception e){
```
vyommani: can you please  replace "Exception e" to the specific exception  which are thrown from try block ?

fateh288: Fixed. No exception thrown by these functions according to the documentation

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: I was adding it as a safety net based on other surrounding variables such as policyEnginer. Not really required. Removed it on second thoughts since it is never set to null unlike policyEngine.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: I see just 4 constructors - line 511, 515, 519 and 565 - in RangerBasePlugin.java
@vyommani Can you point to the other 2 constructors you are seeing ?
Constructor 1: Always calls constructor 3
Constructor 2: Always calls constructor 4
So I added pluginAnalytics function calls only for constructor 3 and constructor 4. 
Adding it to all the 4 constructors would result in double counting

PR #390, opened 2024-09-12
RANGER-4931:Handling Unknown column 'rangerusersyncID' in 'field list'

basapuram-kumar:
Proposed cahnges are mentioned at [RANGER-4931](https://issues.apache.org/jira/browse/RANGER-4931)

## How was this patch tested?
tested this patch on my local cluster during the migration of ranger from 2.3.0 to 2.5.0, and with proposed changes, able to fix this issue.


prabhjyotsingh:
LGTM!

@princeap173 Can you please help review as well.

PR #393, opened 2024-09-13
RANGER-4607: Ranger REST API improvements

pradeepagrawal8184:

RANGER-4545: DELETE /assets/resources/{resource_id} API should return proper status code for non admin users
RANGER-4546: /assets/ugsyncAudits/{sync_source} API is accessible by user without permission on audit module
RANGER-4548: Return proper error message in the response for /tags/tags, /tags/resources and /tags/types API for non admin users
RANGER-4547: The reponse metrics (pagination values) for the /assets/ugsyncAudits/{sync_source} API is not proper
RANGER-4549: Non admin users cannot access /public/v2/api/roles/names and /public/v2/api/roles/name/{name} API, but can access /public/v2/api/roles API
RANGER-4551: No response returned for /assets/policyList/{service_name} API
RANGER-4550: API request to /assets/resource/{id} returns no response
RANGER-4552: Response metrics for /assets/report is not proper, and pagination does not work
RANGER-4553: Response metrics for /xaudit/trx_log not proper
RANGER-4554: Response metrics for /assets/resources not proper
RANGER-4555: Response metrics for /assets/assets API not proper
RANGER-4573: /xaudit/trx_log API not accessible by keyadmin user
RANGER-4578: /xuser/groupgroups and /xuser/groupusers APIs allow creation of entities even without groupId / userId fields in the request
RANGER-4574: /public/v2/api/service/{service_name}/policy/{policy_name} API returns policies for users without access to the policy
RANGER-4575: /plugins/policy/{policy_id}/version/{version_number} API returns policies for users without access to the policy
RANGER-4576: User without access to policy is able to fetch policy details using /plugins/policies/{service_type}/for-resource API endpoint
RANGER-4577: UI and API behaviour for fetching users not consistent for keyadmin users
RANGER-4589: keyadmin user can update the user password via UI but cannot update the user password using /users/{user_id}/passwordchange API
RANGER-4588: /xaudit/trx_log/{trx_log_id} is not accessible by keyadmin user
RANGER-4591: keyadmin user can access non kms related admin audits using /assets/report/{transaction_id} API
RANGER-4594: keyadmin user can mark ROLE_USER users as disabled by setting status to 0 using /users API
RANGER-4595: keyadmin user able to view the user permission objects via /users API
RANGER-4596: keyadmin can fetch the details of admin and auditor users through /users API endpoint
RANGER-4598: ROLE_USER cannot acccess /xusers/groups API but can access /xusers/groups/groupName/{group_name} API
RANGER-4586: XUserREST and UserREST API improvement for keyadmin users
Change-Id: I1fa52a99049d81e58c40d071211d62b278ff8ef1

## How was this patch tested?

Tested affected REST APIs using curl command and found the response as per the expectation.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/XAuditMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XAuditMgr.java
@@ -52,35 +52,35 @@ public class XAuditMgr extends XAuditMgrBase {
 	RangerBizUtil rangerBizUtil;
 
 	public VXTrxLog getXTrxLog(Long id) {
-		checkAdminAccess();
+		checkAllAdminsAccess();
 		return super.getXTrxLog(id);
 	}
 
 	public VXTrxLog createXTrxLog(VXTrxLog vXTrxLog) {
 		checkAdminAccess();
-                rangerBizUtil.blockAuditorRoleUser();
+		rangerBizUtil.blockAuditorRoleUser();
 		return super.createXTrxLog(vXTrxLog);
 	}
 
 	public VXTrxLog updateXTrxLog(VXTrxLog vXTrxLog) {
 		checkAdminAccess();
-                rangerBizUtil.blockAuditorRoleUser();
+		rangerBizUtil.blockAuditorRoleUser();
 		return super.updateXTrxLog(vXTrxLog);
 	}
 
 	public void deleteXTrxLog(Long id, boolean force) {
 		checkAdminAccess();
-                rangerBizUtil.blockAuditorRoleUser();
+		rangerBizUtil.blockAuditorRoleUser();
 		super.deleteXTrxLog(id, force);
 	}
 
 	public VXTrxLogList searchXTrxLogs(SearchCriteria searchCriteria) {
-		checkAdminAccess();
+		checkAllAdminsAccess();
 		return super.searchXTrxLogs(searchCriteria);
 	}
 
 	public VXLong getXTrxLogSearchCount(SearchCriteria searchCriteria) {
-		checkAdminAccess();
+		checkAllAdminsAccess();
 		return super.getXTrxLogSearchCount(searchCriteria);
 	}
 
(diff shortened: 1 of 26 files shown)


security-admin/src/main/java/org/apache/ranger/biz/XAuditMgr.java
```diff
@@ -52,8 +52,11 @@ public class XAuditMgr extends XAuditMgrBase {
 	RangerBizUtil rangerBizUtil;
 
 	public VXTrxLog getXTrxLog(Long id) {
-		checkAdminAccess();
-		return super.getXTrxLog(id);
+		if (rangerBizUtil.isAdmin() || rangerBizUtil.isKeyAdmin() || rangerBizUtil.isAuditAdmin() || rangerBizUtil.isAuditKeyAdmin()) {
```
kumaab: Please see if a private function can be re-used at lines 55, 81 and 90, for eg:

```
private boolean checkAdminAccess(){
    if (rangerBizUtil.isAdmin() || rangerBizUtil.isKeyAdmin() || rangerBizUtil.isAuditAdmin() || rangerBizUtil.isAuditKeyAdmin()){
        return true;
    } else {
        throw restErrorUtil.createRESTException(HttpServletResponse.SC_FORBIDDEN, "User doesn't have permissions to perform this action", true);
    }
}
```

security-admin/src/main/java/org/apache/ranger/rest/XUserREST.java
```diff
@@ -608,69 +622,6 @@ public VXLong countXGroupUsers(@Context HttpServletRequest request) {
 		return xUserMgr.getXGroupUserSearchCount(searchCriteria);
 	}
 
-	// Handle XGroupGroup
```
kumaab, line 611: Please see if the referenced methods (for REST APIs with path /groupgroups/.. ) in XUserMgr and XUserMgrBase can be removed as well?

PR #395, opened 2024-09-24
RANGER-4920: kms docker - Fix kms service url, volume support for configs

fateh288:

1) Fixed service url for kms as key create operation was not succeeding
2) Introduced volume support for kms docker images so that image re-create is not required for change in configs 

## How was this patch tested?

1) Created kms docker image
2) Made some changes to kms-logback.xml
3) Restart kms docker container
4) updated kms-logback.xml correctly used now for logging (image re-create not required)

### DIFF
--- a/plugin-kms/scripts/enable-kms-plugin.sh
+++ b/plugin-kms/scripts/enable-kms-plugin.sh
@@ -377,6 +377,13 @@ then
         	fi
 			archivefn="${HCOMPONENT_CONF_DIR}/.${orgfn}.${dt}"
         	newfn="${HCOMPONENT_CONF_DIR}/.${orgfn}-new.${dt}"
+          if [ "${KMS_IN_DOCKER}" ]; then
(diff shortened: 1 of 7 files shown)


kumaab (approved):

plugin-kms/scripts/enable-kms-plugin.sh
```diff
@@ -377,6 +377,13 @@ then
         	fi
 			archivefn="${HCOMPONENT_CONF_DIR}/.${orgfn}.${dt}"
         	newfn="${HCOMPONENT_CONF_DIR}/.${orgfn}-new.${dt}"
+          if [ "${KMS_IN_DOCKER}" ]; then
+            TEMP_DIR="${HCOMPONENT_
+            CONF_DIR}"/saved
```
mneethiraj: Lines 381 and 382 should be combined?

fateh288: Fixed

PR #397, opened 2024-09-27
RANGER-4943: Error in ElasticSearchAuditDestination shutting down RestHighLevelClient client

FerArribas14:
Error in ElasticSearchAuditDestination shutting down RestHighLevelClient client

RANGER-4943: Error in ElasticSearchAuditDestination shutting down RestHighLevelClient client
Error encountered -> ElasticSearchAccessAuditsServiceTest -> ERROR org.apache.ranger.audit.provider.BaseAuditHandler – Error sending message to ElasticSearch java.lang.RuntimeException: Cannot execute request; I/O reactor status: STOPPED</resthighlevelclient>.

Details:

Error in using a try-with-resources block in ElasticSearchAuditDestination because it closes the ElasticSearch client prematurely. In a try-with-resources block, any resource that implements the AutoCloseable interface will automatically close at the end of the block, which may cause the ElasticSearch client to close before its use is complete.  To avoid this problem, we remove the try-with-resources usage for the ElasticSearch client because it already explicitly closes the client when it is no longer needed, specifically in RestHighLevelClient which implements Closeable's close method.

## How was this patch tested?

By running ./build_ranger_using_docker.sh and unit test ElasticSearchAccessAuditsServiceTest

ExampleDoc
DocsElastic
AuditsApacheRanger


BaoICTHustK67:
Hi is there any update on this ?

kumaab (approved):
+1, the patch has also been reviewed here: https://reviews.apache.org/r/75218/

kulkabhay (commented):
Consider closing the restHighLevelClient before line 361 returns null (when an exception is encountered).

mneethiraj (approved):

PR #410, opened 2024-11-04
RANGER-4670: RANGER-4977: hbase plugin: SELF_AND_ALL_DESCENDANTS Reso…

fateh288:
…urceMatchingScope, support shortcircuit column auth

RANGER-4977: Introduced a new SELF_AND_ALL_DESCENDANTS ResourceMatchingScope which is now used for column family authorization which fixes bug in scan operation wherein a denied column was also being returned in the scan operation. 

RANGER-4670: A property/flag can be set via hbase service configs wherein during column level authorization, it can be used to determine if an authorization at column family can be used to determine if column authorization can be shortcircuited as an optimization (there will be audit behavior changes in this case). This is useful for improving performance especially for multiget and multiput workloads wherein thousands of columns can come in the authorization request and just checking column family level access can improve performance drastically

## How was this patch tested?

New unit test cases have been added for the policy engine for the new ResourceMatchingScope of SELF_AND_ALL_DESCENDANTS. 
New test cases have also been added for hbase. The test cases cover the scan operation bug and also covers new unit test cases for testing correctness of column authorization shortcircuiting. These test cases do not work on master branch (because they are ignored using junit ignore annotation) currently https://issues.apache.org/jira/browse/RANGER-4686 but I have verified that they work on my private fork.


agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
```diff
@@ -1512,4 +1504,27 @@ private List<RangerConditionEvaluator> createPolicyConditionEvaluators(RangerPol
...
+						case SELF_AND_ALL_DESCENDANTS: {
+							ret = matchType != RangerPolicyResourceMatcher.MatchType.NONE;
+							break;
+						}
+						default: {
+							ret = matchType == RangerPolicyResourceMatcher.MatchType.SELF || matchType == RangerPolicyResourceMatcher.MatchType.SELF_AND_ALL_DESCENDANTS;
+							break;
+						}
+				}
+		}
```
mneethiraj: when scope is null, shouldn't the return be true if matchType is SELF or SELF_AND_ALL_DESCENDANTS?

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
```diff
@@ -1512,4 +1504,27 @@ private List<RangerConditionEvaluator> createPolicyConditionEvaluators(RangerPol
...
+			ret = matchType !=  RangerPolicyResourceMatcher.MatchType.NONE;
+		}
+		else if (scope!=null) {
+				switch (scope) {
+						case SELF_OR_DESCENDANTS: {
+							ret = matchType != RangerPolicyResourceMatcher.MatchType.NONE;
+							break;
+						}
+						case SELF_AND_ALL_DESCENDANTS: {
+							ret = matchType != RangerPolicyResourceMatcher.MatchType.NONE;
```
mneethiraj: shouldn't matchType be SELF_AND_ALL_DESCENDENTS here?

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
```diff
@@ -1512,4 +1504,27 @@ private List<RangerConditionEvaluator> createPolicyConditionEvaluators(RangerPol
...
 
+	private static boolean isMatchForResourceMatchingScope(final RangerAccessRequest.ResourceMatchingScope scope, final RangerPolicyResourceMatcher.MatchType matchType, boolean isAnyMatch) {
+		boolean ret = false;
+		if (isAnyMatch){
+			ret = matchType !=  RangerPolicyResourceMatcher.MatchType.NONE;
+		}
+		else if (scope!=null) {
+				switch (scope) {
+						case SELF_OR_DESCENDANTS: {
+							ret = matchType != RangerPolicyResourceMatcher.MatchType.NONE;
```
mneethiraj: when scope is SELF_OR_DESCENDENTS, is it okay to return true when matchType is ANCESTOR?

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
fateh288: scope!=null always as per current implementation of RangerAccessRequestImpl .
`private ResourceMatchingScope resourceMatchingScope = ResourceMatchingScope.SELF;`
The default value is always SELF. But I see it can be somehow null if interface is implemented differently (I think we should prevent this)
Can there (or is there currently ) be a use case of null ResourceMatchingScope ?
However, I think a check for not null would be required for current refactoring logic as we cannot have null in switch case and we should return isMatched=False in this case unless there is an explicit use case for the same 
(if ResourceMatchingScope is null then I think isMatched should be false as matchType seems irrelevant).

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
fateh288: matchType=DESCENDANT is the scenario we are also interested in.
if policy resource is (t1,cf1,c1) and accessed resource is (t1,cf1) then matchType is DESCENDANT which is what we want to match as it is a valid policy to evaluate for this case.

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
fateh288: This would be out of scope for this PR.
SELF_OR_DESCENDANT ResourceMatchingScope is being used in many other places of the code. We would need more analysis and testing for making changes to this part of the code. This PR tries to keep current behavior as is.

mneethiraj: My point is - this PR introduces a regression in handing SELF_OR_DESCENDANT scope for ANCESTOR match. Please review and update.

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
mneethiraj: @fateh288  - the regression in handling scope==null should be addressed in this PR.

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
mneethiraj: @fateh288  - my bad. The regression was for earlier comment. I agree that handling of ANCESTOR is outside the scope of this PR.

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
mneethiraj: @fateh288  - case blocks for `SELF_OR_DESCENDANTS` and `SELF_AND_ALL_DESCENDANTS` are the same, which doesn't look correct. Can you review?

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
fateh288: Yes, I identified the issue is in test cases defined in test_policyengine_resource_with_req_expressions.json
The ResourceMatchingScope defined in these test cases is SELF_OR_CHILD which is not a valid value in the ResourceMatchingScope enum and results in null scope. 

I tried changing ResourceMatchingScope to SELF here and all the test cases pass here. Do you suggest doing this ? Or should we add SELF_OR_CHILD as a valid ResourceMatchingScope ?

Yes, I can handle null as a scope or prevent regressions, but ideally it should be an invalid scenario if implemented correctly.

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java
fateh288: Yes, 'case' blocks for both SELF_OR_DESCENDANTS and SELF_AND_ALL_DESCENDANTS would be the same. 

ResourceMatchingScope=SELF_OR_DESCENDANTS will result in isMatched = true for all cases of matchType in {SELF, DESCENDANT, ANCESTOR, SELF_AND_ALL_DESCENDANTS} --- I am not sure here if matchType is Ancestor then why isMatched=true makes sense. 
For ResourceMatchingScope=SELF_OR_DESCENDANTS, I tried changing 
`ret = matchType == RangerPolicyResourceMatcher.MatchType.SELF || matchType == RangerPolicyResourceMatcher.MatchType.SELF_AND_ALL_DESCENDANTS || matchType == RangerPolicyResourceMatcher.MatchType.DESCENDANT;`

After this change, agents-common test cases pass successfully even after I change it to above (do you suggest changing it to this ?).

The same works for SELF_AND_ALL_DESCENDANTS too.
 
In the current implementation, the difference between SELF_OR_DESCENDANTS and SELF_AND_ALL_DESCENDANTS comes in implementation of updateAccessResult() https://github.com/apache/ranger/blob/c9ee929d025dd73689ba1f242e3efdd23d8c9c8a/agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerDefaultPolicyEvaluator.java#L533

fateh288: @mneethiraj @kulkabhay 
Any ideas on how to proceed? 
I have resolved the merge conflicts.
Essentially with SELF_AND_ALL_DESCENDANTS we want same behavior as SELF_OR_DESCENDANTS but the special case for matchType!=DESCENDANTS needs to be bypassed

fateh288:
New changes proposed in https://github.com/apache/ranger/pull/417

PR #413, opened 2024-11-25
RANGER-5005: Use sortpom maven plugin to improve readability of pom files

mneethiraj:
Keeping pom.xml file contents sorted will improve readability. This PR uses sortpom maven plugin to enforce that pom.xml file contents are in sorted order.

## How was this patch tested?
Build with unit tests completed successfully.


plugin-nestedstructure/pom.xml
```diff
@@ -15,92 +15,90 @@
...
-                <version>15.4</version>
-            </dependency>
-            <dependency>
-                <groupId>org.testng</groupId>
-                <artifactId>testng</artifactId>
-                <version>${testng.version}</version>
-                <scope>test</scope>
-            </dependency>
-        </dependencies>
+    <dependencies>
```
kumaab, line 40: A lot of dependencies here use hard-coded versions, may be we should keep them in root pom.xml ?

plugin-schema-registry/pom.xml
```diff
@@ -15,27 +15,25 @@
...
         <version>3.0.0-SNAPSHOT</version>
         <relativePath>..</relativePath>
     </parent>
+    <artifactId>ranger-schema-registry-plugin</artifactId>
+    <packaging>jar</packaging>
+    <name>SchemaRegistry Security Plugin</name>
+    <description>SchemaRegistry Security Plugin</description>
 
     <properties>
         <avro.version>1.11.3</avro.version>
```
kumaab, line 33: Is it better to keep them in root pom.xml to be consistent?

ranger-authn/pom.xml
```diff
@@ -15,56 +15,36 @@
...
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
     <dependencies>
         <dependency>
-            <groupId>org.apache.hadoop</groupId>
-            <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version}</version>
-            <exclusions>
```
kumaab, line 40: It appears that Exclusions for `hadoop-common` were removed but not added back.

ranger-metrics/pom.xml
```diff
@@ -15,48 +15,31 @@
...
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>${hadoop.version}</version>
-            <exclusions>
```
kumaab, line 41: Exclusions are removed from here as well.

ranger-plugin-classloader/pom.xml
```diff
@@ -51,4 +50,5 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <url>http://maven.apache.org</url>
```
kumaab: Minor: Line 53 can be removed ?

ugsync/pom.xml
```diff
@@ -37,42 +73,47 @@
...
                 </exclusion>
             </exclusions>
         </dependency>
         <dependency>
-            <groupId>commons-lang</groupId>
-            <artifactId>commons-lang</artifactId>
-            <version>${commons.lang.version}</version>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-csv</artifactId>
+            <version>1.0</version>
```
kumaab: minor: can be moved to the root pom.xml.

plugin-schema-registry/pom.xml
mneethiraj: When a dependency is used only within a plugin module, I suggest keeping its version property in plugin module's pom.xml. For rest of Ranger libraries, version properties should be consolidated in root pom.xml.

plugin-nestedstructure/pom.xml
mneethiraj: For dependencies referenced only in a plugin module, I suggest keeping their version properties in that plugin module's pom.xml. Converting hardcoded to properties is a good idea.

ranger-authn/pom.xml
mneethiraj: Exclusions and version defined in root pom.xml (for hadoop-common) will be applied here too. Hence remove them from child modules.

ranger-metrics/pom.xml
mneethiraj: Exclusions and version defined in root pom.xml will be applied here. Hence removed them from here.

PR #414, opened 2024-11-25
RANGER-5007: Expose service configs in RangerBasePlugin

fateh288:

Expose service configs map in RangerBasePlugin

Use case: 
Read properties and change plugin behavior without service restart.
There seems to be an emerging pattern wherein properties set via service configs need to be read to change plugin behavior.
Example: https://issues.apache.org/jira/browse/RANGER-4670 

## How was this patch tested?

mvn clean compile verify

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -76,7 +76,7 @@ public class RangerBasePlugin {
 	private final List<RangerChainedPlugin>   chainedPlugins;
 	private final boolean                     dedupStrings;
 	private       boolean                     isUserStoreEnricherAddedImplcitly = false;
-	private 			Map<String, String>					serviceConfigs;
+	private       Map<String, String>         serviceConfigs;
 
 
 	public RangerBasePlugin(String serviceType, String appId) {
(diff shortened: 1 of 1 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -308,7 +309,7 @@ public void setPolicies(ServicePolicies policies) {
 		if (LOG.isDebugEnabled()) {
 			LOG.debug("==> setPolicies(" + policies + ")");
 		}
-
+		this.serviceConfigs = policies != null ? (policies.getServiceConfig() != null ? policies.getServiceConfig() : new HashMap<>()) : new HashMap<>();
```
mneethiraj: `this.serviceConfigs = (policies != null && policies.getServiceConfig() != null) ? policies.getServiceConfig() : new HashMap<>();`

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -76,6 +76,7 @@ public class RangerBasePlugin {
 	private final List<RangerChainedPlugin>   chainedPlugins;
 	private final boolean                     dedupStrings;
 	private       boolean                     isUserStoreEnricherAddedImplcitly = false;
+	private 			Map<String, String>					serviceConfigs;
```
mneethiraj: White space/tab alignment.

PR #420, opened 2024-12-03
RANGER-4076: Support Java 17 for build and runtime

RakeshGuptaDev:

Added support for Java 17 in both build and runtime environments, alongside support for Java 8.

## How was this patch tested?

This patch has been successfully tested and verified with Java versions 8, 11, and 17. The verification process included both manual and Docker setup. Now we are able to build and run on same jdk.

JDK 8 maven build successfully with Test-cases
mvn clean compile package install

JDK 11 maven build successfully with Test-cases
mvn clean compile package install

JDK 17 maven build successfully with Test-cases
mvn clean compile package install -Pranger-all-modules-jdk17

Validated policy enforcement for below scenario :
Scenario 1:
Successfully validated usersync, tagsync and policy enforcement(policy-condition) for HDFS, HBase and Hive services using Ranger with JDK 8 for both build and runtime environments.

Scenario 2:
Successfully validated usersync, tagsync and policy enforcement(policy-condition) for HDFS, HBase services using Ranger with JDK 17 for both build and runtime environments.


mneethiraj:
Good work @RakeshGuptaDev! One concern is that many Ranger plugins need to run in JDK version prior to 17 (JDK8 or JDK 11 or JDK15) - like HDFS, Hive, HBase. This wouldn't allow us to switch to Graal script engine, as it is only available in JDK17 onwards (https://www.graalvm.org/jdk17/docs/).

One option to consider building Ranger libraries used by plugins (agents-common, agents-audit, agents-cred, all plugins) with JDK8 and servers (admin, usersync, tagsync, KMS) with JDK17.

RakeshGuptaDev:
Hi @mneethiraj,

Thank you for your feedback. I understand your concern that Ranger plugins need to support JDK versions prior to 17, including JDK 8, JDK 11, and JDK 15.

In the current patch, I have used GraalVM version 21.3.x, which is compatible with JDK 8 (https://www.graalvm.org/21.3/docs/introduction/index.html#available-distributions). I have also successfully validated plugins like HDFS, HBase, and Hive across JDK 8 and JDK 11.

For JDK 17 and later, I have used the latest available GraalVM version.

To address this compatibility challenge, we can consider the following approach:

 **Replace Nashorn:**

- Use GraalVM version 21.3.x for JDK 8.
- Upgrade to the latest available GraalVM version for JDK 15 and later.

This approach aims to maintain compatibility across all supported JDK versions. Please share your thoughts or suggest any alternative solutions.

PR #427, opened 2024-12-06
RANGER-5021: Add timeouts to jobs in CI

kumaab:

Add timeouts to jobs in CI

## How was this patch tested?
CI Run 2 successful with timeout of 15 mins.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -33,6 +33,7 @@ on:
 jobs:
   build-8:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK 8
@@ -51,6 +52,7 @@ jobs:
 
   build-11:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v4
       - name: Set up JDK 11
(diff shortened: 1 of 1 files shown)


mneethiraj:
@kumaab  - would this change result in CI failure when build doesn't complete in 30 minutes?

kumaab:
@mneethiraj 
Yes, that's right. Should we increase the time across different stages?

mneethiraj:
> @mneethiraj Yes, that's right. Should we increase the time across different stages?

I suggest increasing overall timeout to 1 hour, to avoid unnecessary failures due to any slowness in build and testing.

kumaab:
@adoroszlai 
Could you please review this PR, thanks!

adoroszlai (commented):
LGTM, most runs seem to be around 20/20/10 minutes for `build-8`, `build-11`, `docker-build`, respectively.

However, there are some runs where `build-8` took significantly more time.  Is such difference expected?

- 60 minutes: https://github.com/apache/ranger/actions/runs/12247100043
- 43 minutes: https://github.com/apache/ranger/actions/runs/12247100740

kumaab:
Yes, with multi-threaded builds `build-8` and `build-11`, runs generally last around 20 mins but sometimes it may run longer may be due to m2 cache not being present and time it takes to populate m2. `docker-build` is faster with caches in place.

CC: @mneethiraj

adoroszlai:
Thanks @kumaab, I didn't notice cache miss.  That should explain it.

PR #462, opened 2024-12-18
RANGER-5060: checkstyle compliance updates - ranger-util module

kumaab:

  Updated `ranger-util module` sources to comply with the check-style requirements.

  ## How was this patch tested?
  `mvn clean verify` on local machine.


ranger-util/src/main/java/org/apache/ranger/common/RangerVersionInfo.java
```diff
@@ -23,113 +23,113 @@
...
-		System.out.println("Ranger " + getVersion());
-		System.out.println("Subversion " + getUrl() + " -r " + getRevision());
-		System.out.println("Compiled by " + getUser() + " on " + getDate());
-		System.out.println("From source with checksum " + getSrcChecksum());
-	}
-
+    private static Package                 myPackage;
+    private static RangerVersionAnnotation version;
+
+    private RangerVersionInfo(){
```
mneethiraj, line 33: Space before '{' missing. Shouldn't this result in build failure?

kumaab: Apparently, rules don't work for empty functions.

Like for ex, the rule does not kick in here:
```
public void doNothing(){
        // do nothing
    }
```
whereas it kicks in here:

```
public boolean doNothing(){
        return true;
    }
```

mneethiraj (approved):

PR #465, opened 2024-12-19
RANGER-5075: fix for plugin installation failures due to missing libraries

mneethiraj:
- updated plugins packaging to include libraries used during plugin installation, like commons-lang3, commons-compress
- updated dependencies for distro/pom.xml to ensure that the packaging starts only after build for all modules complete

## How was this patch tested?
- brought up Ranger and plugin services in docker setup and verified that failures seen during plugin installation (in creating key store/trust store) are no more present.


mneethiraj:
merged in master and ranger-2.6 branches

PR #483, opened 2024-12-31
RANGER-5080: Add docker support for MS SQL Server

kumaab:

Adding support for MS SQL Server - ranger database in docker.

`2019-latest` is the version used from dockerhub: https://hub.docker.com/r/microsoft/mssql-server
There were issues with `2022-latest` image with missing libraries and issues in bringing the SQL server up, so using the `2019-latest` version in this PR.

SSL is enabled in the image by default, the PR uses self-signed certs (and not certs from a CA) to setup db for `ranger` and `ranger-kms` services . 

## How was this patch tested?
The patch required testing on `amd64` compatible architecture.
CI step: `View sql server logs` shows database container logs after startup, no errors are seen.
`ranger-sqlserver` container is in running state in the logs.
`ranger-kms` completed installation with `ranger-sqlserver` successfully.


security-admin/scripts/setup.sh
```diff
@@ -746,7 +747,12 @@ update_properties() {
 	if [ "${DB_FLAVOR}" == "MSSQL" ]
 	then
 		propertyName=ranger.jpa.jdbc.url
-		newPropertyValue="jdbc:sqlserver://${DB_HOST};databaseName=${db_name}"
+		if [ "${DOCKER_ENV}" == "true" ]
```
mneethiraj: Instead of using `DOCKER_ENV` to determine whether to add `trustServerCertificate=true` to connection string, I suggest introducing a configuration (`CONNECTION_STRING_ADDITIONAL_PARAMS`)  in install.properties file.

kms/scripts/setup.sh
```diff
@@ -605,7 +606,12 @@ update_properties() {
 	if [ "${DB_FLAVOR}" == "MSSQL" ]
 	then
 		propertyName=ranger.ks.jpa.jdbc.url
-		newPropertyValue="jdbc:sqlserver://${DB_HOST};databaseName=${db_name}"
+    		if [ "${DOCKER_ENV}" == "true" ]
```
mneethiraj: Instead of using `DOCKER_ENV` to determine whether to add `trustServerCertificate=true` to connection string, I suggest introducing a configuration (`CONNECTION_STRING_ADDITIONAL_PARAMS`) in install.properties file.

security-admin/scripts/setup.sh
```diff
@@ -747,9 +748,9 @@ update_properties() {
 	if [ "${DB_FLAVOR}" == "MSSQL" ]
 	then
 		propertyName=ranger.jpa.jdbc.url
-		if [ "${DOCKER_ENV}" == "true" ]
+		if [ "${CONNECTION_STRING_ADDITIONAL_PARAMS}" != "" ]
 		then
-  			newPropertyValue="jdbc:sqlserver://${DB_HOST};databaseName=${db_name};trustServerCertificate=true;"
+  			newPropertyValue=${CONNECTION_STRING_ADDITIONAL_PARAMS}
```
mneethiraj: `newPropertyValue="jdbc:sqlserver://${DB_HOST};databaseName=${db_name};${CONNECTION_STRING_ADDITIONAL_PARAMS}`"

mneethiraj: Also, add the following to install.properties used in docker setup:
`CONNECTION_STRING_ADDITIONAL_PARAMS="trustServerCertificate=true;"`

kms/scripts/setup.sh
kumaab: Sure, this is better.

security-admin/scripts/setup.sh
kumaab: Sure, this is better.

mneethiraj (approved):

kumaab:
I'll revert changes in maven.yml added to perform testing soon.

mneethiraj (approved):

PR #487, opened 2025-01-03
RANGER-5061: security-admin module(common package): update for code readability improvementRanger 5061 common

vyommani:

Fixes all checkstyle related errors in common package for security-admin module

## How was this patch tested?

mvn clean install

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/common/db/RangerTransactionSynchronizationAdapter.java
+++ b/security-admin/src/main/java/org/apache/ranger/common/db/RangerTransactionSynchronizationAdapter.java
@@ -28,9 +28,7 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.TransactionDefinition;
-import org.springframework.transaction.TransactionStatus;
 import org.springframework.transaction.TransactionSystemException;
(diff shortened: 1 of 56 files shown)


security-admin/src/main/java/org/apache/ranger/common/db/RangerTransactionSynchronizationAdapter.java
```diff
@@ -42,22 +37,22 @@
 
 import javax.persistence.OptimisticLockException;
 
+import java.util.ArrayList;
+import java.util.List;
+
 @Component
 public class RangerTransactionSynchronizationAdapter extends TransactionSynchronizationAdapter {
```
kumaab: According to checkstyle rules, the build should fail on blank line following an open curly brace. Is it not happening?

vyommani: No it is not failing for me, i checked again may be for class declaration it is optional.

mneethiraj (approved):

security-admin/src/main/java/org/apache/ranger/common/db/RangerTransactionSynchronizationAdapter.java
mneethiraj: @kumaab  - fail on checkstyle violation is not enabled in security-admin/pom.xml.

PR #505, opened 2025-01-09
RANGER-5061: checkstyle compliance updates - security-admin org.apach…

pradeepagrawal8184:

Check-style fix for following packages of security-admin module:

org.apache.ranger.rest

## How was this patch tested?

Tested Build without any checkstyle violation in above packages.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
@@ -105,15 +105,18 @@ import java.util.stream.Collectors;
 @Scope("request")
 @Transactional(propagation = Propagation.REQUIRES_NEW)
 public class GdsREST {
-    public static final String GDS_POLICY_EXPR_CONDITION = "expression";
     private static final Logger LOG      = LoggerFactory.getLogger(GdsREST.class);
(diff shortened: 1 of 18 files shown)


security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
```diff
@@ -2096,14 +1957,123 @@ RangerPolicy updatePolicyWithModifiedGrants(RangerPolicy policy, List<RangerGran
         return policy;
     }
 
+    private Long getOrCreateDataShare(Long datasetId, Long serviceId, Long zoneId, String serviceName) throws Exception {
+        LOG.debug("==> GdsREST.getOrCreateDataShare(dataSetId={} serviceId={} zoneId={} seviceName={})", datasetId);
```
kumaab: More params need to be passed in here.

kumaab (approved):

security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
pradeepagrawal8184: Thanks, resolved now.

mneethiraj (approved):

PR #513, opened 2025-01-20
RANGER-5102: Enable append mode for audits to hdfs with config param

kumaab:
Ranger audits to HDFS currently use APPEND mode in case of errors/exceptions encountered in writing audits to HDFS destination (to prevent large number of audit files) and fallbacks to WRITE mode if unable to APPEND.

 

Add a config param to enable APPEND mode (to be used under specific circumstances).

## How was this patch tested?

Verified that the config param `xasecure.audit.destination.hdfs.file.append.enabled` enables/disables append mode in docker with hive, hadoop and ranger containers.

### DIFF
--- a/agents-audit/src/main/java/org/apache/ranger/audit/utils/AbstractRangerAuditWriter.java
+++ b/agents-audit/src/main/java/org/apache/ranger/audit/utils/AbstractRangerAuditWriter.java
@@ -45,12 +45,13 @@ import java.util.Properties;
 public abstract class AbstractRangerAuditWriter implements RangerAuditWriter {
     private static final Logger logger = LoggerFactory.getLogger(AbstractRangerAuditWriter.class);
 
-    public static final String  PROP_FILESYSTEM_DIR              = "dir";
-    public static final String  PROP_FILESYSTEM_SUBDIR           = "subdir";
-    public static final String  PROP_FILESYSTEM_FILE_NAME_FORMAT = "filename.format";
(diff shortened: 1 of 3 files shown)


rameeshm (approved):

agents-audit/src/main/java/org/apache/ranger/audit/utils/AbstractRangerAuditWriter.java
```diff
@@ -290,13 +293,13 @@ public PrintWriter createWriter() throws Exception {
         if (logWriter == null) {
             boolean appendMode = false;
 
-            // if append is supported, reuse last log file
+            // if append is supported and enabled via config param, reuse last log file
             if (reUseLastLogFile && isAppendEnabled()) {
```
mneethiraj: add condition `auditPath != null` as well:
```
  if (auditPath != null && reUseLastLogFile && isAppendEnabled()) {
```

kumaab: done, thanks.

mneethiraj (approved):

kumaab:
Verified the changes in HA setup with Hive and the below command:
```
hdfs debug recoverLease -path /ranger/audit/hive/hiveServer2/20250224/hiveServer2_ranger_audit_hostname3.log --retries 3
``` 
that when append is `true`, the audit switches to file spool and when writing to destination again it uses the same file whereas when append is `false`, the audit is switches to file spool and when writing to destination again it creates a new file.

CC: @mneethiraj @rameeshm @vperiasamy

PR #516, opened 2025-01-22
RANGER-5110: Update the FQDN of users and groups whenever it changes

eubnara:

If the FQDN of users and groups, usersync doesn't update it properly and if `ranger.usersync.deletes.enabled` is true, those users and groups becomes hidden. It should be handled properly.

## How was this patch tested?

Manual test on private internal Hadoop cluster.

### DIFF
--- a/ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
+++ b/ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
@@ -294,6 +294,13 @@ public class PolicyMgrUserGroupBuilder extends AbstractUserGroupSource implement
         if (!isStartupFlag && computeDeletes) {
             LOG.info("Computing deleted users/groups");
 
+            userCache.clear();
+            groupCache.clear();
+            groupUsersCache.clear();
+            groupNameMap.clear();
+            userNameMap.clear();
+            buildUserGroupInfo();
+
             if (MapUtils.isNotEmpty(sourceGroups)) {
                 updateDeletedGroups(sourceGroups);
             }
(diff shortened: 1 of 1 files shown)


ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -1732,15 +1714,26 @@ private void computeDeletedGroups(Map<String, Map<String, String>> sourceGroups)
...
+        for (Map.Entry<String, XGroupInfo> entry : groupCache.entrySet()) {
+            String groupName = entry.getKey();
+            XGroupInfo groupInfo = entry.getValue();
             Map<String, String> groupOtherAttrs = groupInfo.getOtherAttrsMap();
             String              groupDN         = groupOtherAttrs != null ? groupOtherAttrs.get(UgsyncCommonConstants.FULL_NAME) : null;
 
             if (StringUtils.isNotEmpty(groupDN) && !sourceGroups.containsKey(groupDN)
                     && StringUtils.equalsIgnoreCase(groupOtherAttrs.get(UgsyncCommonConstants.SYNC_SOURCE), currentSyncSource) &&
                     StringUtils.equalsIgnoreCase(groupOtherAttrs.get(UgsyncCommonConstants.LDAP_URL), ldapUrl)) {
-                if (ISHIDDEN.equals(groupInfo.getIsVisible())) {
```
eubnara, line 1743: This is bug introduced by typo. Fix it also in this PR.

kimsehwan96:
+1 . Is there any plan to review and merge this? @mneethiraj 

We hit this bug in production (Ranger 2.7.0 + LDAP sync, `deletes.enabled=true`).
A user was moved to another OU, and usersync marked the user as hidden and removed all group memberships.

PR #518, opened 2025-01-26
RANGER-5116: updated Ranger plugin to support configurations to initialize kerberos identity

mneethiraj:

updated Ranger plugin to support initialization of Kerberos identity with following configurations 

```
ranger.plugin.myservice.kerberos.initialize=true
ranger.plugin.myservice.kerberos.login.type=keytab
ranger.plugin.myservice.kerberos.keytab.principal=myservice/_HOST@REALM
ranger.plugin.myservice.kerberos.keytab.file=/etc/keytabs/myservice.keytab
```

## How was this patch tested?

Build with tests completed successfully.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -144,8 +144,8 @@ public class RangerBasePlugin {
             String kerbLoginType = pluginConfig.get(pluginConfig.getPropertyPrefix() + ".kerberos.login.type");
 
             if (StringUtils.equalsIgnoreCase(kerbLoginType, "keytab")) {
-                String kerbPrincipal = pluginConfig.get(pluginConfig.getPropertyPrefix() + ".keytab.principal");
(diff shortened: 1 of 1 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -135,6 +138,55 @@ public RangerBasePlugin(RangerPluginConfig pluginConfig) {
         setIsFallbackSupported(pluginConfig.getBoolean(pluginConfig.getPropertyPrefix() + ".is.fallback.supported", false));
         setServiceAdmins(serviceAdmins);
 
+        boolean initKerb = pluginConfig.getBoolean(pluginConfig.getPropertyPrefix() + ".kerberos.initialize", false);
```
kumaab: How about keeping the ".kerberos" suffixes as constants in the same file, that way they can be reused at different places.

mneethiraj: Intoduced a variable `kerbPrefix` to minimize repeated configuration prefix.

````
String  kerbPrefix = pluginConfig.getPropertyPrefix() + ".kerberos";
````

kumaab (approved):

PR #521, opened 2025-01-29
RANGER-5122: fixed path for images

developster:

Load images from the correct path. This bug appears only when context path of the ranger web app is changed to something else than the default root (/).

Also updated setup.sh to allow changing the "ranger.contextName" from install.properties.

Fixes RANGER-5122.

## How was this patch tested?

Tested manually.


fimugdha:
@developster such type of fix is not required, as only Ranger application is executed from Ranger Embedded Tomcat and Ranger server uses dedicated port to run.

This will force every user to change their Ranger Admin urls getting used for communicating to Ranger Rest APIs.

However this fix is breaking Ranger Admin url while accessing it with Knox gateway.

cc @pradeepagrawal8184 / @mneethiraj / @kumaab

PR #525, opened 2025-02-03
RANGER-5126: Add coverage for testing ranger upgrades

kumaab:

Adding a workflow for ranger upgrade path testing.

## How was this patch tested?

CI


kumaab:
Re-opened here: #610

PR #531, opened 2025-02-12
RANGER-5131: remove finalize from policy engine

vyommani:

Removal of finalize() method from PolicyEngine class

As part of our efforts to ensure compatibility with future JDK versions, this code change removes the deprecated finalize() method from the PolicyEngine class. The finalize() method has been deprecated since JDK 9 and is slated for removal in future JDK versions.

Impact Analysis

A thorough analysis was conducted to assess the impact of removing the finalize() method. The results indicate that there is no adverse impact on the functionality of the PolicyEngine class.

Existing Cleanup Mechanisms

It's worth noting that we already have a cleanup() method in the RangerBasePlugin class, which allows users to release resources held by the PolicyEngine. This method provides a manual way to perform cleanup operations, ensuring that resources are properly released when needed.

## How was this patch tested?

mvn clean install

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/contextenricher/RangerAbstractContextEnricher.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/contextenricher/RangerAbstractContextEnricher.java
@@ -300,13 +300,4 @@ public abstract class RangerAbstractContextEnricher implements RangerContextEnri
 
         return ret;
     }
-
-    @Override
-    protected void finalize() throws Throwable {
-        try {
-            cleanup();
-        } finally {
-            super.finalize();
-        }
-    }
 }
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/PolicyEngine.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/PolicyEngine.java
@@ -338,15 +338,6 @@ public class PolicyEngine {
         return toString(new StringBuilder()).toString();
     }
 
-    @Override
-    protected void finalize() throws Throwable {
-        try {
-            cleanup();
-        } finally {
-            super.finalize();
-        }
-    }
-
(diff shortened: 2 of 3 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/contextenricher/RangerAbstractContextEnricher.java
```diff
@@ -300,13 +300,4 @@ public Properties readProperties(String fileName) {
 
         return ret;
     }
-
```
mneethiraj, line 303: Is it okay to simply remove finalize() methods, without an alternate method to execute the cleanup performed currently?

vyommani: Removing the finalize() method without an alternative cleanup method is not advisable. However, my analysis reveals that we already have robust cleanup mechanisms in place.
Upon reviewing the Ranger source code, I found that the PolicyEngine is referenced in the following classes:
    
    RangerPolicyAdminImpl
    RangerPolicyEngineImpl

Both RangerPolicyAdminImpl and RangerPolicyEngineImpl have a releaseResources(boolean isForced) method, which calls the preCleanup(isForced) method on the PolicyEngine. This preCleanup method performs the following cleanup operations:
    
    Calls policyRepository.preCleanup()
    Calls tagPolicyRepository.preCleanup()
    Calls zonePolicyRepositories.preCleanup()

These preCleanup methods on the repository iterate through the Context Enrichers and call preCleanup on individual instances.

Upon further review, I noticed that the setPolicies method in RangerBasePlugin calls the releaseResources method on RangerPolicyEngineImpl. This ensures that resources held by the old policy engine are properly released before updating to the new policy.

Given that we already have alternative cleanup method (releaseResources), I believe we can safely remove the finalize() method. Please review my findings and let me know if I've missed any code paths.

PR #535, opened 2025-02-20
RANGER-5151: fix for error while writing audit logs to HDFS

mneethiraj:

Updated to call `closeFileIfNeeded()` only from `getLogFileStream()`, to avoid one thread closing the file which is being written by another thread.

## How was this patch tested?

Configured plugin to rollover audit log every 3 minutes, generated audits for a long duration and verified that the errors reported in the JIRA are no more seen.

```
  xasecure.audit.destination.hdfs.file.rollover.enable.periodic.rollover=true
  xasecure.audit.destination.hdfs.file.rollover.periodic.rollover.check.sec=180
  xasecure.audit.destination.hdfs.file.rollover.period=3m
```


rameeshm (approved):

kumaab (commented):
LGTM.

PR #538, opened 2025-02-24
RANGER-5148: Fix redundant role cache modify triggered by role versio…

Gooch0922:
…n changes in concurrent scenarios.

(Please fill in changes proposed in this fix. Create an issue in ASF JIRA before opening a pull request and
set the title of the pull request which starts with
the corresponding JIRA issue number. (e.g. RANGER-XXXX: Fix a typo in YYY))

## How was this patch tested?

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/common/RangerRoleCache.java
+++ b/security-admin/src/main/java/org/apache/ranger/common/RangerRoleCache.java
@@ -46,6 +46,7 @@ public class RangerRoleCache {
     RangerRoleCacheWrapper roleCacheWrapper;
 
     private RangerRoleCache() {
+        roleCacheWrapper         = new RangerRoleCacheWrapper();
         RangerAdminConfig config = RangerAdminConfig.getInstance();
 
         waitTimeInSeconds = config.getInt("ranger.admin.policy.download.cache.max.waittime.for.update", MAX_WAIT_TIME_FOR_UPDATE);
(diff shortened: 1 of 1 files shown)


security-admin/src/main/java/org/apache/ranger/common/RangerRoleCache.java
```diff
@@ -69,21 +70,42 @@ public static RangerRoleCache getInstance() {
...
+            if (roleCacheWrapper.getRolesVersion() < rangerRoleVersionInDB) {
+                boolean lockResult = false;
+                try {
+                    lockResult = lock.tryLock(waitTimeInSeconds, TimeUnit.SECONDS);
+
+                    if (lockResult) {
+                        if (roleCacheWrapper.getRolesVersion() < rangerRoleVersionInDB) {
+                            ret = roleCacheWrapper.getLatestRangerRoles(
+                                    serviceName, roleDBStore, lastKnownRoleVersion, rangerRoleVersionInDB);
+                        }
```
mneethiraj, line 88: `ret` should be updated in `else` clause, right?

```
if (roleCacheWrapper.getRolesVersion() < rangerRoleVersionInDB) {
  ret = roleCacheWrapper.getLatestRangerRoles(serviceName, roleDBStore, lastKnownRoleVersion, rangerRoleVersionInDB);
} else {
  ret = roleCacheWrapper.getRoles();
}
```

security-admin/src/main/java/org/apache/ranger/common/RangerRoleCache.java
```diff
@@ -46,6 +46,7 @@ public class RangerRoleCache {
     RangerRoleCacheWrapper roleCacheWrapper;
```
mneethiraj: Consider marking `roleCacheWrapper` as `final`.

security-admin/src/main/java/org/apache/ranger/common/RangerRoleCache.java
Gooch0922: The variable "ret" has already been initialized and the cache has been retrieved once at line 76. There is no need to retrieve the cache repeatedly, right?

mneethiraj: `ret` initialized in line 76 is an older version, hence the `if` block at line 78 is entered. Consider 2 threads:
1. thread-1: `ret` is initialized to version=100; acquired lock at line 81
2. thread-2: `ret` is initialized to version=100; waiting for the lock at line 81
3. thread-1: refreshes `roleCacheWrapper` and `ret` to the latest version, say 101, at line 85; releases the lock
4. thread-2: acquires the lock, finds that `roleCacheWrapper` has the lastest version; skips line 85; leaves `ret` at version 100

The call from thread-2 will return version 100; the expected version is 101. Makes sense?

PR #545, opened 2025-03-20
RANGER-5134: Fix policy label processing in concurrent sessions

kumaab:

Fixes concurrent processing of policy labels in separate transactions. 
Added check to verify policy exists before policy to policy label mapping is created.

## How was this patch tested?
Tested these scenarios in concurrent sessions successfully:
```
- added same non-existing label to the same policy. 
- deleted a label from one policy in one session, added a non-existing label to the same policy in another session, one of them gets applied.
- updated label for a policy and deleted the same policy in another session, both sessions don't see the deleted policy.
- added same labels to different policies concurrently.

### CODE REVIEW

File: security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java [hunk 1]
```diff
@@ -6741,15 +6741,28 @@ public void run() {
             getOrCreateLabel();
         }
 
+        private boolean doesPolicyExist(XXPolicy xPolicy) {
+            XXPolicy     xxExisting = daoMgr.getXXPolicy().getById(xPolicy.getId());
+            LOG.info("Found existing policy {}", xxExisting);
+            RangerPolicy existing   = policyService.getPopulatedViewObject(xxExisting);
```
mneethiraj: Is it necessary to get `RangerPolicy` from `xxExisting`? Why not `return xxExisting != null;`

security-admin/src/main/resources/META-INF/jpa_named_queries.xml
```diff
@@ -120,7 +120,10 @@
                 order by obj.policyId, obj.id
                         </query>
         </named-query>
-
+		<named-query name="XXPolicyLabelMap.findByPolicyIdAndLabelId">
+			<query>SELECT obj FROM XXPolicyLabelMap obj WHERE obj.id = :policyLabelId AND obj.policyId = :policyId
```
mneethiraj: `obj.id = :policyLabelId` ==> `obj.policyLabelId = :policyLabelId`

security-admin/src/main/java/org/apache/ranger/common/db/RangerTransactionSynchronizationAdapter.java
```diff
@@ -206,8 +205,18 @@ private void runRunnables(final List<Runnable> runnables, final boolean isParent
...
+
+                    if (isParentTransactionCommitted) {
+                        if (!isThisTransactionCommitted) {
+                            LOG.info("Failed to commit runnable:[{}]. Will retry!", runnable);
+                            count ++;
+                        } else {
+                            LOG.debug("Committed runnable:[{}].", runnable);
+                        }
+                    }
+                } while (isParentTransactionCommitted && !isThisTransactionCommitted && count < 5);
```
mneethiraj: Why is it necessary to limit the number of retries? The right fix would be to update Runnable implementations to return success when the objects referenced (policy, label, ..) don't exist in the database any more.

kumaab: I'll remove the test code and update, it was not intended for review.

security-admin/src/main/resources/META-INF/jpa_named_queries.xml
kumaab: Ah, this is why database exceptions were seen, thanks.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -6760,14 +6769,25 @@ private void getOrCreateLabel() {
...
-                xxPolicyLabelMap.setPolicyId(xPolicy.getId());
-                xxPolicyLabelMap.setPolicyLabelId(xxPolicyLabel.getId());
+                    xxPolicyLabelMap.setPolicyId(xPolicy.getId());
+                    xxPolicyLabelMap.setPolicyLabelId(xxPolicyLabel.getId());
 
-                xxPolicyLabelMap = rangerAuditFields.populateAuditFieldsForCreate(xxPolicyLabelMap);
+                    xxPolicyLabelMap = rangerAuditFields.populateAuditFieldsForCreate(xxPolicyLabelMap);
 
-                daoMgr.getXXPolicyLabelMap().create(xxPolicyLabelMap);
+                    if (doesPolicyExist(xPolicy)) {
```
mneethiraj: If the policy doesn't exist, getOrCreateLabel() method should return at the earliest i.e. at line 6750. No label should be created in such case (line 6764.

kumaab: makes sense.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -6741,33 +6741,53 @@ public void run() {
...
+
+                        xxPolicyLabel = rangerAuditFields.populateAuditFieldsForCreate(xxPolicyLabel);
+                        xxPolicyLabel = daoMgr.getXXPolicyLabels().create(xxPolicyLabel);
+                    }
                 }
-            }
 
-            if (xxPolicyLabel != null) {
-                XXPolicyLabelMap xxPolicyLabelMap = new XXPolicyLabelMap();
+                if (xxPolicyLabel != null) {
```
mneethiraj: Is `if` at line 6772 necessary? `xxPolicyLabel` should not be null here.

security-admin/src/main/java/org/apache/ranger/common/db/RangerTransactionSynchronizationAdapter.java
```diff
@@ -206,8 +197,17 @@ private void runRunnables(final List<Runnable> runnables, final boolean isParent
                             LOG.debug("Failed to commit TransactionService transaction, throwable:[{}]", String.valueOf(e));
                         }
                     }
-                }
-                while (isParentTransactionCommitted && !isThisTransactionCommitted);
+
+                    isThisTransactionCommitted = result == runnable;
```
mneethiraj, line 201: Is moving of lines 200 - 209 outside the `try` block necessary?

kumaab: I think it's slightly better to keep it outside since `log.info` says there will be a retry once all exceptions have been caught from either try blocks. `log.info` might be missed if this is kept within the try block.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
kumaab: Sure, will remove it, it seems redundant.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -6741,33 +6741,53 @@ public void run() {
...
-                xxPolicyLabel = daoMgr.getXXPolicyLabels().findByName(policyLabel);
+            if (doesPolicyExist(xPolicy)) {
+                LOG.debug("Searching for policyLabel: {}", policyLabel);
+                XXPolicyLabel xxPolicyLabel = daoMgr.getXXPolicyLabels().findByName(policyLabel);
+                LOG.debug("Search returned: {}", xxPolicyLabel);
 
                 if (xxPolicyLabel == null) {
-                    xxPolicyLabel = new XXPolicyLabel();
+                    LOG.debug("Searching again for policyLabel: {}", policyLabel);
+                    xxPolicyLabel = daoMgr.getXXPolicyLabels().findByName(policyLabel);
```
kumaab: @mneethiraj 
do you think searching again is redundant here and can be avoided ?

mneethiraj: The second attempt to retrieve from the database is indeed reduncent, since the result will be same as the first attempt - given they both are executing the same transaction. The second attempt can be removed.

mneethiraj (approved):

kumaab:
Stress test successful:
- Updated the same policy across 2 different sessions with 50k labels. No errors were seen.

PR #546, opened 2025-03-24
RANGER-5174: fix for script engine instantiation failure

mneethiraj:

Updated script engine creator implementations to use the classloader that loaded the creator instance, while creating script engine instances, instead of current thread's classloader.

## How was this patch tested?

Verified that Trino plugin successfully evaluates expressions in row-filter policies.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/util/GraalScriptEngineCreator.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/util/GraalScriptEngineCreator.java
@@ -72,7 +72,7 @@ public class GraalScriptEngineCreator implements ScriptEngineCreator {
         ScriptEngine ret = null;
 
         if (clsLoader == null) {
-            clsLoader = Thread.currentThread().getContextClassLoader();
+            clsLoader = getDefaultClassLoader();
         }
 
         try {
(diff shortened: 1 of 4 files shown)


vyommani (approved):
The code changes look good to me. I noticed that all three script engine creators (Graal, JavaScript, and Nashorn) implement the ScriptEngineCreator interface.

I was wondering if there's a specific reason for not implementing the 'getDefaultClassloader' method for all three ScriptEngineCreator implementations, instead relying on the default implementation for all three?

mneethiraj:
> The code changes look good to me. I noticed that all three script engine creators (Graal, JavaScript, and Nashorn) implement the ScriptEngineCreator interface.
> 
> I was wondering if there's a specific reason for not implementing the 'getDefaultClassloader' method for all three ScriptEngineCreator implementations, instead relying on the default implementation for all three?

I added the `default` method in the interface to avoid code duplication in all 3 implementations. Do you see any issues with using this approach?

rameeshm (approved):

vyommani:
> I added the `default` method in the interface to avoid code duplication in all 3 implementations. Do you see any issues with using this approach?

No, I completely agree with you - avoiding code duplication is a top priority, and your solution effectively achieves that.

PR #554, opened 2025-04-02
RANGER-5178: Update callers of executeOnTransactionCommit to check for existence of objects

kumaab:

`executeOnTransactionCommit` in `RangerTransactionSynchronizationAdapter.java` is called multiple places for doing post transaction commit tasks in a subsequent transaction following the parent transaction.
 
The PR adds checks for appropriate objects referenced in classes `PolicyRefUpdater` and `RoleRefUpdater`.

## How was this patch tested?
Pending.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
@@ -516,42 +516,50 @@ public class PolicyRefUpdater {
             return ret;
         }
 
+        private boolean doesPolicyExist(XXPolicy xPolicy) {
+            return daoMgr.getXXPolicy().getById(xPolicy.getId()) != null;
+        }
+
         private void createPolicyAssociation(Long id, String name) {
             LOG.debug("===> PolicyPrincipalAssociator.createPolicyAssociation(policyId={}, type={}, name={}, id={})", xPolicy.getId(), type.name(), name, id);
 
-            switch (type) {
-                case USER: {
-                    XXPolicyRefUser xPolUser = rangerAuditFields.populateAuditFields(new XXPolicyRefUser(), xPolicy);
+            if (doesPolicyExist(xPolicy)) {
+                switch (type) {
+                    case USER: {
+                        XXPolicyRefUser xPolUser = rangerAuditFields.populateAuditFields(new XXPolicyRefUser(), xPolicy);
 
-                    xPolUser.setPolicyId(xPolicy.getId());
-                    xPolUser.setUserId(id);
-                    xPolUser.setUserName(name);
+                        xPolUser.setPolicyId(xPolicy.getId());
+                        xPolUser.setUserId(id);
+                        xPolUser.setUserName(name);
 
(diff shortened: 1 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
```diff
@@ -516,42 +516,50 @@ private Long createPrincipal(String user) {
...
+            return daoMgr.getXXPolicy().getById(xPolicy.getId()) != null;
+        }
+
         private void createPolicyAssociation(Long id, String name) {
             LOG.debug("===> PolicyPrincipalAssociator.createPolicyAssociation(policyId={}, type={}, name={}, id={})", xPolicy.getId(), type.name(), name, id);
 
-            switch (type) {
-                case USER: {
-                    XXPolicyRefUser xPolUser = rangerAuditFields.populateAuditFields(new XXPolicyRefUser(), xPolicy);
+            if (doesPolicyExist(xPolicy)) {
```
mneethiraj, line 526: In addition to checking for existence of policy, shouldn't existence of user/group/role be checked as well?

```
if (doesPolicyExist(xPolicy) && doesPrincipalExist(name, type)) {
```

security-admin/src/main/java/org/apache/ranger/biz/RoleRefUpdater.java
```diff
@@ -345,45 +345,63 @@ private Long createPrincipal(String user) {
...
+            return false;
+        }
+
         private void createRoleAssociation(Long id, String name) {
             LOG.debug("===> RolePrincipalAssociator.createRoleAssociation(roleId={}, type={}, name={}, id={})", roleId, type.name(), name, id);
 
-            switch (type) {
-                case USER: {
-                    XXRoleRefUser xRoleRefUser = rangerAuditFields.populateAuditFieldsForCreate(new XXRoleRefUser());
+            if (doesPrincipalExist(name, type)) {
```
mneethiraj: In addition to checking for existence of principal (user/group/role), shouldn't existence of role be checked as well?

```
if (doesRoleExist(id) && doesPrincipalExist(name, type)) {
```

security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
kumaab: Check for principal is done as a pre-check from the caller of this function in the same thread, for reference:
`Long id = createOrGetPrincipal(isAdmin);` in `boolean doAssociate(boolean isAdmin) {`

security-admin/src/main/java/org/apache/ranger/biz/RoleRefUpdater.java
kumaab: Actually figured out `doesPrincipalExist(name, type)` is redundant as it happens in the caller `Long id = createOrGetPrincipal(isAdmin);` in `boolean doAssociate(boolean isAdmin) {`, check for `roleId` should be done instead, will update the PR.

mneethiraj (approved):

PR #555, opened 2025-04-04
RANGER-5185: fix potential NPE in references to RangerBasePlugin.policyEngine

mneethiraj:

Updated RangerBasePlugin to ensure that references to policyEngine member doesn't cause NPE.

## How was this patch tested?

All unit tests pass successfully.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -611,7 +611,7 @@ public class RangerBasePlugin {
             } else {
                 LOG.warn("Leaving current policy engine as-is");
                 LOG.warn("Policies are not saved to cache. policyVersion in the policy-cache may be different than in Ranger-admin, even though the policies are the same!");
-                LOG.warn("Ranger-PolicyVersion:[{}], Cached-PolicyVersion:[{}]", policies != null ? policies.getPolicyVersion() : -1L, this.policyEngine != null ? this.policyEngine.getPolicyVersion() : -1L);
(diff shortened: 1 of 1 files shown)


rameeshm (approved):

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -883,14 +883,11 @@ public RangerRoles getRangerRoles() {
     }
 
     public Set<RangerRole> getRangerRoleForPrincipal(String principal, String type) {
-        Set<RangerRole>          ret         = new HashSet<>();
-        Set<RangerRole>          rangerRoles = null;
+        Set<RangerRole>          ret          = new HashSet<>();
+        RangerPolicyEngine       policyEngine = this.policyEngine;
```
vyommani, line 887: Do we need this change ?. I can see  getRangerRole is already checking policyEngine  for null as follows
 RangerPolicyEngine policyEngine = this.policyEngine;

        if (policyEngine != null) {
            return policyEngine.getRangerRoles();
        }

mneethiraj: @vyommani  - which line number are you referring to?  Local variable`policyEngine` (line 887) is referenced later in line 893. Hence it is necessary to check for null value here.

vyommani: Yes, you're right. Upon reviewing the code changes, I realized I was only focusing on the modified code.

vyommani (approved):
Looks OK to me.

PR #556, opened 2025-04-10
RANGER-5169: Error while migrating Masterkey from older format to exter…

vikaskr22:
…nal key store

Initialisation of metadata in RangerMasterkey (like message digest, algorithm etc) was happening selectively for generateMasterKey() methods only. But to migrate key from DB to any external keyStore,  generateMasterKey() is not called and hence these metadata remains null.

Fix is to invoke init() method from RangerMasterkey constructor itself so that metadata remains available for all invocations.

## How was this patch tested?

Complete Ranger build, Unit Testing, created new docker image of KMS and performed basic key operations.


dhavalshah9131 (approved):
Changes looks fine.

PR #569, opened 2025-05-01
RANGER-5205: Liquibase based Zero-Downtime-Upgrade framework for Rang…

fateh288:
…er/KMS

A new framework using liquidate to support ZDU

## How was this patch tested?

Unit tests have been added
Tested in a private cluster
WIP: Testing in docker setup

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStoreProvider.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStoreProvider.java
@@ -66,7 +66,7 @@ public class RangerKeyStoreProvider extends KeyProvider {
     private static final String CREDENTIAL_PATH              = "ranger.ks.jpa.jdbc.credential.provider.path";
     private static final String MK_CREDENTIAL_ALIAS          = "ranger.ks.masterkey.credential.alias";
     private static final String DB_CREDENTIAL_ALIAS          = "ranger.ks.jpa.jdbc.credential.alias";
-    private static final String DB_PASSWORD                  = "ranger.ks.jpa.jdbc.password";
+    public static final String DB_PASSWORD                  = "ranger.ks.jpa.jdbc.password";
     private static final String HSM_ENABLED                  = "ranger.ks.hsm.enabled";
     private static final String HSM_PARTITION_PASSWORD_ALIAS = "ranger.ks.hsm.partition.password.alias";
     private static final String HSM_PARTITION_PASSWORD       = "ranger.ks.hsm.partition.password";
@@ -79,6 +79,9 @@ public class RangerKeyStoreProvider extends KeyProvider {
     private static final String AZURE_CLIENT_SECRET_ALIAS    = "ranger.kms.azure.client.secret.alias";
     private static final String AZURE_CLIENT_SECRET          = "ranger.kms.azure.client.secret";
     private static final String AWS_KMS_ENABLED              = "ranger.kms.awskms.enabled";
+    public static final String DB_URL                        = "ranger.ks.jpa.jdbc.url";
+    public static final String DB_USER                       = "ranger.ks.jpa.jdbc.user";
+    public static final String DB_DRIVER                     = "ranger.ks.jpa.jdbc.driver";
     private static final String AWS_CLIENT_SECRETKEY_ALIAS   = RangerAWSKMSProvider.AWS_CLIENT_SECRETKEY + ".alias";
     private static final String AWS_CLIENT_SECRETKEY         = RangerAWSKMSProvider.AWS_CLIENT_SECRETKEY;
     private static final String TENCENT_KMS_ENABLED          = "ranger.kms.tencentkms.enabled";
--- a/liquibase-database-upgrade/changelogs/kms/db/postgres/create_trigger_kmsencoded_postgres.sql
+++ b/liquibase-database-upgrade/changelogs/kms/db/postgres/create_trigger_kmsencoded_postgres.sql
@@ -1,0 +1,40 @@
+-- Licensed to the Apache Software Foundation (ASF) under one or more
+-- contributor license agreements.  See the NOTICE file distributed with
+-- this work for additional information regarding copyright ownership.
+-- The ASF licenses this file to You under the Apache License, Version 2.0
+-- (the "License"); you may not use this file except in compliance with
+-- the License.  You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+--liquibase formatted sql
+
+-- Create a trigger function
+CREATE OR REPLACE FUNCTION copy_data_trigger_function()
+RETURNS TRIGGER AS
+'
+BEGIN
+     -- If column_a is updated and is not null, update column_b
+     IF NEW.kms_encoded IS NOT NULL THEN
+          NEW.kms_encoded_value := NEW.kms_encoded;
+     END IF;
+     -- If column_b is updated and is not null, update column_a
+     IF NEW.kms_encoded_value IS NOT NULL THEN
+            NEW.kms_encoded := NEW.kms_encoded_value;
+     END IF;
+     RETURN NEW;
+END;
+'
+LANGUAGE plpgsql;
+
+-- Create a trigger
+CREATE TRIGGER copy_data_trigger
+BEFORE INSERT OR UPDATE ON ranger_keystore
+FOR EACH ROW
+EXECUTE PROCEDURE copy_data_trigger_function();
--- a/liquibase-database-upgrade/changelogs/kms/db/test_3.x/sleep.xml
+++ b/liquibase-database-upgrade/changelogs/kms/db/test_3.x/sleep.xml
@@ -1,0 +1,37 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
+        xmlns:pro="http://www.liquibase.org/xml/ns/pro"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+        http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-latest.xsd">
+
+    <changeSet id="6_sleep" author="fateh">
+        <sql dbms="postgresql" >
+            select pg_sleep(60);
+        </sql>
+        <sql dbms="mysql" >
+            select SLEEP(60);
+        </sql>
+        <rollback>empty</rollback>
+    </changeSet>
+
+</databaseChangeLog>
--- a/liquibase-database-upgrade/scripts/README.md
+++ b/liquibase-database-upgrade/scripts/README.md
@@ -1,0 +1,66 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+# KMS
+## Sample exports and commands
+```
+export JAVA_HOME=/usr/java/jdk1.8.0
+export RANGER_KMS_HOME=/var/lib/ranger-kms
+#get process folder for kms: ps -ef | grep ranger-kms
+export RANGER_KMS_CONF=/var/run/process/1546336381-ranger_kms-RANGER_KMS_SERVER/conf
+export SQL_CONNECTOR_JAR=/usr/share/java/postgresql-connector-java.jar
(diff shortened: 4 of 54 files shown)


liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseDBUtilsMain.java
```diff
@@ -0,0 +1,108 @@
...
+                            LOG.info("isFinalizeComplete={}", Constants.RETURN_CODE_SUCCESS);
+                            response.put("op_status", Constants.RETURN_CODE_SUCCESS);
+                            response.put("isFinalizeComplete", Constants.RETURN_CODE_SUCCESS);
+                        } else {
+                            LOG.info("isFinalizeComplete={}", Constants.RETURN_CODE_FAILURE);
+                            response.put("op_status", Constants.RETURN_CODE_SUCCESS);
+                            response.put("isFinalizeComplete", Constants.RETURN_CODE_FAILURE);
+                        }
+                        break;
+//                    case "releaseLock":
```
dhavalshah9131: Can we shift this to TODO?

liquibase-database-upgrade/changelogs/kms/db/postgres/create_trigger_kmsencoded_postgres.sql
```diff
@@ -0,0 +1,40 @@
...
+
+--liquibase formatted sql
+
+-- Create a trigger function
+CREATE OR REPLACE FUNCTION copy_data_trigger_function()
+RETURNS TRIGGER AS
+'
+BEGIN
+     -- If column_a is updated and is not null, update column_b
+     IF NEW.kms_encoded IS NOT NULL THEN
```
dhavalshah9131: To avoid unintended overwrites and to make trigger logic more robust I think it will be better to check 

 -- If kms_encoded is set and kms_encoded_value is null, copy to kms_encoded_value
    IF NEW.kms_encoded IS NOT NULL AND NEW.kms_encoded_value IS NULL THEN

liquibase-database-upgrade/changelogs/kms/db/postgres/create_trigger_kmsencoded_postgres.sql
```diff
@@ -0,0 +1,40 @@
...
+          NEW.kms_encoded_value := NEW.kms_encoded;
+     END IF;
+     -- If column_b is updated and is not null, update column_a
+     IF NEW.kms_encoded_value IS NOT NULL THEN
```
dhavalshah9131: To avoid unintended overwrites and to make trigger logic more robust I think it will be better to check
-- If kms_encoded_value is set and kms_encoded is null, copy to kms_encoded
    IF NEW.kms_encoded_value IS NOT NULL AND NEW.kms_encoded IS NULL THEN

liquibase-database-upgrade/changelogs/kms/db/test_3.x/sleep.xml
```diff
@@ -0,0 +1,37 @@
...
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd
+        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
+        http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-latest.xsd">
+
+    <changeSet id="6_sleep" author="fateh">
+        <sql dbms="postgresql" >
+            select pg_sleep(60);
+        </sql>
+        <sql dbms="mysql" >
```
dhavalshah9131, line 31: Thought out the implementation it seems we are working on postgressql and reason of adding mysql here?

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseCommandResultsUtil.java
```diff
@@ -0,0 +1,46 @@
...
+        boolean res = (Boolean) tagExistsResult.getResult("tagExistsResult");
+        LOG.info("doesTagExist({})={}", tagExistsResult, res);
+        return res;
+    }
+
+    public static boolean isUpdateSuccessful(CommandResults updateResult) {
+    /*
+    TODO: revisit this
+     */
+        boolean operationInfoResult = updateResult.getResult("operationInfo") != null &&
```
dhavalshah9131: No null check for updateResult.
Can raise Null pointer exception here.

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseCommandResultsUtil.java
```diff
@@ -0,0 +1,46 @@
...
+                ((OperationInfo) updateResult.getResult("operationInfo")).getOperationOutcome().equals("success");
```
dhavalshah9131: No type-checking before casting from Object to OperationInfo, which could lead to ClassCastException .

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseCommandResultsUtil.java
```diff
@@ -0,0 +1,46 @@
...
+        boolean updateReportResult = updateResult.getResult("updateReport") != null && (((UpdateReportParameters) updateResult.getResult("updateReport")).getSuccess());
```
dhavalshah9131: No type-checking before casting from Object to UpdateReportParameters, which could lead to ClassCastException .

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibasePropertiesFactory.java
```diff
@@ -0,0 +1,118 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(LiquibasePropertiesFactory.class);
+
+    Map<String, IConfigProvider> configProviders;
+
+    @Autowired
+    public LiquibasePropertiesFactory(@Lazy Map<String, IConfigProvider> configProviders) {
+        this.configProviders = configProviders;
+    }
+
+    public String getUrl(String serviceName) {
```
dhavalshah9131, line 38: Cleaner and more concise version 

```
public String getUrl(String serviceName) {
    IConfigProvider provider = configProviders.get(serviceName);
    if (provider == null) {
        LOG.error("ConfigProvider for service {} is null", serviceName);
        return "";
    }
    return provider.getUrl();
}
```

Also same for below methods

liquibase-database-upgrade/src/test/java/org/apache/ranger/db/upgrade/FileDBForTest.java
```diff
@@ -0,0 +1,186 @@
...
+        return false; // String not found
+    }
+
+    public static void tag(String filePath, String tagName) {
+        createFileIfNotExists(filePath);
+        if (isFileEmpty(filePath)) {
+            List<String> lines = new ArrayList<>();
+            lines.add(",,," + tagName);
+            writeLines(filePath, lines, true);
+            //writeLineToFile(filePath,",,,"+tagName);
```
dhavalshah9131: Remove dead code

dhavalshah9131 (commented):
Please review the comments

liquibase-database-upgrade/changelogs/kms/db/test_3.x/sleep.xml
fateh288: All the changesets/changelogs I have added in this PR are for demonstration purposes only to showcase what is possible in liquibase and the features it provides.
This particular changeset I am trying to demonstrate that a single changeset with a unique id and author can be used for multiple databases if we specify the dbms attribute.
Here sleep.xml has a changeset with changeSet id="6_sleep" author="fateh" which is used to specify behavior across multiple databases i.e. postgres as well as mysql. 
This can drastically help reduce the number of changesets one has to keep track of since single changeset can now be used for multiple dbs

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseDBUtilsMain.java
fateh288: Addressed

liquibase-database-upgrade/changelogs/kms/db/postgres/create_trigger_kmsencoded_postgres.sql
fateh288: Addressed

liquibase-database-upgrade/changelogs/kms/db/postgres/create_trigger_kmsencoded_postgres.sql
fateh288: Addressed

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseCommandResultsUtil.java
fateh288: @dhavalshah9131 
If updateResult is indeed null somehow, I think it is ok that an exception is thrown. 
I did give it a thought on whether we should return true or false by default if updateResult is null but it can go either ways and a wrong return value can result in incorrect logic being applied in the next steps which we dont want.  
If there is a more elegant way to handle such scenarios I will be happy to incorporate the same. Let me know your suggestions

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseCommandResultsUtil.java
fateh288: Addressed

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibaseCommandResultsUtil.java
fateh288: Addressed

liquibase-database-upgrade/src/main/java/org/apache/ranger/db/upgrade/LiquibasePropertiesFactory.java
fateh288: Thanks !!
Incorporated

liquibase-database-upgrade/src/test/java/org/apache/ranger/db/upgrade/FileDBForTest.java
fateh288: Addressed

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStoreProvider.java
```diff
@@ -66,7 +66,7 @@ public class RangerKeyStoreProvider extends KeyProvider {
     private static final String CREDENTIAL_PATH              = "ranger.ks.jpa.jdbc.credential.provider.path";
     private static final String MK_CREDENTIAL_ALIAS          = "ranger.ks.masterkey.credential.alias";
     private static final String DB_CREDENTIAL_ALIAS          = "ranger.ks.jpa.jdbc.credential.alias";
-    private static final String DB_PASSWORD                  = "ranger.ks.jpa.jdbc.password";
+    public static final String DB_PASSWORD                  = "ranger.ks.jpa.jdbc.password";
```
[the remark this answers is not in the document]
fateh288: @dhavalshah9131 @vikaskr22 @spolavarapu1 
Do you think this is a valid concern from copilot?
I don't think so but let me know your thoughts

vikaskr22:
@fateh288 , thanks for the implementation. 

Following is the review comments on design part, not on implementation.

As per one of the discussion with you and my understanding, consider following scenario:

There is a change in the format of the data in one particular column. Say Table1.Col1, as per current approach, during schema upgrade, a new column  (say col2) of required type will be created and older data will be copied to new column after transformation. Now we may have two scenario:
1. New binaries not yet applied, means, read/write will continue to happen with old column. How you are handling the case, where a column is being updated by the application after copying from col1-> col2 ? If we are using a trigger on each data modification that will copy  to new column. Then, we have multiple triggers for the same column ( if multiple modification is happening), and these triggers may also overlap with the job/cursor that is copying old format data to new column.

This scenario is posing two risks, 
1.  Triggers processing ( due to multiple updates on same column ) may be processed in different order and older data may be written to new column. 
2. what if some trigger processing/execution fails, We may log this, report this. But do we have any systematic way ( out of box from the framework itself) to detect this and retry ?

I can think of one solution that may work but please let me know if it fits in the current approach:

Step1: Apply the schema changes, means, create a new column.
Step2: Using dynamic configuration (needs to be implemented), let running instances know that ZDU process has started and in such cases they will write to old column in older format and additionally they will also write an event in one table (say ZDU upgrade table, a new table). After writing at both places , then only, transaction will be successful. 

Step3: As part of your framework, that contains logic to migrate data from col1 to col2, you code/cursor should read event in the insertion order to process this and once it is processed, then only it should be deleted from table. If any Runtime error occurs, since it has not been deleted, it will be retried.

This approach ensures, it would be processed  in the order in which they occurred. And , out of bix from the framework, there would be way to know if all migration done or not.

We should also consider adding one step in "Finalisation Step", to check that this new event table should be empty.

This is just an idea, feel free to add your input if above scenario can be handled in different or better way.

fateh288:
@vikaskr22 
`New binaries not yet applied, means, read/write will continue to happen with old column. How you are handling the case, where a column is being updated by the application after copying from col1-> col2 ? `
--- great point. No, this case is not being handled - any update to old column after it has been copied but before trigger is created won't be updated in the new column. For KMS, it was observed that copy one column to another for 1M rows took 35 seconds. Realistically we won't have 1M rows and would just be a few hundred max rows so time between copy of column and trigger creation would be a few seconds max. 

`If we are using a trigger on each data modification that will copy to new column. Then, we have multiple triggers for the same column ( if multiple modification is happening), and these triggers may also overlap with the job/cursor that is copying old format data to new column.` 
--- Trigger is being created after the copy job is completed, so I am not clear how data copy job and trigger will overlap  


```
This scenario is posing two risks,

Triggers processing ( due to multiple updates on same column ) may be processed in different order and older data may be written to new column.
```
--- from my understanding, is there are multiple modifications to same column, each of the triggers are processed sequentially - this behavior is out of the box for databases -- we can confirm this incase there is any gap in my understanding
`what if some trigger processing/execution fails, We may log this, report this. But do we have any systematic way ( out of box from the framework itself) to detect this and retry ?`
--- My understanding is that if the trigger fails, then the original transaction also fails i.e. original update will be reverted if the trigger fails to copy the data to the new column basically trigger is now a part of original transaction

```
Step1: Apply the schema changes, means, create a new column.
Step2: Using dynamic configuration (needs to be implemented), let running instances know that ZDU process has started and in such cases they will write to old column in older format and additionally they will also write an event in one table (say ZDU upgrade table, a new table). After writing at both places , then only, transaction will be successful.
Step3: As part of your framework, that contains logic to migrate data from col1 to col2, you code/cursor should read event in the insertion order to process this and once it is processed, then only it should be deleted from table. If any Runtime error occurs, since it has not been deleted, it will be retried.
```
--- I am not clear on this. The old bits / old kms instance will not have logic to write to a new table. Old application is not aware of new schema changes, new table that needs to be written to. Writing to both new and old columns can be done by new KMS instance so that new data is available to the old instances too. But how old instances can write to new column I don't fully understand.

liquibase-database-upgrade/scripts/README.md
```diff
@@ -0,0 +1,66 @@
...
+# KMS
+## Sample exports and commands
+```
+export JAVA_HOME=/usr/java/jdk1.8.0
+export RANGER_KMS_HOME=/var/lib/ranger-kms
+#get process folder for kms: ps -ef | grep ranger-kms
+export RANGER_KMS_CONF=/var/run/process/1546336381-ranger_kms-RANGER_KMS_SERVER/conf
+export SQL_CONNECTOR_JAR=/usr/share/java/postgresql-connector-java.jar
+export LIQUI_SEARCH_PATH=/root/ranger-3.0.0-SNAPSHOT-liquibase-upgrade/lib
+#get HADOOP_CREDSTORE_PASSWORD: cat /var/run/process/1546336381-ranger_kms-RANGER_KMS_SERVER/proc.json | grep HADOOP
```
Sanket-Shelar, line 26: @fateh288 could you please let me know how we can retrieve the **HADOOP_CREDSTORE_PASSWORD** in this case? Will it be the same as DB password?

PR #570, opened 2025-05-05
RANGER-5208: refactor agents-audit module to move audit destination implementations to separate projects

mneethiraj:
- introduced following sub-modules under existing agents-audit module:
  - core, dest-cloudwatch, dest-es, dest-hdfs, dest-kafka, dest-log4j, dest-solr, orc-util
- plugin projects dependencies and packaging updated to include only Solr and HDFS audit destination implementations
  - deployments that require other audit destinations should update plugin dependencies and packaging

## How was this patch tested?
- verified build completes successfully
- verified successful auditing to Solr and HDFS from following plugins: HDFS, Hive, HBase, Kafka plugins


embeddedwebserver/pom.xml
```diff
@@ -81,6 +81,16 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.apache.ranger</groupId>
+            <artifactId>ranger-audit-dest-es</artifactId>
```
kumaab, line 86: I think this may be a typo, is `ranger-audit-dest-hdfs` expected here?

mneethiraj: No, this is intentional. Ranger server initializes Solr and Elasticsearch when necessary (by creating the collection that stores audit logs); hence the references to Solr and Elasticsearch audit destinations.

PR #571, opened 2025-05-08
RANGER-5210: private variable isPolicyEngineShared is set to wrong value

vyommani:

In the RangerBasePlugin class, the isPolicyEngineShared variable is incorrectly set to true even when the policy engine instance isn't shared.
 
When delta sync is enabled, the PolicyRefresher downloads modified policies and updates the policy engine. If locking is enabled, it updates the existing policy engine instance; otherwise, it creates a new instance.
 
The setPolicies method creates a new policy engine instance by calling RangerPolicyEngineImpl.getPolicyEngine. However, this method might return the same instance, not a new one.
 
The isPolicyEngineShared flag is set to true if newPolicyEngine isn't null, which is incorrect. Instead, it should be set to true only if newPolicyEngine is the same instance as the old policy engine.

## How was this patch tested?

I tested locally as follows
mvn clean install


mneethiraj (approved):

kumaab:
@vyommani please rebase PR to re-run the CI checks, thanks!

PR #575, opened 2025-05-13
RANGER-5211 : KMS Metric collection thread safety check should not pa…

vikaskr22:
…rse config everytime

KMS metrics collector checks before collection whether the collection is thread-safe or not. Default is false. This config was introduced to avoid unnecessary locking for the non-functional metrics. But for businesses where metrics need to be accurate, it can be marked true. To check this flag, existing code was reading the config value from kmsConf every time. And this required to create one Configuration object and this acquires lock on the underlying config. Hence it was blocking many other threads who actually needs to parse and read something.

I have noticed around 30-40 threads always waiting for the kmsConf lock and it was acquired by the thread that was simply checking whether metric collection is thread safe or not.

Since this config doesn't change during execution, It can be simply read and kept during initialisation once , and can be read from there whenever required.
 
## How was this patch tested?

Build got passed, Unit tests got passed.
Further I created ranger-kms docker image and verified the metric collection by hitting metric API.

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSWebApp.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSWebApp.java
@@ -78,6 +78,7 @@ public class KMSWebApp implements ServletContextListener {
     private static KeyProviderCryptoExtension keyProviderCryptoExtension;
     private static KMSMetricsCollector        kmsMetricsCollector;
 
+    private static boolean isMetricsCollectionThreadsafe;
     private JmxReporter jmxReporter;
 
     public static Configuration getConfiguration() {
@@ -133,7 +134,7 @@ public class KMSWebApp implements ServletContextListener {
     }
 
     public static boolean isMetricCollectionThreadSafe() {
-        return Boolean.parseBoolean(KMSWebApp.getConfiguration().get(HADOOP_KMS_METRIC_COLLECTION_THREADSAFE, "false"));
+        return isMetricsCollectionThreadsafe;
     }
 
     public static KMSMetricsCollector getKmsMetricsCollector() {
@@ -181,6 +182,7 @@ public class KMSWebApp implements ServletContextListener {
 
             kmsAudit = new KMSAudit(kmsConf);
 
+            isMetricsCollectionThreadsafe = Boolean.valueOf(kmsConf.get(HADOOP_KMS_METRIC_COLLECTION_THREADSAFE, "false"));
             KMSMetricWrapper kmsMetricWrapper = KMSMetricWrapper.getInstance(isMetricCollectionThreadSafe());
 
             kmsMetricsCollector = kmsMetricWrapper.getKmsMetricsCollector();


dhavalshah9131 (approved):

spolavarpau1 (approved):

kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSWebApp.java
```diff
@@ -133,7 +134,7 @@ public static KMSAudit getKMSAudit() {
     }
 
     public static boolean isMetricCollectionThreadSafe() {
-        return Boolean.parseBoolean(KMSWebApp.getConfiguration().get(HADOOP_KMS_METRIC_COLLECTION_THREADSAFE, "false"));
+        return isMetricsCollectionThreadsafe;
```
vyommani, line 137: The proposed code change, changes the  behaviour.
Instead of relying on getKmsMetricsCollector to set 'isMetricCollectionThreadsafe', I suggest initializing it statically at class loading time:
private static boolean isMetricCollectionThreadsafe = Boolean.valueOf(KMSWebApp.getConfiguration().get(HADOOP_KMS_METRIC_COLLECTION_THREADSAFE, "false"));
Although this incurs a minor cost at class loading time, it's a safer approach that avoids potential behaviour change.

vikaskr22: Thanks Vyom for the review.

We are not using getKmsMetricsCollector method to set this flag.

Earlier it was parsing and reading from kmsConf every time inside `isMetricCollectionThreadSafe()` but now we set `isMetricCollectionThreadsafe = Boolean.valueOf(KMSWebApp.getConfiguration().get(HADOOP_KMS_METRIC_COLLECTION_THREADSAFE, "false"));` inside `contextInitialized(ServletContextEvent sce) ` method, it's one of the life cyle method of **_ServletContextListener_** . Place to put initialisation stuff.

And also, we need kmsConf that is being read and initialised in this method only.

I am not sure if I got your question right, but can you pls help me understand how this  code change is changing the behaviour ?

kumaab:
@vikaskr22 please rebase PR to re-run the CI checks, thanks!

kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSWebApp.java
vyommani: It appears you're correct that the isMetricCollectionThreadSafe variable isn't initialized in getKMSMetricsCollector. Since isMetricCollectionThreadSafe() is a public static method, its behavior now depends on whether the context is initialized, which introduces a behavioral change.

At line 186, the code calls isMetricCollectionThreadSafe(), and at line 185, you're setting the value of the isMetricsCollectionThreadsafe variable, which is used within isMetricCollectionThreadSafe().

Given that isMetricCollectionThreadSafe() is a public static method, I recommend initializing the isMetricsCollectionThreadsafe variable during class loading to ensure consistent behavior in the isMetricCollectionThreadSafe() method.

vikaskr22: I am getting the thought process, that public static things should not be initialised within instance methods. But this method is not a normal instance method:
```
@Override
    public void contextInitialized(ServletContextEvent sce) {
 }
```
It's a servlet context listener and it gets invoked when the application first gets initialised. Please see below the entry of this class inside web.xml
```
<listener>
    <listener-class>org.apache.hadoop.crypto.key.kms.server.KMSWebApp</listener-class>
  </listener>
```

In earlier implementation also, when it was invoking kmsConf everytime, kmsConf was being initialised inside this method only.
I have only addressed the issue of reading the same value from config file , now it reads once and keep it with them.

And on following: 

> At line 186, the code calls isMetricCollectionThreadSafe(), and at line 185, you're setting the value of the isMetricsCollectionThreadsafe variable, which is used within isMetricCollectionThreadSafe().

Yes, I could have used the var directly. But this method is not being called from this place only. It is being called from few other classes as well. So just to keep it consistent, I called the method instead of using var directly.

Pls let me know if I am missing anything ?

PR #580, opened 2025-05-27
RANGER-5212: Search filter in Roles tab does not consider all relevant role columns

RakeshGuptaDev:

When using the Role Name filter, it only matches values from the Role Name column. It does not match values from the Roles column, even though that column contains the role

## How was this patch tested?

Tested and Validated search for below API.

/roles/lookup/roles?subRoleName=r2
/roles/lookup/roles?subRoleNamePartial=r2

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/util/SearchFilter.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/util/SearchFilter.java
@@ -70,6 +70,8 @@ public class SearchFilter {
     public static final String ROLE_NAME_PARTIAL        = "roleNamePartial";      // search
     public static final String GROUP_NAME_PARTIAL       = "groupNamePartial";     // search
     public static final String USER_NAME_PARTIAL        = "userNamePartial";      // search
+    public static final String SUB_ROLE_NAME            = "subRoleName";          // search
+    public static final String SUB_ROLE_NAME_PARTIAL    = "subRoleNamePartial";   // search
     public static final String SERVICE_NAME_PREFIX      = "serviceNamePrefix";    // search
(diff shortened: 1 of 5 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/util/SearchFilter.java
```diff
@@ -74,6 +74,8 @@ public class SearchFilter {
     public static final String ZONE_NAME_PREFIX         = "zoneNamePrefix";       // search
     public static final String POLICY_NAME_PREFIX       = "policyNamePrefix";
 
+    public static final String ASSOCIATED_ROLE_NAME_PARTIAL = "associatedRoleNamePartial";   // search
```
mneethiraj: @RakeshGuptaDev  - instead of introdcing new seach parameter `associatedRoleNamePartial`, I suggest updating `RoleDBStore.getRolesForUser()` to call `roles = getRoles(filter);` for admin users i.e. as shown below. This will result in consistent handling of role name filter.

```
public RangerRoleList getRolesForUser(SearchFilter filter, RangerRoleList rangerRoleList) {
  List<RangerRole> roles       = new ArrayList<>();
  ...
  if (userSession != null && userSession.getUserRoleList().size() == 1 && userSession.getUserRoleList().contains(RangerConstants.ROLE_USER) && userSession.getLoginId() != null) {
    ...
  } else {
    roles = getRoles(filter)
  }
  ...
}
```

Please make sure to handle pagination parameters appropriately.

RakeshGuptaDev: Handled.

PR #582, opened 2025-05-30
RANGER-5217: revert mandatory distinct clause as getPolicy request may fail for oracle

pradeepagrawal8184:

getPolicy request may fail as the distinct query clause is not valid for oracle. This PR reverts the changes done by RANGER-4410 commit https://github.com/apache/ranger/commit/7528f5d1385bbbcba44bb22157a7e12df38f0570

## How was this patch tested?

After the build tried the get policy call which was failing earlier.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/util/SearchFilter.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/util/SearchFilter.java
@@ -155,7 +155,7 @@ public class SearchFilter {
     private boolean               getCount   = true;
     private String                sortBy;
     private String                sortType;
-    private boolean               isDistinct = true;
+    private boolean               isDistinct;
 
     public SearchFilter() {
         this((Map<String, String>) null);


mneethiraj:
@pradeepagrawal8184  - wouldn't this revert cause regression in GDS search API (enhancements in RANGER-4410)?

pradeepagrawal8184:
> @pradeepagrawal8184 - wouldn't this revert cause regression in GDS search API (enhancements in [RANGER-4410](https://issues.apache.org/jira/browse/RANGER-4410))?

1) This PR is reverting the change done via RANGER-4410 which regressed the download the getPolicy call in oracle env.
2) From RANGER-4410 changes, i am unable to figure out why distinct was needed accordingly i can optimize the sql call. 
i tried the call with and without distinct clause in mysql and results were same so i could not figure out much.

kishorgollapalliwar:
@pradeepagrawal8184 and @mneethiraj disabling distinct is expected to break following APIs

- /service/gds/dataset/summary?dataShareNamePartial=Test_datashare
- /service/gds/datashare/summary?dataShareNamePartial=Test_datashare

PR #584, opened 2025-06-05
RANGER-5215 : Policy authroisation fails for Ranger Plugins in case of users/groups converted by Ranger userysnc as per given Regex

dhavalshah9131:
**Problem Statement:**

Currently, when Ranger Usersync is configured with case conversion and special character replacement using regex, it transforms the original user/group names from the source (e.g., AD/LDAP) before storing them in the Ranger Admin database.

**Example:**

Original name in LDAP/AD: John-jacobs
Usersync configuration:

- ranger.usersync.ldap.username.caseconversion = lower
- ranger.usersync.mapping.username.regex = s/[-]/_/g
- Transformed and stored name in Ranger: john_jacobs

**Issue:**

If a Ranger plugin (e.g., Hive) uses the original name John-jacobs during authorization checks, it fails because Ranger Admin only recognizes the transformed name john_jacobs.

**Error Example:**

_Permission denied: user [John-jacobs] does not have [SELECT] privilege on [vehicle/cars/*]_ 
**Solution:**

To ensure consistency, the same transformation logic used by Usersync must also be applied on the plugin side before authorization. This transformation should be made available as a utility library packaged with the plugins.

**Configurability:**

This feature must be configurable at the plugin level via a property (e.g., ranger.plugin.<serviceType>.supports.name.transformation), allowing users to enable or disable it based on their environment needs.

In ranger-admin-site.xml

ranger.plugins.ldap.username.caseconversion
ranger.plugins.ldap.groupname.caseconversion
ranger.plugins.mapping.username.handler
ranger.plugins.mapping.groupname.handler
ranger.plugins.mapping.regex.separator
ranger.plugins.mapping.username.regex
ranger.plugins.mapping.groupname.regex

## How was this patch tested?

1.) Build successful with unit test.
2.) Manul testing

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/model/UgsyncNameTransformRules.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/model/UgsyncNameTransformRules.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *=-0987654321`o6 bftware distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.ranger.plugin.model;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.apache.ranger.plugin.util.RangerUserStoreUtil;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+
+import java.util.Map;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.FIELD)
+public class UgsyncNameTransformRules extends RangerBaseModelObject implements java.io.Serializable {
+
+    private static final long   serialVersionUID = 1L;
+    Map<String, String> nameTransformRules;
+    private final String name;
+
+    public UgsyncNameTransformRules() {this(null, null, null);}
+
+    public UgsyncNameTransformRules(String guid, String name, Map<String, String> nameTransformRules) {
+        super();
+        setGuid(guid);
+        setId(Long.valueOf(1));
+        setVersion(Long.valueOf(1));
+        this.name               = name;
+        this.nameTransformRules = nameTransformRules;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    public Map<String, String> getNameTransformRules() {
+        return nameTransformRules;
+    }
+
+    @Override
+    public String toString() {
+        return "{nameTransformRules=" + RangerUserStoreUtil.getPrintableOptions(nameTransformRules)
+                + "}";
+    }
+}
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPluginContext.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPluginContext.java
@@ -27,6 +27,7 @@ import org.apache.ranger.plugin.model.RangerPolicy;
 import org.apache.ranger.plugin.resourcematcher.RangerResourceMatcher;
 import org.apache.ranger.plugin.service.RangerAuthContext;
 import org.apache.ranger.plugin.service.RangerAuthContextListener;
+import org.apache.ranger.ugsyncutil.transform.Mapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -39,13 +40,49 @@ public class RangerPluginContext {
 
     private final RangerPluginConfig                                                         config;
     private final Map<String, Map<RangerPolicy.RangerPolicyResource, RangerResourceMatcher>> resourceMatchers = new HashMap<>();
-    private final ReentrantReadWriteLock                                                     lock             = new ReentrantReadWriteLock(true); // fair lock
+    private final ReentrantReadWriteLock                                                     lock             = new ReentrantReadWriteLock(true);// fair lock
+    private       Mapper                                                                     userNameTransformInst;
+    private       Mapper                                                                     groupNameTransformInst;
+    private       String                                                                     userNameCaseConversion;
+    private       String                                                                     groupNameCaseConversion;
     private       RangerAuthContext                                                          authContext;
     private       RangerAuthContextListener                                                  authContextListener;
     private       RangerAdminClient                                                          adminClient;
 
     public RangerPluginContext(RangerPluginConfig config) {
         this.config = config;
+    }
+
+    public Mapper getUserNameTransformInst() {
+        return userNameTransformInst;
+    }
+
+    public void setUserNameTransformInst(Mapper userNameTransformInst) {
+        this.userNameTransformInst = userNameTransformInst;
+    }
+
+    public Mapper getGroupNameTransformInst() {
+        return groupNameTransformInst;
+    }
+
+    public void setGroupNameTransformInst(Mapper groupNameTransformInst) {
+        this.groupNameTransformInst = groupNameTransformInst;
+    }
+
+    public String getUserNameCaseConversion() {
+        return userNameCaseConversion;
+    }
+
+    public void setUserNameCaseConversion(String userNameCaseConversion) {
+        this.userNameCaseConversion = userNameCaseConversion;
+    }
+
+    public String getGroupNameCaseConversion() {
+        return groupNameCaseConversion;
+    }
+
+    public void setGroupNameCaseConversion(String groupNameCaseConversion) {
+        this.groupNameCaseConversion = groupNameCaseConversion;
     }
 
     public RangerPluginConfig getConfig() {
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
@@ -40,6 +40,7 @@ import org.apache.ranger.plugin.util.RangerPerfTracer;
 import org.apache.ranger.plugin.util.RangerReadWriteLock;
 import org.apache.ranger.plugin.util.RangerRoles;
 import org.apache.ranger.plugin.util.ServicePolicies;
+import org.apache.ranger.ugsyncutil.transform.Mapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -83,6 +84,53 @@ public class RangerPolicyEngineImpl implements RangerPolicyEngine {
         policyEngine     = new PolicyEngine(servicePolicies, pluginContext, roles, isUseReadWriteLock);
         serviceConfig    = new ServiceConfig(servicePolicies.getServiceConfig());
         requestProcessor = new RangerDefaultRequestProcessor(policyEngine);
+
+        Map<String, String> serviceConfigMap = servicePolicies.getServiceConfig();
+        if (MapUtils.isNotEmpty(serviceConfigMap)) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("==> RangerBasePlugin(" + serviceConfigMap.keySet() + ")");
+            }
+            pluginContext.setUserNameCaseConversion(serviceConfigMap.get(RangerCommonConstants.PLUGINS_USERNAME_CASE_CONVERSION_PARAM));
+            pluginContext.setGroupNameCaseConversion(serviceConfigMap.get(RangerCommonConstants.PLUGINS_GROUPNAME_CASE_CONVERSION_PARAM));
+            String mappingUserNameHandler = serviceConfigMap.get(RangerCommonConstants.PLUGINS_MAPPING_USERNAME_HANDLER);
+            try {
+                if (mappingUserNameHandler != null) {
+                    Class<Mapper> regExClass        = (Class<Mapper>) Class.forName(mappingUserNameHandler);
+                    Mapper        userNameRegExInst = regExClass.newInstance();
+                    if (userNameRegExInst != null) {
+                        String baseProperty = RangerCommonConstants.PLUGINS_MAPPING_USERNAME;
+                        userNameRegExInst.init(baseProperty, getAllRegexPatterns(baseProperty, serviceConfigMap),
+                                serviceConfigMap.get(RangerCommonConstants.PLUGINS_MAPPING_SEPARATOR));
+                        pluginContext.setUserNameTransformInst(userNameRegExInst);
+                    } else {
+                        LOG.error("RegEx handler instance for username is null!");
+                    }
+                }
+            } catch (ClassNotFoundException cne) {
+                LOG.error("Failed to load " + mappingUserNameHandler + " " + cne);
+            } catch (Throwable te) {
+                LOG.error("Failed to instantiate " + mappingUserNameHandler + " " + te);
+            }
+            String mappingGroupNameHandler = serviceConfigMap.get(RangerCommonConstants.PLUGINS_MAPPING_GROUPNAME_HANDLER);
+            try {
+                if (mappingGroupNameHandler != null) {
+                    Class<Mapper> regExClass         = (Class<Mapper>) Class.forName(mappingGroupNameHandler);
+                    Mapper        groupNameRegExInst = regExClass.newInstance();
+                    if (groupNameRegExInst != null) {
+                        String baseProperty = RangerCommonConstants.PLUGINS_MAPPING_GROUPNAME;
+                        groupNameRegExInst.init(baseProperty, getAllRegexPatterns(baseProperty, serviceConfigMap),
+                                serviceConfigMap.get(RangerCommonConstants.PLUGINS_MAPPING_SEPARATOR));
+                        pluginContext.setGroupNameTransformInst(groupNameRegExInst);
+                    } else {
+                        LOG.error("RegEx handler instance for groupname is null!");
+                    }
+                }
+            } catch (ClassNotFoundException cne) {
+                LOG.error("Failed to load " + mappingGroupNameHandler + " " + cne);
+            } catch (Throwable te) {
+                LOG.error("Failed to instantiate " + mappingGroupNameHandler + " " + te);
+            }
+        }
     }
 
     private RangerPolicyEngineImpl(final PolicyEngine policyEngine, RangerPolicyEngineImpl other) {
@@ -552,6 +600,31 @@ public class RangerPolicyEngineImpl implements RangerPolicyEngine {
 
     public RangerAccessRequestProcessor getRequestProcessor() {
         return requestProcessor;
+    }
+
+    private List<String> getAllRegexPatterns(String baseProperty, Map<String, String> serviceConfig) throws Throwable {
+        List<String> regexPatterns = new ArrayList<String>();
+        String       baseRegex     = serviceConfig.get(baseProperty);
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("==> getAllRegexPatterns(" + baseProperty + ")");
+            LOG.debug("baseRegex = " + baseRegex);
+            LOG.debug("pluginConfig = " + serviceConfig.keySet());
+        }
+        if (baseRegex == null) {
+            return regexPatterns;
+        }
+        regexPatterns.add(baseRegex);
+        int    i         = 1;
+        String nextRegex = serviceConfig.get(baseProperty + "." + i);
+        while (nextRegex != null) {
+            regexPatterns.add(nextRegex);
+            i++;
+            nextRegex = serviceConfig.get(baseProperty + "." + i);
+        }
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("<== getAllRegexPatterns(" + regexPatterns + ")");
+        }
+        return regexPatterns;
     }
 
     private RangerAccessResult zoneAwareAccessEvaluationWithNoAudit(RangerAccessRequest request, int policyType) {
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
@@ -31,24 +31,29 @@ import org.apache.ranger.plugin.policyengine.RangerAccessResource;
 import org.apache.ranger.plugin.policyengine.RangerMutableResource;
 import org.apache.ranger.plugin.policyengine.RangerPluginContext;
 import org.apache.ranger.plugin.util.RangerAccessRequestUtil;
+import org.apache.ranger.plugin.util.RangerCommonConstants;
 import org.apache.ranger.plugin.util.RangerPerfTracer;
 import org.apache.ranger.plugin.util.RangerUserStoreUtil;
+import org.apache.ranger.ugsyncutil.transform.Mapper;
+import org.apache.ranger.ugsyncutil.util.UgsyncCommonConstants;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public class RangerDefaultRequestProcessor implements RangerAccessRequestProcessor {
     private static final Logger LOG                              = LoggerFactory.getLogger(RangerDefaultRequestProcessor.class);
     private static final Logger PERF_CONTEXTENRICHER_REQUEST_LOG = RangerPerfTracer.getPerfLogger("contextenricher.request");
 
     protected final PolicyEngine policyEngine;
-    private   final boolean      useRangerGroups;
-    private   final boolean      useOnlyRangerGroups;
-    private   final boolean      convertEmailToUser;
+    private final   boolean      useRangerGroups;
+    private final   boolean      useOnlyRangerGroups;
+    private final   boolean      convertEmailToUser;
 
     public RangerDefaultRequestProcessor(PolicyEngine policyEngine) {
         this.policyEngine = policyEngine;
@@ -96,6 +101,63 @@ public class RangerDefaultRequestProcessor implements RangerAccessRequestProcess
 
                 if (reqImpl.getClusterType() == null) {
                     reqImpl.setClusterType(pluginContext.getClusterType());
+                }
+
+                RangerPluginConfig config = policyEngine.getPluginContext().getConfig();
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug("RangerPluginConfig = " + config.getPropertyPrefix());
+                }
+                if (config != null) {
+                    boolean isNameTransformationSupported = config.getBoolean(config.getPropertyPrefix() + RangerCommonConstants.PLUGIN_CONFIG_SUFFIX_NAME_TRANSFORMATION, false);
+
+                    if (LOG.isDebugEnabled()) {
+                        LOG.debug("isNameTransformationSupported = " + isNameTransformationSupported);
+                    }
+                    if (isNameTransformationSupported) {
+                        String user = request.getUser();
+                        if (policyEngine.getPluginContext().getUserNameCaseConversion().equalsIgnoreCase(UgsyncCommonConstants.UGSYNC_LOWER_CASE_CONVERSION_VALUE)) {
+                            user = user.toLowerCase();
+                        } else if (policyEngine.getPluginContext().getUserNameCaseConversion().equalsIgnoreCase(UgsyncCommonConstants.UGSYNC_UPPER_CASE_CONVERSION_VALUE)) {
+                            user = user.toUpperCase();
+                        }
+                        Mapper userNameTransformInst = policyEngine.getPluginContext().getUserNameTransformInst();
+                        if (userNameTransformInst != null) {
+                            String originalUser    = request.getUser();
+                            String transformedUser = userNameTransformInst.transform(user);
+
+                            if (LOG.isDebugEnabled()) {
+                                LOG.debug("Original username = {}, Transformed username = {}", originalUser, transformedUser);
+                            }
+
+                            reqImpl.setUser(transformedUser);
+                        }
+                        Mapper groupNameTransformInst = policyEngine.getPluginContext().getGroupNameTransformInst();
+                        if (groupNameTransformInst != null) {
+                            Set<String> transformedGroups = request.getUserGroups().stream()
+                                    .filter(Objects::nonNull)
+                                    .map(group -> {
+                                        String originalGroupName = group;
+                                        String conversionType    = policyEngine.getPluginContext().getGroupNameCaseConversion();
+
+                                        if (UgsyncCommonConstants.UGSYNC_LOWER_CASE_CONVERSION_VALUE.equalsIgnoreCase(conversionType)) {
+                                            group = group.toLowerCase();
+                                        } else if (UgsyncCommonConstants.UGSYNC_UPPER_CASE_CONVERSION_VALUE.equalsIgnoreCase(conversionType)) {
+                                            group = group.toUpperCase();
+                                        }
+
+                                        String transformedGroup = groupNameTransformInst.transform(group);
+
+                                        if (LOG.isDebugEnabled()) {
+                                            LOG.debug("Original group name = {}, Transformed group name = {}", originalGroupName, transformedGroup);
+                                        }
+
+                                        return transformedGroup;
+                                    })
+                                    .collect(Collectors.toSet());
+
+                            reqImpl.setUserGroups(transformedGroups);
+                        }
+                    }
                 }
 
                 convertEmailToUsername(reqImpl);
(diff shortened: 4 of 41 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
```diff
@@ -98,6 +103,63 @@ public void preProcess(RangerAccessRequest request) {
                     reqImpl.setClusterType(pluginContext.getClusterType());
                 }
 
+                RangerPluginConfig config = policyEngine.getPluginContext().getConfig();
+                if (LOG.isDebugEnabled()) {
```
kumaab: Please remove all references of LOG.isDebugEnabled() unless absolutely required. References of the same were extensively removed as part of `RANGER-5017` and is also documented here: https://cwiki.apache.org/confluence/display/RANGER/Apache+Ranger+Java+Style+Guide

ugsync-util/src/test/java/org/apache/ranger/ugsynutil/transform/TestRegEx.java
```diff
@@ -17,17 +17,17 @@
...
+import org.apache.ranger.ugsyncutil.transform.RegEx;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-import static org.junit.Assert.assertEquals;
```
kumaab: Why is static `assertEquals` being removed?

agents-common/src/main/java/org/apache/ranger/plugin/model/UgsyncNameTransformRules.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.Map;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.FIELD)
+public class UgsyncNameTransformRules extends RangerBaseModelObject implements java.io.Serializable {
+
```
kumaab: The latest PR build shows 6 checkstyle violations, please review.

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
```diff
@@ -83,6 +84,53 @@ public RangerPolicyEngineImpl(ServicePolicies servicePolicies, RangerPluginConte
         policyEngine     = new PolicyEngine(servicePolicies, pluginContext, roles, isUseReadWriteLock);
         serviceConfig    = new ServiceConfig(servicePolicies.getServiceConfig());
         requestProcessor = new RangerDefaultRequestProcessor(policyEngine);
+
+        Map<String, String> serviceConfigMap = servicePolicies.getServiceConfig();
```
rameeshm: can the whole pluginContext setting be done upfront in basePlugin before RangerPolicyEngineImpl is created?

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
```diff
@@ -83,6 +84,53 @@ public RangerPolicyEngineImpl(ServicePolicies servicePolicies, RangerPluginConte
...
+                        String baseProperty = RangerCommonConstants.PLUGINS_MAPPING_GROUPNAME;
+                        groupNameRegExInst.init(baseProperty, getAllRegexPatterns(baseProperty, serviceConfigMap),
+                                serviceConfigMap.get(RangerCommonConstants.PLUGINS_MAPPING_SEPARATOR));
+                        pluginContext.setGroupNameTransformInst(groupNameRegExInst);
+                    } else {
+                        LOG.error("RegEx handler instance for groupname is null!");
+                    }
+                }
+            } catch (ClassNotFoundException cne) {
+                LOG.error("Failed to load " + mappingGroupNameHandler + " " + cne);
```
rameeshm: Print the whole stack trace, do this for all the occurance
LOG.error("Failed to load " + mappingGroupNameHandler + " " , cne);

ugsync-util/src/test/java/org/apache/ranger/ugsynutil/transform/TestRegEx.java
dhavalshah9131: Shown as unused import

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -1919,6 +1899,13 @@ public Map<String, String> getServiceConfigForPlugin(Long serviceId) {
                 }
             }
         }
+        Map<String, String> ugsyncConfigs = getLatestUgsyncConfig();
```
fateh288: @dhavalshah9131 instead of reading the configs only for ugsync, can we read the configs starting with a certain prefix from ranger admin site xml-- this will help https://github.com/apache/ranger/pull/595 too

dhavalshah9131: @fateh288 
Updated the PR. Kindly review it.

security-admin/src/main/java/org/apache/ranger/common/PropertiesUtil.java
```diff
@@ -163,6 +164,36 @@ public static Properties getProps() {
...
+                String value = StringUtils.isNotEmpty(entry.getValue()) ? entry.getValue() : null;
+                if (value != null) {
+                    configMap.put(key, value);
+                }
+            }
+        }
+        return configMap;
+    }
+
+    public static Map<String, String> getConfigMapWithPrefixAndDefaultValue(String confPrefix, Set<String> excludedKeys) {
```
fateh288: There is very high code duplication in this method with the above method getConfigMapWithPrefixAndDefaultValue(String confPrefix). 
I think it would be better to have a separate utility called removeConfigsIfPresent() which remove the excluded keys from the hashmap.
So first we can call getConfigMapWithPrefixAndDefaultValue(String confPrefix) and then call removeConfigsIfPresent(HashMap<String, String> configs, Set<String> toRemoveConfigs) to achieve the same desired result

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -1919,6 +1899,22 @@ public Map<String, String> getServiceConfigForPlugin(Long serviceId) {
                 }
             }
         }
+        Set<String> excludedKeys = new HashSet<>(
```
fateh288: I think we should read these properties we want to exclude from the configs as a separate config instead of hardcoding these properties in the code to improve extensibility.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -2860,6 +2856,45 @@ public void updateServiceAuditConfig(String searchUsrGrpRoleName, REMOVE_REF_TYP
         LOG.debug("<=== ServiceDBStore.updateServiceAuditConfig( searchUsrGrpRoleName : {} removeRefType : {})", searchUsrGrpRoleName, removeRefType);
     }
 
+    public Map<String, String> getPluginsSpecialConfigsForNameTransformed() {
```
fateh288: We should move this and any property / configs related operations to PropertiesUtil class instead of having it in ServiceDBStore

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -1919,6 +1899,22 @@ public Map<String, String> getServiceConfigForPlugin(Long serviceId) {
...
             }
         }
+        Set<String> excludedKeys = new HashSet<>(
+                Arrays.asList(
+                RangerCommonConstants.PLUGINS_MAPPING_SEPARATOR,
+                RangerCommonConstants.PLUGINS_MAPPING_USERNAME,
+                RangerCommonConstants.PLUGINS_MAPPING_GROUPNAME
+        )
+        );
+        Map<String, String> rangerPluginsPrefixConfig = PropertiesUtil.getConfigMapWithPrefixAndDefaultValue("ranger.plugins.", excludedKeys);
```
fateh288: Why are we naming this method as getConfigMapWithPrefixAndDefaultValue and not  getConfigMapWithPrefix() ?
I don't see any default value being passed to this function / being read within the method

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
fateh288: Also, these all seem to be static methods

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -2860,6 +2856,45 @@ public void updateServiceAuditConfig(String searchUsrGrpRoleName, REMOVE_REF_TYP
...
 
+    public Map<String, String> getPluginsSpecialConfigsForNameTransformed() {
+        Map<String, String> configs = new HashMap<String, String>();
+        configs.put(RangerCommonConstants.PLUGINS_MAPPING_SEPARATOR, getRegexSeparator());
+        configs.putAll(getAllRegexPatternsConfig(RangerCommonConstants.PLUGINS_MAPPING_USERNAME));
+        configs.putAll(getAllRegexPatternsConfig(RangerCommonConstants.PLUGINS_MAPPING_GROUPNAME));
+        return configs;
+    }
+
+    public String getRegexSeparator() {
```
fateh288: This method can be private

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -2860,6 +2856,45 @@ public void updateServiceAuditConfig(String searchUsrGrpRoleName, REMOVE_REF_TYP
...
+                ret = val;
+            } else {
+                LOG.warn("More than one character found in RegEx Separator, using default RegEx Separator");
+            }
+        }
+        LOG.info(String.format("Using %s as the RegEx Separator", ret));
+        return ret;
+    }
+
+    public Map<String, String> getAllRegexPatternsConfig(String baseProperty) {
```
fateh288: This method can be private

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
fateh288: Another suggestion I have is - instead of removing these configs in the code, can we consider naming these configs such that they don't start with "ranger.plugins." ? This would keep the logic uniform - that any property starting with "ranger.plugins." would go into all all plugins. Adding exceptions to the rules adds code complexity and confusion

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
fateh288: We should have a constant string defined for "ranger.plugins." similar to RANGER_PLUGIN_CONFIG_PREFIX.

fateh288 (changes requested):

rameeshm (commented):
@dhavalshah9131  Please correct the JIRA description as it has Typo. 
RANGER-5215 : Policy authorization fails for Ranger Plugins in case of users/groups converted by Ranger userysnc as per given Regex

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
```diff
@@ -98,6 +103,53 @@ public void preProcess(RangerAccessRequest request) {
                     reqImpl.setClusterType(pluginContext.getClusterType());
                 }
 
+                RangerPluginConfig config = policyEngine.getPluginContext().getConfig();
```
fateh288, line 106: Can we break this code segment into multiple utility functions.
Two functions I can think of for better readability are 1) getTransformedUser()
2) getTransformedGroups()
After calling these methods, we can then set it reqImpl.setUser(transformedUser); / reqImpl.setUserGroups(transformedGroups);

fateh288 (approved):
I have one more refactoring suggestion, but other than that it looks good

ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
```diff
@@ -36,212 +37,174 @@
...
-    public static final String DEFAULT_CONFIG_FILE                                                   = "ranger-ugsync-default.xml";
-    public static final String UGSYNC_ENABLED_PROP                                                   = "ranger.usersync.enabled";
-
-    private static final String UGSYNC_SOURCE_CLASS_PARAM                                            = "ranger.usersync.source.impl.class";
-    private static final String UGSYNC_SINK_CLASS_PARAM                                              = "ranger.usersync.sink.impl.class";
-    private static final String UGSYNC_SOURCE_CLASS                                                  = "org.apache.ranger.unixusersync.process.UnixUserGroupBuilder";
-    private static final String UGSYNC_SINK_CLASS                                                    = "org.apache.ranger.unixusersync.process.PolicyMgrUserGroupBuilder";
-    private static final String LGSYNC_SOURCE_CLASS                                                  = "org.apache.ranger.ldapusersync.process.LdapUserGroupBuilder";
-
+    public static final String CONFIG_FILE         = "ranger-ugsync-site.xml";
```
kumaab: Can we keep the formatting changes, these were done as part of [RANGER-5064](https://issues.apache.org/jira/browse/RANGER-5064), why are they being reverted for this change?

dhavalshah9131: By mistake it gt reverted. I have updated the PR.
Thanks

kumaab: Oh I see, thanks.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -222,147 +222,113 @@
...
+    public static final     Integer                       ITERATION_COUNT                   = PropertiesUtil.getIntProperty("ranger.password.iteration.count", PasswordUtils.DEFAULT_ITERATION_COUNT);
+    public static final     String                        RANGER_PLUGIN_AUDIT_FILTERS       = "ranger.plugin.audit.filters";
+    public static final     String                        RANGER_PLUGINS_CONFIG_CONF_PREFIX = "ranger.plugins.conf.";
+    public static final     String                        HIDDEN_PASSWORD_STR               = "*****";
+    public static final     String                        CONFIG_KEY_PASSWORD               = "password";
+    public static final     String                        ACCESS_TYPE_DECRYPT_EEK           = "decrypteek";
+    public static final     String                        ACCESS_TYPE_GENERATE_EEK          = "generateeek";
+    public static final     String                        ACCESS_TYPE_GET_METADATA          = "getmetadata";
+    private static final    Logger                        LOG                               = LoggerFactory.getLogger(ServiceDBStore.class);
```
kumaab, line 239: Extensive refactoring changes seem unnecessary, please see guidelines here (RANGER-5017): https://cwiki.apache.org/confluence/display/RANGER/Apache+Ranger+Java+Style+Guide

- logger should be at the top
- LOG.isDebugEnabled() should not be used unless it's a heavy operation.
- Avoid String concatenation in log statements.

dhavalshah9131: Following the steps mentioned in [RANGER-5018](https://issues.apache.org/jira/browse/RANGER-5018) Code refactoring is been done.

logger should be at the top
=> LOG seems to be private. As per steps mentioned [RANGER-5018](https://issues.apache.org/jira/browse/RANGER-5018) it gets shfited after public variables.

LOG.isDebugEnabled() should not be used unless it's a heavy operation.
=> Fixed.

Avoid String concatenation in log statements.
=> Fixed.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
```diff
@@ -98,6 +103,18 @@ public void preProcess(RangerAccessRequest request) {
...
+                LOG.debug("RangerPluginConfig = " + config.getPropertyPrefix());
```
kumaab: Avoid string concat, use {} instead, see other usages as well.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1274,6 +1275,76 @@ private RangerAdminClient getAdminClient() throws Exception {
...
+                    LOG.error("Failed to instantiate " + mappingGroupNameHandler + " ", te);
+                }
+            }
+        }
+    }
+
+    private List<String> getAllRegexPatterns(String baseProperty, Map<String, String> serviceConfig) throws Throwable {
+        List<String> regexPatterns = new ArrayList<String>();
+        String       baseRegex     = serviceConfig.get(baseProperty);
+        LOG.debug("==> getAllRegexPatterns(" + baseProperty + ")");
```
kumaab: Avoid string concat, use {} instead, see other usages as well.

ugsync-util/src/main/java/org/apache/ranger/ugsyncutil/transform/RegEx.java
```diff
@@ -17,34 +17,27 @@
...
+    public void init(String baseProperty, List<String> regexPatterns, String regexSeparator) {
+        AbstractMapper.logger.info("Initializing for " + baseProperty);
         try {
-            List<String> regexPatterns  = config.getAllRegexPatterns(baseProperty);
-            String       regexSeparator = config.getRegexSeparator();
-
             populateReplacementPatterns(baseProperty, regexPatterns, regexSeparator);
         } catch (Throwable t) {
-            logger.error("Failed to initialize {}", baseProperty, t.fillInStackTrace());
+            AbstractMapper.logger.error("Failed to initialize " + baseProperty, t.fillInStackTrace());
```
kumaab: Avoid string concat, use {} instead, see other usages as well.

kumaab (approved):

ugsync-util/src/main/java/org/apache/ranger/ugsyncutil/transform/RegEx.java
```diff
@@ -17,34 +17,27 @@
...
     public LinkedHashMap<String, String> getReplacementPattern() {
         return replacementPattern;
     }
 
     @Override
-    public void init(String baseProperty) {
-        logger.info("Initializing for {}", baseProperty);
-
+    public void init(String baseProperty, List<String> regexPatterns, String regexSeparator) {
+        AbstractMapper.logger.info("Initializing for {}", baseProperty);
```
mneethiraj: Why is it necessary to add class name qualifier when referencing static member `logger`? Alternatively, consider adding a static import:

```
import static org.apache.ranger.ugsyncutil.transform.AbstractMapper.logger;
```

ugsync-util/src/test/java/org/apache/ranger/ugsynutil/transform/TestRegEx.java
```diff
@@ -49,21 +49,21 @@ public void setUp() {
     public void testUserNameTransform() {
         userRegexPatterns.add("s/\\s/_/");
         userNameRegEx.populateReplacementPatterns(userNameBaseProperty, userRegexPatterns, mappingSeparator);
-        assertEquals("test_user", userNameRegEx.transform("test user"));
+        Assert.assertEquals("test_user", userNameRegEx.transform("test user"));
```
mneethiraj: Instead of qualifying each reference to a static method with the class name, consider statically importing the methods, like:

```
import static org.junit.Assert.assertEquals;
```

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -2672,6 +2672,9 @@ public ServicePolicies getSecureServicePoliciesIfUpdated(@PathParam("serviceName
             throw restErrorUtil.createRESTException(httpCode, logMsg, logError);
         }
 
+        if (ret != null) {
+            LOG.debug("<== ServiceREST.getSecureServicePoliciesIfUpdated(): configs =" + ret.getServiceConfig());
```
mneethiraj: - to be consistent, remove `<== ` from the log message
- replace string concat with parameterized message; also replace `if (ret != null)` at line 2675 with an inline check:
```
LOG.debug("getSecureServicePoliciesIfUpdated(): configs ={}", ret == null ? ret : ret.getServiceConfig());
```

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -194,7 +191,8 @@ public class ServiceREST {
     public static final String PURGE_RECORD_TYPE_LOGIN_LOGS         = "login_records";
     public static final String PURGE_RECORD_TYPE_TRX_LOGS           = "trx_records";
     public static final String PURGE_RECORD_TYPE_POLICY_EXPORT_LOGS = "policy_export_logs";
-
+    private static final Logger LOG      = LoggerFactory.getLogger(ServiceREST.class);
```
mneethiraj: To be consistent with rest of the sources, have initialization of `Logger` instances at the beginning of the class.

security-admin/src/main/java/org/apache/ranger/common/RangerServicePoliciesCache.java
```diff
@@ -358,12 +358,13 @@ ServicePolicies getLatestOrCached(String serviceName, ServiceStore serviceStore,
 
                             if (servicePoliciesForDeltas != null && servicePoliciesForDeltas.getPolicyDeltas() != null) {
                                 LOG.debug("Deltas were requested. Returning deltas from lastKnownVersion:[{}]", lastKnownVersion);
-
                                 if (isDeltaCacheReinitialized) {
                                     this.deltaCache = new ServicePolicyDeltasCache(lastKnownVersion, servicePoliciesForDeltas);
                                 }
 
+                                LOG.debug("servicePoliciesForDeltas = " + servicePoliciesForDeltas.getServiceConfig());
```
mneethiraj: replace string concat with parameterized log message.
```
LOG.debug("servicePoliciesForDeltas = {}", servicePoliciesForDeltas.getServiceConfig());
```

security-admin/src/main/java/org/apache/ranger/common/RangerServicePoliciesCache.java
```diff
@@ -358,12 +358,13 @@ ServicePolicies getLatestOrCached(String serviceName, ServiceStore serviceStore,
...
                                 ret = servicePoliciesForDeltas;
+                                LOG.debug("ret = " + ret.getServiceConfig());
```
mneethiraj: replace string concat with parameterized log message.
```
LOG.debug("ret = {}", ret.getServiceConfig());
```

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
mneethiraj: @dhavalshah9131  - to be consistent with rest of the code base, I suggest keeping instantiation of Logger instances at start of the class, even if they are private static.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -1818,6 +1785,10 @@ public ServicePolicies getServicePoliciesIfUpdated(String serviceName, Long last
             }
         }
 
+        if (ret != null) {
```
mneethiraj: - replace `if (ret != null)` with inline check as shown below
- remove `<==` in the log message

```
LOG.debug("getServicePoliciesIfUpdated({}, {}, {}): configs = {}", serviceName, lastKnownVersion, needsBackwardCompatibility, ret == null ? null ? ret.getServiceConfig());
```

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -1842,6 +1813,10 @@ public ServicePolicies getServicePolicyDeltas(String serviceName, Long lastKnown
             ret = getServicePolicies(serviceName, lastKnownVersion, true, SUPPORTS_POLICY_DELTAS, cachedPolicyVersion);
         }
 
+        if (ret != null) {
```
mneethiraj: - replace if (ret != null) with inline check as shown below
- replace `<===` with `<==`
```
LOG.debug("<== ServiceDBStore.getServicePolicyDeltas({}, {}): ret = {}", serviceName, lastKnownVersion, ret == null ? ret : ret.getServiceConfig());
```

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -4112,6 +4099,10 @@ private ServicePolicies getServicePolicies(String serviceName, Long lastKnownVer
             ret.setTagPolicies(tagPolicies);
         }
 
+        if (ret != null) {
```
mneethiraj: - replace `if (ret != null)` with inline check as shown below
- remove `<==` in the log message
```
LOG.debug("ServiceDBStore.getServicePolicies({}, {}): ret = {}", serviceName, lastKnownVersion, ret == null ? null : ret.getServiceConfig());
```

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1274,6 +1275,76 @@ private RangerAdminClient getAdminClient() throws Exception {
...
+                }
+            }
+        }
+    }
+
+    private List<String> getAllRegexPatterns(String baseProperty, Map<String, String> serviceConfig) throws Throwable {
+        List<String> regexPatterns = new ArrayList<String>();
+        String       baseRegex     = serviceConfig.get(baseProperty);
+        LOG.debug("==> getAllRegexPatterns({})", baseProperty);
+        LOG.debug("baseRegex = {}", baseRegex);
```
mneethiraj: Consider consoldating debug log messages in 1331 and 1332:
```
LOG.debug("baseRegex = {}, pluginConfig = {}", baseRegex, serviceConfig == null ? null  : serviceConfig.keySet());
``

File: agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java [hunk 29]
```diff
@@ -1274,6 +1275,76 @@ private RangerAdminClient getAdminClient() throws Exception {
...
+        LOG.debug("pluginConfig = {}", serviceConfig != null ? serviceConfig.keySet() : "null");
+        if (baseRegex == null) {
+            return regexPatterns;
```
mneethiraj: I suggest avoiding multiple `return` statements in a method.

```
  if (baseRegEx != null) {
      ...
  }

  LOG.debug("<== getAllRegexPatterns({})", regexPatterns);

  return regexPatterns;
```

agents-common/src/main/java/org/apache/ranger/plugin/model/UgsyncNameTransformRules.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.Map;
+
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@XmlRootElement
+@XmlAccessorType(XmlAccessType.FIELD)
+public class UgsyncNameTransformRules extends RangerBaseModelObject implements java.io.Serializable {
+    private static final long   serialVersionUID = 1L;
+    private final        String name;
```
mneethiraj: - are instances of `UgsyncNameTransformRules` serialized/deserialied in REST API calls i.e. parameters or return values? If not, I suggest removing @Json annotations.
- why is the member `name` final? If this is necessary, why not `nameTransformRules` as well?
- Is the default constructor at line 43 necessary, if the members are final?

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
dhavalshah9131: Hi @mneethiraj ,

LOG seems to be private. As per steps mentioned [RANGER-5018](https://issues.apache.org/jira/browse/RANGER-5018) it gets shfited after public variables.

ugsync-util/src/main/java/org/apache/ranger/ugsyncutil/transform/RegEx.java
dhavalshah9131: No need of import AbstractMapper is inherited.

agents-common/src/main/java/org/apache/ranger/plugin/model/UgsyncNameTransformRules.java
dhavalshah9131: Removed class [UgsyncNameTransformRules.java].
Its not needed .

Thanks for pinting on it.

ugsync-util/pom.xml
```diff
@@ -47,6 +47,28 @@
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
```
mneethiraj: Are libraries `log4j-api` and `log4j-over-slf4j` needed in ugsync-util module? If not, please remove these dependencies.

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPluginContext.java
```diff
@@ -40,6 +41,10 @@ public class RangerPluginContext {
     private final RangerPluginConfig                                                         config;
     private final Map<String, Map<RangerPolicy.RangerPolicyResource, RangerResourceMatcher>> resourceMatchers = new HashMap<>();
     private final ReentrantReadWriteLock                                                     lock             = new ReentrantReadWriteLock(true); // fair lock
+    private       Mapper                                                                     userNameTransformInst;
```
mneethiraj: I suggest moving new members to `RangerAuthContext` class.

ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
```diff
@@ -1435,6 +1396,37 @@ public boolean isSyncSourceValidationEnabled() {
         return isSyncSourceValidationEnabled;
     }
 
+    public Map<String, String> getNameTransformationRules() {
```
mneethiraj: Methods  `getNameTransformationRules()` and `getAllRegexPatternsConfig()` don't seem to be used. Please review and remove.

ugsync-util/src/main/java/org/apache/ranger/ugsyncutil/transform/RegEx.java
```diff
@@ -69,7 +62,7 @@ public String transform(String attrValue) {
         return result;
     }
 
-    protected void populateReplacementPatterns(String baseProperty, List<String> regexPatterns, String regexSeparator) {
+    public void populateReplacementPatterns(String baseProperty, List<String> regexPatterns, String regexSeparator) {
```
mneethiraj: This is referenced from this class, and its test class `TestRegEx`. I suggest changing visibility to package-private i.e. remove `public`. Btw, there is a typo in the test class package name: `org.apache.ranger.ugsynutil.transform` => `org.apache.ranger.ugsyncutil.transform` (missing `c` in `ugsyncutil`).

security-admin/src/main/java/org/apache/ranger/common/PropertiesUtil.java
```diff
@@ -526,4 +544,72 @@ protected void processProperties(ConfigurableListableBeanFactory beanFactory, Pr
...
+        String ret = UgsyncCommonConstants.DEFAULT_MAPPING_SEPARATOR;
+        String val = PropertiesUtil.getProperty(RangerCommonConstants.PLUGINS_CONF_MAPPING_SEPARATOR);
+        if (StringUtils.isNotEmpty(val)) {
+            if (val.length() == 1) {
+                ret = val;
+            } else {
+                LOG.warn("More than one character found in RegEx Separator, using default RegEx Separator");
+            }
+        }
+        LOG.info(String.format("Using %s as the RegEx Separator", ret));
```
mneethiraj: Replace use of `String.format` with parameterized log message.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
```diff
@@ -157,6 +174,52 @@ public void enrich(RangerAccessRequest request) {
         }
     }
 
+    private String getTransformedUser(PolicyEngine policyEngine, RangerAccessRequest request) {
+        String user                   = request.getUser();
+        String userNameCaseConversion = policyEngine.getPluginContext().getUserNameCaseConversion();
+
+        if (UgsyncCommonConstants.UGSYNC_LOWER_CASE_CONVERSION_VALUE.equalsIgnoreCase(userNameCaseConversion)) {
```
mneethiraj: Consider avoiding string comparision on every authz call; instead have the configured parsed to a boolean in RangerAuthContext and access boolean from here.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
```diff
@@ -157,6 +174,52 @@ public void enrich(RangerAccessRequest request) {
...
+        if (groupNameTransformInst == null || request.getUserGroups() == null) {
+            return request.getUserGroups();
+        }
+
+        return request.getUserGroups().stream()
+                .filter(Objects::nonNull)
+                .map(group -> {
+                    String originalGroup = group;
+
+                    if (UgsyncCommonConstants.UGSYNC_LOWER_CASE_CONVERSION_VALUE.equalsIgnoreCase(caseConversion)) {
```
mneethiraj: Consider avoiding string comparision on every authz call, and for each group; instead have the configured parsed to a boolean in RangerAuthContext and access boolean from here.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerDefaultRequestProcessor.java
```diff
@@ -157,6 +174,52 @@ public void enrich(RangerAccessRequest request) {
...
+        }
+
+        return user;
+    }
+
+    private Set<String> getTransformedGroups(PolicyEngine policyEngine, RangerAccessRequest request) {
+        Mapper groupNameTransformInst = policyEngine.getPluginContext().getGroupNameTransformInst();
+        String caseConversion         = policyEngine.getPluginContext().getGroupNameCaseConversion();
+
+        if (groupNameTransformInst == null || request.getUserGroups() == null) {
```
mneethiraj: When `groupNameTransformInst` is null, `return` at line 202 will result in case-conversion to be not applied. Please review and update.

dhavalshah9131 (commented):
I agree with @mneethiraj changes.

PR #585, opened 2025-06-05
RANGER-4993 : Ranger KMS - Missing HSTS Headers for 404 Not found requests port 9494

dhavalshah9131:
RANGER-4993 : Ranger KMS - Missing HSTS Headers for 404 Not found requests port 9494

HSTS header is missing for 404 Not found requests .

E.G.
curl -ivk -u  keyadmin:Admin123  [https://<fqdn>:9494/](https://%3Cfqdn%3E:9494/)

OR

curl -ivk --negotiate -u : [https://<fqdn>:9494](https://%3Cfqdn%3E:9494/)

```
< HTTP/1.1 404 Not Found
HTTP/1.1 404 Not Found
< Content-Type: text/html;charset=utf-8
Content-Type: text/html;charset=utf-8
< Content-Language: en
Content-Language: en
< Content-Length: 682
Content-Length: 682
< Date: Mon, 26 Aug 2024 13:12:21 GMT
Date: Mon, 26 Aug 2024 13:12:21 GMT
< Server: Apache Ranger
Server: Apache Ranger
```

## How was this patch tested?

Successfully validated the HSTS header for above API and all the existing APIs such as GET keys, create keys, rollover keys , delete keys.

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/HSTSFilter.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/HSTSFilter.java
@@ -34,7 +34,6 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
 public class HSTSFilter implements Filter {
-
     static final Logger LOG = LoggerFactory.getLogger(HSTSFilter.class);
 
     @Override
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/HSTSFilter.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/HSTSFilter.java
@@ -44,16 +44,13 @@ public class HSTSFilter implements Filter {
     @Override
     public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
             throws IOException, ServletException {
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("===> HSTSFilter:doFilter()");
-        }
+        LOG.debug("===> HSTSFilter:doFilter()");
         String path = ((HttpServletRequest) request).getRequestURI();
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("==> HSTSFilter:doFilter() path = " + path);
-        }
+        LOG.debug("==> HSTSFilter:doFilter() path = " + path);
         HttpServletResponse resp = (HttpServletResponse) response;
         resp.setHeader("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload");
         chain.doFilter(request, response);
+        LOG.debug("<=== HSTSFilter:doFilter()");
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


kms/src/main/webapp/WEB-INF/web.xml
```diff
@@ -18,56 +18,66 @@
...
+    </filter-mapping>
+
+    <filter-mapping>
+        <filter-name>MDCFilter</filter-name>
+        <url-pattern>/kms/*</url-pattern>
+    </filter-mapping>
+
+    <filter-mapping>
+        <filter-name>HSTSFilter</filter-name>
+        <url-pattern>/*</url-pattern>
```
vikaskr22: @dhavalshah9131  , For any URL starting with /KMS , All filters will match the pattern including "/*" , and header would be updated twice.
Either we should remove the logic to update "Strict-Transport-Security" header from auth filter or pattern for HSTSFilter should be only "/"

dhavalshah9131: Hi @vikaskr22 ,

Made the changes.
HSTS Filter will be invoke first now.

kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/HSTSFilter.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public void init(FilterConfig filterConfig) throws ServletException {
+        // Initialization logic if needed
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+            throws IOException, ServletException {
+        if (LOG.isDebugEnabled()) {
```
vikaskr22: Is this IF condition really required ? I remember as part of code-style-changes , we started using unconditional DEBUG statement for simple logging.

kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/HSTSFilter.java
```diff
@@ -0,0 +1,63 @@
...
+            LOG.debug("===> HSTSFilter:doFilter()");
+        }
+        String path = ((HttpServletRequest) request).getRequestURI();
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("==> HSTSFilter:doFilter() path = " + path);
+        }
+        HttpServletResponse resp = (HttpServletResponse) response;
+        resp.setHeader("Strict-Transport-Security", "max-age=31536000; includeSubDomains; preload");
+        chain.doFilter(request, response);
+    }
```
vikaskr22, line 54: Should we add one return log statement like following:

LOG.debug("<=== HSTSFilter:doFilter()");

vikaskr22 (changes requested):
I have added few minor review comments, other things look good to me.

vikaskr22 (approved):
Looks good to me now.

PR #592, opened 2025-06-17
RANGER-5224: dedupTags removes the valid tags while deduplicating tags

vyommani:

When tag de duplication is enabled in Apache Ranger, deleting and recreating one resource causes the tag-based policy to fail for another resource that retains the same tag in Apache Atlas. After recreating the first resource, a user with access via the tag-based policy is unexpectedly denied access to the second resource, despite the tag still being associated with it.
 
In the dedupTags() method, if a tag’s ID is higher than the retained ID, the tag is removed from the tags map instead of updating its ID. This can invalidate the tag’s mapping for the second resource after the first resource’s deletion, breaking the policy.

## How was this patch tested?

Modified the extsting test(TestServiceTags.java) and newly added test will fail without fix and passes after fix.

### DIFF
--- a/agents-common/src/test/java/org/apache/ranger/plugin/util/TestServiceTags.java
+++ b/agents-common/src/test/java/org/apache/ranger/plugin/util/TestServiceTags.java
@@ -25,8 +25,11 @@ import org.junit.Test;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 public class TestServiceTags {
     private static final RangerServiceResource[] RESOURCES = {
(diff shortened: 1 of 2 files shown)


agents-common/src/test/java/org/apache/ranger/plugin/util/TestServiceTags.java
```diff
@@ -112,6 +115,61 @@ public void testDedup_DupTagsWithAttr_MultipleCalls() {
...
+                new RangerTag("PCI", Collections.emptyMap()),
+                new RangerTag("PII", Collections.singletonMap("type", "email"))
+        };
+
+        ServiceTags svcTags = createServiceTags(tags, RESOURCES);
+        assertEquals(5, svcTags.getTags().size());
+        // Should remove 3 duplicates (2 PII, 1 PCI)
+        assertEquals(3, svcTags.dedupTags());
+
+        // Verify: 2 tags remain (one PII, one PCI)
```
rameeshm, line 134: Should we have test coverage where only 1 tags is there and it gets removed?

agents-common/src/test/java/org/apache/ranger/plugin/util/TestServiceTags.java
```diff
@@ -112,6 +115,61 @@ public void testDedup_DupTagsWithAttr_MultipleCalls() {
...
+        RangerTag newTag = new RangerTag("PII", Collections.singletonMap("type", "email"));
+        long newTagId = 23L;
+        newTag.setId(newTagId);
+        svcTags1.getTags().put(newTagId, newTag);
+        svcTags1.getResourceToTagIds().get(1L).add(newTagId);
+
+        assertEquals(0, svcTags1.dedupTags());
+        assertEquals(2, svcTags1.getTags().size());
+        assertEquals(cachedTag, svcTags1.getTags().get(piiTagId));
+        assertFalse(svcTags1.getTags().containsKey(newTagId));
```
fateh288, line 160: @vyommani 
assertEquals(cachedTag, svcTags1.getTags().get(piiTagId));
I don't understand why this assertion holds true. 
piiTagId has been removed frm svcTags1 in line#150, then why should svcTags1 should still have this tag id ?

fateh288 (commented):
@vyommani @mneethiraj 
Can you please explain me a few things in dedup tags (a lot of my comments might be uninformed of the original design decisions and might not be for this PR but maybe another one)- 
My understanding is that we use "cachedTags" to deduplicate the "tags". "tags" should still the source of truth for any tag based operations and should be accessed via setter/getter.
But when I see the logic, I see cachedTags being used in RangerServiceTagsDeltaUtil for applyDelta (was this by design ? -- shouldn't we somehow try to use dedup method and maybe another method in the ServiceTags class instead of having this logic of applying deltas in RangerServiceTagsDeltaUtil) - was there a reason why "cachedTags" was not made private  -- also shouldn't only one of the two "tags" vs "cachedTags" be exposed via getter/setter and the other be totally private and only for bookkeeping purposes)

agents-common/src/test/java/org/apache/ranger/plugin/util/TestServiceTags.java
vyommani: The assertion assertEquals(cachedTag, svcTags1.getTags().get(piiTagId)) holds true because when svcTag1.dedupTags() is called at line 157, it updates the tag ID of the newly added tag (with a higher ID) to match the cached tag ID (piiTagId). Although piiTagId was removed from svcTags1 at line 150, the dedupTags() method effectively restores it by updating the new tag's ID to the cached ID, allowing the assertion to pass.

agents-common/src/test/java/org/apache/ranger/plugin/util/TestServiceTags.java
vyommani: done

fateh288 (approved):
The changes LGTM.
@vyommani can you please add a few lines of comments / docstring for the dedup function for future reference. Thanks !!

pradeepagrawal8184 (approved):

rameeshm (approved):

PR #593, opened 2025-06-18
RANGER-3174: Support for strong cryptography algorithm-PBKDF2WithHmacSHA256

vikaskr22:
…PBE encryption

Currently Ranger KMS supports following Password based encryption  (PBE) to encrypt/decrypt MasterKey and Zone keys:
**PBEWithMD5AndDES, PBEWithMD5AndTripleDES**

Both algorithms are weak .

With this commit, now KMS will start supporting PBKDF2WithHmacSHA256 algorithm if FIPS is enabled. This algorithm is FIPS complaint and being provided by many SecurityProviders. 
Currently I have kept this if FIPS is enabled, means, if keyStoreType is **"bcfks"**

So if FIPS is enabled, then following will happen:
1. All new keys material will be encrypted/decrypted using PBKDF2WithHmacSHA256 algorithm.
2. Any older keys (including MasterKey) will be re-encrypted using  PBKDF2WithHmacSHA256 on KMS service start. This is an one time operation. Here re-encryption means, decrypting the older keys using older algorithm and re-encryption it using the current PBKDF2WithHmacSHA256 algorithm.

Please note that key material is not changing, only it is being re-encrypted before storing into DB. hence, there is no risk of data loss.
 

## How was this patch tested?
1. basic mvn build and UnitTest cases.
2. Manual testing in docker setup. I used following steps to prepare  docker container for this testing:
   -  I used **bc-fips-2.0.0.jar** as SecurityProvider for PBKDF2WithHmacSHA256
   -  Copied **bc-fips-2.0.0.jar** jar to the `/usr/lib/jvm/java-8-openjdk-arm64/jre/lib/ext/`
   - Update java.security file at `/etc/java-8-openjdk/security/java.security` to contain following content:
        
        security.provider.1=org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider
        org.bouncycastle.fips.approved_only=true

   - **For rangerkms.jceks** (that contains MasterKey password and DB password) : created corresponding keystore in bcfks formart named  rangerkms.localbcfks
   - **dbks-site.xml** :  Updated following properties in this file:
        
        Key: ranger.ks.jpa.jdbc.credential.provider.path
        Value: localbcfks://file/opt/ranger/kms/ews/webapp/WEB-INF/classes/conf/rangerkms.localbcfks
        
        Key: ranger.keystore.file.type
        Value: bcfks

**Scenarios Tested:**

1. **Fresh setup:** Masterkey and Zone keys got created using latest algorithm. And basic encryption/decryption was working fine.
2. **Cluster having old keys with older algorithm:**
      - Here created one zone key (zonekey1) and one encryption Zone and copied one test file inside the zone key
      - Means, file got encrypted using DEK , and this DEK was encrypted using corresponding zone keys . This zone keys was   encrypted using PBEWithMD5AndDES.
      - Then, created a new ranger-kms docker image with new changes. And started the container.
      -  Here Masterkey and all zone keys got re-encypted using new algorithm.
      -  Files kept inside encryptionZone was accessible after keys re-encryption.
      -  Also, basic key life cycle like create/rotate/delete was also working.

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
@@ -27,6 +27,7 @@ import org.apache.hadoop.crypto.key.RangerKeyStoreProvider.KeyMetadata;
 import org.apache.ranger.entity.XXRangerKeyStore;
 import org.apache.ranger.kms.dao.DaoManager;
 import org.apache.ranger.kms.dao.RangerKMSDao;
+import org.apache.ranger.plugin.util.JsonUtilsV2;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,7 +38,6 @@ import javax.crypto.SealedObject;
 import javax.crypto.SecretKey;
 import javax.crypto.SecretKeyFactory;
 import javax.crypto.spec.PBEKeySpec;
-import javax.crypto.spec.PBEParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 import javax.xml.bind.DatatypeConverter;
 
@@ -78,6 +78,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -95,16 +96,23 @@ public class RangerKeyStore extends KeyStoreSpi {
     private static final String  AZURE_KEYVAULT_ENABLED  = "ranger.kms.azurekeyvault.enabled";
     private static final String  METADATA_FIELDNAME      = "metadata";
     private static final int     NUMBER_OF_BITS_PER_BYTE = 8;
-    private static final String  SECRET_KEY_HASH_WORD    = "Apache Ranger";
+    private static final String SECRET_KEY_HASH_WORD     = "Apache Ranger";
+    public  static final String KEY_CRYPTO_ALGO_NAME     = "keyCryptoAlgoName";
 
     private final    RangerKMSDao        kmsDao;
     private final    RangerKMSMKI        masterKeyProvider;
     private final    boolean             keyVaultEnabled;
+    private          boolean             isFIPSEnabled;
     private final    Map<String, Object> deltaEntries = new ConcurrentHashMap<>();
     private volatile Map<String, Object> keyEntries   = new ConcurrentHashMap<>();
 
     public RangerKeyStore(DaoManager daoManager) {
         this(daoManager, false, null);
+    }
+
+    public RangerKeyStore(boolean isFIPSEnabled, DaoManager daoManager) {
+        this(daoManager);
+        this.isFIPSEnabled = isFIPSEnabled;
     }
 
     public RangerKeyStore(DaoManager daoManager, Configuration conf, KeyVaultClient kvClient) {
@@ -131,7 +139,7 @@ public class RangerKeyStore extends KeyStoreSpi {
 
         if (entry instanceof SecretKeyEntry) {
             try {
-                ret = unsealKey(((SecretKeyEntry) entry).sealedKey, password);
+                ret = unsealKey((SecretKeyEntry) entry, password);
             } catch (Exception e) {
                 logger.error("engineGetKey({}) error", alias, e);
             }
@@ -470,13 +478,20 @@ public class RangerKeyStore extends KeyStoreSpi {
         logger.debug("<== addSecureKeyByteEntry({})", alias);
     }
 
+    private String addEncrAlgoNameInKeyAttrib(String jsonAttrib) throws Exception {
+        Map<String, String> attribMap = JsonUtilsV2.jsonToMap(jsonAttrib);
+        attribMap.put(KEY_CRYPTO_ALGO_NAME, SupportedPBECryptoAlgo.PBKDF2WithHmacSHA256.getAlgoName());
+        return JsonUtilsV2.mapToJson(attribMap);
+    }
+
     public void addKeyEntry(String alias, Key key, char[] password, String cipher, int bitLength, String description, int version, String attributes) throws KeyStoreException {
         logger.debug("==> addKeyEntry({})", alias);
 
         SecretKeyEntry entry;
 
         try {
-            entry = new SecretKeyEntry(sealKey(key, password), cipher, bitLength, description, version, attributes);
+            String updatedAttribute = isFIPSEnabled ? addEncrAlgoNameInKeyAttrib(attributes) : attributes;
+            entry = new SecretKeyEntry(sealKey(key, password), cipher, bitLength, description, version, updatedAttribute);
         } catch (Exception e) {
             logger.error("addKeyEntry({}) error", alias, e);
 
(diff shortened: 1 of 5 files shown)


kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -470,13 +479,20 @@ public void addSecureKeyByteEntry(String alias, Key key, String cipher, int bitL
         logger.debug("<== addSecureKeyByteEntry({})", alias);
     }
 
+    private String addEncrAlgoNameInKeyAttrib(String jsonAttrib) throws Exception {
+        Map<String, String> attribMap = JsonUtilsV2.jsonToMap(jsonAttrib);
+        attribMap.put(KEY_ENCR_ALGO_NAME, SupportedPBECryptoAlgo.PBKDF2WithHmacSHA256.getAlgoName());
```
rameeshm: For readability please consider KEY_ENCR_ALGO_NAME => KEY_CRYPTO_ALGO_NAME or KEY_ENCRYPT_ALGO_NAME, please update the variable name as well.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerMasterKey.java
```diff
@@ -171,16 +175,22 @@ public void init() {
...
-        mkKeySize      = getIntConfig("ranger.kms.service.masterkey.password.size", DEFAULT_MK_KeySize);
-        saltSize       = getIntConfig("ranger.kms.service.masterkey.password.salt.size", DEFAULT_SALT_SIZE);
-        salt           = getConfig("ranger.kms.service.masterkey.password.salt", DEFAULT_SALT);
-        pbeAlgo        = getConfig("ranger.kms.service.masterkey.password.encryption.algorithm", DEFAULT_CRYPT_ALGO);
-        mdAlgo         = getConfig("ranger.kms.service.masterkey.password.md.algorithm", defaultMdAlgo);
-        iterationCount = getIntConfig("ranger.kms.service.masterkey.password.iteration.count", DEFAULT_ITERATION_COUNT);
-        paddingString  = Joiner.on(",").skipNulls().join(mkCipher, mkKeySize, saltSize, pbeAlgo, mdAlgo, iterationCount, salt);
-
+        isFipsEnabled       = getConfig("ranger.keystore.file.type", KeyStore.getDefaultType()).equalsIgnoreCase("bcfks");
+        defaultMdAlgo       = isFipsEnabled ? "SHA-512" : "MD5";
```
rameeshm, line 179: Should we consider having strong default Algorithm instead of MD5?

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
vikaskr22: Thanks for the feedback. I have incorporated the suggestion.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerMasterKey.java
vikaskr22: @rameeshm , Agree. In fact this is the plan but not only for MessageDigest.
If you see,  default is  **SHA-512** for FIPS cluster.  I am working on one more JIRA where KMS will will use all FIPS compliant algorithm and related parameters for all environments, not only for FIPS.
Some examples, like, the way we generate Secret/IV etc.

To answer this question, default is SHA-512 for FIPS. MD5 is for non-fips.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -919,59 +935,107 @@ private SealedObject sealKey(Key key, char[] password) throws Exception {
         logger.debug("==> sealKey()");
```
spolavarpau1: Can you add class name here as well similar to line #964

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -919,59 +935,107 @@ private SealedObject sealKey(Key key, char[] password) throws Exception {
...
-        logger.debug("<== sealKey(): ret={}", ret);
+        logger.debug("<== RangerKeyStore.sealKey()");
 
-        return ret;
+        return new RangerSealedObject(key, cipher, salt);
     }
 
-    private Key unsealKey(SealedObject sealedKey, char[] password) throws Exception {
+    private Key unsealKey(SecretKeyEntry secretKeyEntry, char[] password) throws Exception {
         logger.debug("==> unsealKey()");
```
spolavarpau1: Please add classname here as well to be in sync with other debug logs

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -919,59 +935,107 @@ private SealedObject sealKey(Key key, char[] password) throws Exception {
...
 
-        // Create SecretKey
-        SecretKeyFactory secretKeyFactory = SecretKeyFactory.getInstance("PBEWithMD5AndTripleDES");
-        PBEKeySpec       pbeKeySpec       = new PBEKeySpec(password);
-        SecretKey        secretKey        = secretKeyFactory.generateSecret(pbeKeySpec);
+        // fetch encryption algo name
+        String encrAlgoName = JsonUtilsV2.jsonToMap(secretKeyEntry.attributes).get(KEY_CRYPTO_ALGO_NAME);
+        SupportedPBECryptoAlgo encrAlgo;
+        if (Objects.nonNull(encrAlgoName) && !encrAlgoName.isEmpty()) {
```
spolavarpau1: Any reason why we are not using StringUtils.isEmpty() instead?

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
vikaskr22: thanks for the review. Suggestion has been incorporated.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
vikaskr22: thanks for the review. Suggestion has been incorporated.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
vikaskr22: thanks for the review. Suggestion has been incorporated.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStoreProvider.java
```diff
@@ -123,6 +128,10 @@ public RangerKeyStoreProvider(Configuration conf) throws Throwable {
         final DaoManager   daoManager  = rangerKMSDB.getDaoManager();
         final RangerKMSMKI masterKeyProvider;
 
+        String storeType = conf.get("ranger.keystore.file.type", KeyStore.getDefaultType());
+        this.isFIPSEnabled = StringUtils.equalsIgnoreCase("bcfks", storeType) ? true : false;
+        logger.info("isFIPSEnabled=" + isFIPSEnabled);
```
kumaab: Please avoid string concatenation, please check other references as well.

vikaskr22: Incorporated the review comments. Thanks.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -916,62 +931,110 @@ private String convertAlias(String alias) {
...
-        logger.debug("<== sealKey(): ret={}", ret);
+    private Key unsealKey(SecretKeyEntry secretKeyEntry, char[] password) throws Exception {
+        logger.debug("==> RangerKeyStore.unsealKey()");
 
-        return ret;
-    }
+        // fetch encryption algo name
+        String encrAlgoName = JsonUtilsV2.jsonToMap(secretKeyEntry.attributes).get(KEY_CRYPTO_ALGO_NAME);
+        SupportedPBECryptoAlgo encrAlgo;
+        if (StringUtils.isNotEmpty(encrAlgoName)) {
```
dhavalshah9131: Can we change this using Ternary Operator


`SupportedPBECryptoAlgo encrAlgo = StringUtils.isNotEmpty(encrAlgoName) ? SupportedPBECryptoAlgo.valueOf(encrAlgoName) : SupportedPBECryptoAlgo.PBEWithMD5AndTripleDES`

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -916,62 +931,110 @@ private String convertAlias(String alias) {
...
+        // Get the AlgorithmParameters from RangerSealedObject
+        AlgorithmParameters algorithmParameters = null;
+        byte[] salt = null;
+        PBEKeySpec pbeKeySpec = null;
+
+        if (SupportedPBECryptoAlgo.PBKDF2WithHmacSHA256.equals(encrAlgo)) {
+            salt = ((RangerSealedObject) sealedKey).getSalt();
+            pbeKeySpec = new PBEKeySpec(password, salt, 20, encrAlgo.getKeyLength());
+        } else { // not yet re-encrypted, older algo
+            if (sealedKey instanceof RangerSealedObject) {
```
dhavalshah9131: Can we change this using Ternary Operator

`algorithmParameters = (sealedKey instanceof RangerSealedObject)
        ? ((RangerSealedObject) sealedKey).getParameters(encrAlgo.getAlgoName())
        : new RangerSealedObject(sealedKey).getParameters(encrAlgo.getAlgoName());`

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
vikaskr22: Fixed. Thanks.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
vikaskr22: Fixed. Thanks.

spolavarpau1 (approved):

PR #595, opened 2025-06-19
RANGER-5230: Support synchronous policy cache download via service co…

fateh288:
…nfigs for testing plugins

Proposes new config via service config `ranger.plugin.<serviceType>.policy.cache.refresh.mode = sync` that can be used to synchronously download policies during policy evaluation to reduce time waiting for plugin tests to download latest policies

## How was this patch tested?
Not yet tested. Requesting initial feedback

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -102,6 +102,7 @@ public class RangerBasePlugin {
     private       RangerRoles                 roles;
     private       boolean                     isUserStoreEnricherAddedImplcitly;
     private       Map<String, String>         serviceConfigs;
+    private boolean synchronousPolicyCacheUpdateFlag;
 
     public RangerBasePlugin(String serviceType, String appId) {
         this(new RangerPluginConfig(serviceType, null, appId, null, null, null));
@@ -444,6 +445,9 @@ public class RangerBasePlugin {
 
         this.serviceConfigs = (policies != null && policies.getServiceConfig() != null) ? policies.getServiceConfig() : new HashMap<>();
 
+        this.synchronousPolicyCacheUpdateFlag = isSynchronousPolicyCacheUpdateEnabled();
+        LOG.debug("isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+
         if (pluginConfig.isEnableImplicitUserStoreEnricher() && policies != null && !ServiceDefUtil.isUserStoreEnricherPresent(policies)) {
             String retrieverClassName = pluginConfig.get(RangerUserStoreEnricher.USERSTORE_RETRIEVER_CLASSNAME_OPTION, RangerAdminUserStoreRetriever.class.getCanonicalName());
             String retrieverPollIntMs = pluginConfig.get(RangerUserStoreEnricher.USERSTORE_REFRESHER_POLLINGINTERVAL_OPTION, Integer.toString(60 * 1000));
@@ -653,6 +657,11 @@ public class RangerBasePlugin {
     }
 
     public RangerAccessResult isAccessAllowed(RangerAccessRequest request, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyCacheUpdateFlag) {
+            LOG.debug("==> isAccessAllowed isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+            refreshPoliciesAndTags();
+        }
+
         RangerAccessResult ret          = null;
         RangerPolicyEngine policyEngine = this.policyEngine;
 
@@ -694,6 +703,11 @@ public class RangerBasePlugin {
     }
 
     public Collection<RangerAccessResult> isAccessAllowed(Collection<RangerAccessRequest> requests, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyCacheUpdateFlag) {
+            LOG.debug("==> isAccessAllowed isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+            refreshPoliciesAndTags();
+        }
+
         Collection<RangerAccessResult> ret          = null;
         RangerPolicyEngine             policyEngine = this.policyEngine;
 
@@ -735,6 +749,11 @@ public class RangerBasePlugin {
     }
 
     public RangerAccessResult evalDataMaskPolicies(RangerAccessRequest request, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyCacheUpdateFlag) {
+            LOG.debug("==> evalDataMaskPolicies isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+            refreshPoliciesAndTags();
+        }
+
         RangerPolicyEngine policyEngine = this.policyEngine;
         RangerAccessResult ret          = null;
 
@@ -765,6 +784,11 @@ public class RangerBasePlugin {
     }
 
     public RangerAccessResult evalRowFilterPolicies(RangerAccessRequest request, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyCacheUpdateFlag) {
+            LOG.debug("==> evalRowFilterPolicies isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+            refreshPoliciesAndTags();
+        }
+
         RangerPolicyEngine policyEngine = this.policyEngine;
         RangerAccessResult ret          = null;
 
@@ -795,6 +819,11 @@ public class RangerBasePlugin {
     }
 
     public void evalAuditPolicies(RangerAccessResult result) {
+        if (this.synchronousPolicyCacheUpdateFlag) {
+            LOG.debug("==> evalAuditPolicies isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+            refreshPoliciesAndTags();
+        }
+
         RangerPolicyEngine policyEngine = this.policyEngine;
 
         if (policyEngine != null) {
@@ -817,6 +846,11 @@ public class RangerBasePlugin {
     }
 
     public RangerResourceACLs getResourceACLs(RangerAccessRequest request, Integer policyType) {
+        if (this.synchronousPolicyCacheUpdateFlag) {
+            LOG.debug("==> getResourceACLs isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyCacheUpdateFlag);
+            refreshPoliciesAndTags();
+        }
+
         RangerResourceACLs ret          = null;
         RangerPolicyEngine policyEngine = this.policyEngine;
 
@@ -1442,5 +1476,19 @@ public class RangerBasePlugin {
     private static final class LogHistory {
         long lastLogTime;
         int  counter;
+    }
+
+    private boolean isSynchronousPolicyCacheUpdateEnabled() {
+        boolean ret = false;
+        if (this.getServiceConfigs() != null && this.pluginConfig != null && this.pluginConfig.getServiceType() != null) {
+            String configPolicyCacheRefreshMode = String.format("ranger.plugin.%s.policy.cache.refresh.mode", this.pluginConfig.getServiceType());
+            if (this.getServiceConfigs().containsKey(configPolicyCacheRefreshMode)) {
+                String policyCacheRefreshMode = this.getServiceConfigs().get(configPolicyCacheRefreshMode).trim().toLowerCase();
+                if (policyCacheRefreshMode.equals("sync") || policyCacheRefreshMode.equals("synchronous")) {
+                    ret = true;
+                }
+            }
+        }
+        return ret;
     }
 }


fateh288:
@dhavalshah9131 
I think you implemented a way to add service configs for all services using a single config at ranger admin, can you share details on that (is there a specific prefix required for that ?) -- can that be used here ?

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1455,4 +1492,18 @@ private static final class LogHistory {
         long lastLogTime;
         int  counter;
     }
+
+    private boolean isSynchronousPolicyCacheUpdateEnabled() {
+        boolean ret = false;
+        if (this.getServiceConfigs() != null && this.pluginConfig != null && this.pluginConfig.getServiceType() != null) {
+            String configPolicyCacheRefreshMode = "ranger.plugins.conf.policy.cache.refresh.mode";
```
mneethiraj: - to be consistent with rest of policy related plugin configurations, I suggest using properyy prefix `ranger.plugin.<serviceType>.policy.`
-  I suggest renaming `cache.refresh.mode` to `refresh.synchronous`, with default value as `faise`
   `ranger.plugin.hive.policy.refresh.synchronous`
- `isSynchronousPolicyCacheUpdateEnabled()` => `isSynchronousPolicyRefresh()`

fateh288: @mneethiraj 
1) The convention that "ranger admin configs" starting with ranger.plugins.conf will sync as service configs to all the plugins has been introduced in [RANGER-5215](https://issues.apache.org/jira/browse/RANGER-5215):
https://github.com/apache/ranger/blob/master/security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java#L233 
The property prefix you are mentioning will be used when the property needs to be set at "per service level". Anything starting with ranger.plugins.conf will be synced to all services -- thus I have used it here

2) Addressed
3) Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -103,6 +103,7 @@ public class RangerBasePlugin {
     private       RangerRoles                 roles;
     private       boolean                     isUserStoreEnricherAddedImplcitly;
     private       Map<String, String>         serviceConfigs;
+    private boolean synchronousPolicyRefreshFlag;
```
mneethiraj: `Flag` in the name is unnecessary. I suggest renaming to `isSynchronousPolicyRefresh`.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -653,6 +657,11 @@ public Collection<RangerAccessResult> isAccessAllowed(Collection<RangerAccessReq
     }
 
     public RangerAccessResult isAccessAllowed(RangerAccessRequest request, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyRefreshFlag) {
+            LOG.debug("==> isAccessAllowed isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyRefreshFlag);
```
mneethiraj: Is this log necessary, especially given the flag will always be `true` here. I suggest removing the log here, and add inside `refreshPoliciesAndTags()` if its already not present.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -694,6 +703,11 @@ public RangerAccessResult isAccessAllowed(RangerAccessRequest request, RangerAcc
     }
 
     public Collection<RangerAccessResult> isAccessAllowed(Collection<RangerAccessRequest> requests, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyRefreshFlag) {
+            LOG.debug("==> isAccessAllowed isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyRefreshFlag);
```
mneethiraj: Is this log necessary, especially given the flag will always be `true` here. I suggest removing the log here, and add inside `refreshPoliciesAndTags()` if its already not present.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -735,6 +749,11 @@ public Collection<RangerAccessResult> isAccessAllowed(Collection<RangerAccessReq
     }
 
     public RangerAccessResult evalDataMaskPolicies(RangerAccessRequest request, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyRefreshFlag) {
+            LOG.debug("==> evalDataMaskPolicies isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyRefreshFlag);
```
mneethiraj: Is this log necessary, especially given the flag will always be `true` here. I suggest removing the log here, and add inside `refreshPoliciesAndTags()` if its already not present.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -765,6 +784,11 @@ public RangerAccessResult evalDataMaskPolicies(RangerAccessRequest request, Rang
     }
 
     public RangerAccessResult evalRowFilterPolicies(RangerAccessRequest request, RangerAccessResultProcessor resultProcessor) {
+        if (this.synchronousPolicyRefreshFlag) {
+            LOG.debug("==> evalRowFilterPolicies isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyRefreshFlag);
```
mneethiraj: Is this log necessary, especially given the flag will always be `true` here. I suggest removing the log here, and add inside `refreshPoliciesAndTags()` if its already not present.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -795,6 +819,11 @@ public RangerAccessResult evalRowFilterPolicies(RangerAccessRequest request, Ran
     }
 
     public void evalAuditPolicies(RangerAccessResult result) {
+        if (this.synchronousPolicyRefreshFlag) {
+            LOG.debug("==> evalAuditPolicies isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyRefreshFlag);
```
mneethiraj: Is this log necessary, especially given the flag will always be `true` here. I suggest removing the log here, and add inside `refreshPoliciesAndTags()` if its already not present.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -817,6 +846,11 @@ public RangerResourceACLs getResourceACLs(RangerAccessRequest request) {
     }
 
     public RangerResourceACLs getResourceACLs(RangerAccessRequest request, Integer policyType) {
+        if (this.synchronousPolicyRefreshFlag) {
+            LOG.debug("==> getResourceACLs isSynchronousPolicyCacheUpdateEnabled = {}", this.synchronousPolicyRefreshFlag);
```
mneethiraj: Is this log necessary, especially given the flag will always be `true` here. I suggest removing the log here, and add inside `refreshPoliciesAndTags()` if its already not present.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
mneethiraj: This configuration is used in the plugin side (and not in Ranger admin server). Hence, I suggest using existing convention in naming configurations - `"ranger.plugin." + serviceType + ".policy.refresh.synchronous"`.

fateh288: But we want to set it at ranger admin level via ranger-admin-site.xml and not via service configs from Ranger UI / repo settings. 
If we want to set this config on a per service basis from Ranger UI, then we can do what you are recommending --- my initial implementation / commit in this PR had this implementation but I changed it after discussion with @spolavarapu1 
We can go either ways depending on our need - whether we want to set it on per service level from ranger UI service configs or at once via ranger admin configs

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: `This configuration is used in the plugin side (and not in Ranger admin server) `==>. this is not true --- ranger admin configs starting with "ranger.plugins.conf " are now synced to all the plugins too

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1490,7 +1490,7 @@ private static final class LogHistory {
     private boolean isSynchronousPolicyRefresh() {
         boolean ret = false;
         if (this.getServiceConfigs() != null && this.pluginConfig != null && this.pluginConfig.getServiceType() != null) {
-            String synchronousPolicyRefreshConfigName = "ranger.plugins.conf.policy.refresh.synchronous";
+            String synchronousPolicyRefreshConfigName = String.format("ranger.plugin.%s.policy.refresh.synchronous", this.pluginConfig.getServiceType());
             if (this.getServiceConfigs().containsKey(synchronousPolicyRefreshConfigName)) {
                 boolean synchronousPolicyRefreshConfigValue = Boolean.parseBoolean(this.getServiceConfigs().get(synchronousPolicyRefreshConfigName).trim().toLowerCase());
                 if (synchronousPolicyRefreshConfigValue) {
```
mneethiraj: `this.getServiceConfigs().get(synchronousPolicyRefreshConfigName)` would return null if the configuration is not present in the map `this.getServiceConfigs()` right? In that case line 1495 will hit NPE. Can you review and update?

fateh288: I have this check before trying to access the config
`if (this.getServiceConfigs().containsKey(synchronousPolicyRefreshConfigName))`

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -583,6 +584,9 @@ public void setPolicies(ServicePolicies policies) {
 
                     setServiceConfigs(policies.getServiceConfig());
 
+                    this.synchronousPolicyRefresh = isSynchronousPolicyRefresh();
```
vyommani: Thank you for sharing the proposed change to reduce test wait times for policy/tag synchronization in RangerBasePlugin. I understand the goal is to minimize delays by introducing a configuration  to control policy refreshes. However, I have concerns about the approach of calling refreshPoliciesAndTags() within isAccessAllowed() when the synchronousPolicyRefresh flag is enabled.

 isAccessAllowed() is a public API designed solely to evaluate access permissions. Embedding a call to refreshPoliciesAndTags(), even if guarded by a flag, introduces a secondary responsibility of policy/tag synchronisation. This violates SRP and makes the method’s behaviour less predictable, as callers may not expect a refresh to occur.

 Both isAccessAllowed() and refreshPoliciesAndTags() are public methods with distinct purposes. Mixing their responsibilities, even under a configuration flag, could lead to unintended side effects. Since isAccessAllowed() is a performance-critical API, enabling synchronous refreshes could significantly slow down production systems if the flag is misconfigured by a client. This risk is particularly concerning for a public API where users expect consistent and predictable behaviour. 

I’m curious did you explore calling refreshPoliciesAndTags() explicitly before invoking isAccessAllowed(). While modifying every test to include an explicit 
refresh may not be ideal, other options could be considered, such as, Adjust the default sync interval (30s to 200ms or 1 second) to minimize wait times without
embedding refresh logic in isAccessAllowed(). 

Have you conducted performance tests to evaluate the impact of this change? Even if the additional if condition for the flag adds minimal overhead (a few
nanoseconds ), understanding the performance implications is critical, particularly for a public API used in production environments.

I recommend keeping isAccessAllowed() focused on its core responsibility and exploring alternative ways to achieve the goal of reducing test wait times. Let’s maintain the separation of concerns between these public APIs to ensure clarity, predictability, and performance.

Please let me know your thoughts or if there are additional considerations (e.g., specific constraints that led to this design). I’d be happy to discuss further or assist in exploring other solutions.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
mneethiraj: I suggest replacing lines 1492 - 1496 with the following:

```
Map<String, String> serviceConfigs      = this.serviceConfigs;
String              isSyncPolicyRefresh = serviceConfigs == null ? null : serviceConfigs.get(pluginConfig.getPropertyPrefix() + ".policy.refresh.synchronous");

if (Boolean.parseBoolean(isSyncPolicyRefresh)) {
```

Also, I suggest moving above code into `RangerBasePlugin.setServiceConfig()` i.e. remove `isSynchronousPolicyRefresh()` method.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: @vyommani 
Thanks a lot for your comments.
We cannot call refreshPoliciesAndTags from each test as the synchronous policy download config has been developed for ranger integration tests and not unit tests wherein we have control of when to call this method. The components still will call call only isAccessAllowed and not any other methods.
As you mentioned, the performance impact will be minimal since its just a boolean check.
The config is not exposed via UI for any component. The config can only be enabled by ranger admins so while there is a possibility of misconfiguration, I consider it as a low risk as misconfiguration of any property in general can lead to performance degradation. 
I think that configuration management by admins or any privileged users is in general a safe idea as one signs up for any side effects when any config changes are made.
@mneethiraj can add any additional comments on what is his opinion.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: @mneethiraj 
Addressed

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1186,6 +1211,9 @@ public GdsPolicyEngine getGdsPolicyEngine() {
     }
 
     public Map<String, String> getServiceConfigs() {
+        if (serviceConfigs == null) {
```
mneethiraj: Why is it necessary to return a new map in case of `null`?

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -1209,6 +1237,11 @@ private void setServiceConfigs(Map<String, String> serviceConfigs) {
         if (authContext != null && !Objects.equals(oldServiceConfigs, this.serviceConfigs)) {
             authContext.onServiceConfigsUpdate(this.serviceConfigs);
         }
+        String isSyncPolicyRefresh = this.pluginConfig == null ? null : this.serviceConfigs.get(this.pluginConfig.getPropertyPrefix() + ".policy.refresh.synchronous");
+        if (Boolean.parseBoolean(isSyncPolicyRefresh)) {
+            this.synchronousPolicyRefresh = true;
```
mneethiraj: When `this.synchronousPolicyRefresh` is `true`, and the new value is `false` the `if` at 1241 will result in  `this.synchronousPolicyRefresh` not updated. I suggest replacing 1241 - 1243 with the following line:
```
  this.synchronousPolicyRefresh = Boolean.parseBoolean(isSyncPolicyRefresh);
```

fateh288: Sorry for this oversight. 
Fixed in the patch

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
vyommani: @fateh288 

Have you considered adding a new method as shown below?

public RangerAccessResult isAccessAllowedSync(RangerAccessRequest request) {
    refreshPoliciesAndTags();
    return isAccessAllowed(request);
}
This approach keeps the existing isAccessAllowed method unchanged and allows you to use the new method to reduce the execution time of your integration tests. If you decide against adding a new method, please provide a report detailing any performance impact of your code changes. I strongly recommend adding the new method rather than modifying the existing one.

mneethiraj: @vyommani  - with this approach (of having `isAccessAllowedSync()` methods), every plugin must be updated to call `*Sync()` method version instead of  existing calls to`isAccessAllowed()` methods when a flag is set. This will be a significant task, involving updates to large number of modules (20+ plugins?). The performance impact of additional `if` on a boolean member in each authz call should be negligible. I suggest avoiding updates to large number of modules, and go with the approach in this PR.

agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
fateh288: There were a couple of reasons I changed this code:
The getServiceConfigs method is being used by hbase plugin wherein there was no null check and could have thrown NPE. 
While there is an effort in the setServiceConfigs method to set it as a new hashmap when the underying service config is null, there is a possibility that this setter is not reached in the flow --- recently this change/refactoring was made (previously it was always set to new hashmap and hence there was no possibility of NPE in the caller method)
In order for the callers of the getServiceConfig (I am not sure but there could be other callers other than hbase plugin too) method to not face NPE, I am returning new hashmap to the caller of this method instead of null.

mneethiraj: Ok. In that case, I suggest returning `Collections.emptyMap()`, instead of creating a new `HashMap` instance.

fateh288: Addressed

PR #605, opened 2025-07-12
RANGER-3174: Support for strong cryptography algorithm-PBKDF2WithHmacSHA256

kumaab:

(cherry picked from commit bf2ca0f192c8cf382e5aab56cadf86777c43e4d5)

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/RangerMasterKey.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/RangerMasterKey.java
@@ -21,6 +21,7 @@ import java.security.Key;
 import java.security.KeyStore;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.security.spec.AlgorithmParameterSpec;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


kms/src/main/java/org/apache/hadoop/crypto/key/RangerMasterKey.java
```diff
@@ -49,18 +49,23 @@
...
+
     public static final String DBKS_SITE_XML                  = "dbks-site.xml";
 
     private static String password = null;
-    private static String DEFAULT_MD_ALGO;
+    private static String defaultMdAlgo;
+    private static boolean isFipsEnabled;
 
     public static String MK_CIPHER;
     public static Integer MK_KeySize = 0;
```
vikaskr22, line 71: @kumaab , It seems this PR doesn't contain the code-style changes commit ( 05d4ff8ee37fb7d6712cf7fb6c473e45e3074358 ). merging this PR before code-style changes may create many code conflicts. If we are planning to take that commit, I would suggest to merge that first and then this PR. This feature was developed on top of that.

vikaskr22 (changes requested):

PR #628, opened 2025-08-15
RANGER-3696 java.lang.NoClassDefFoundError: org/slf4j/LoggerFactory

dugujiujiang:

(Please fill in changes proposed in this fix. Create an issue in ASF JIRA before opening a pull request and
set the title of the pull request which starts with
the corresponding JIRA issue number. (e.g. RANGER-XXXX: Fix a typo in YYY))

## How was this patch tested?

### DIFF
--- a/ranger-elasticsearch-plugin-shim/pom.xml
+++ b/ranger-elasticsearch-plugin-shim/pom.xml
@@ -55,12 +55,25 @@
 			<artifactId>commons-lang</artifactId>
 			<version>${commons.lang.version}</version>
 		</dependency>
-
 		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
-			<version>${slf4j-api.version}</version>
-			<scope>runtime</scope>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-api</artifactId>
+			<version>${slf4j2.version}</version>
+		</dependency>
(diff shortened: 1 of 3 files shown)


ranger-elasticsearch-plugin-shim/pom.xml
```diff
@@ -55,12 +55,25 @@
 			<artifactId>commons-lang</artifactId>
 			<version>${commons.lang.version}</version>
 		</dependency>
-
 		<dependency>
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
-			<version>${slf4j-api.version}</version>
-			<scope>runtime</scope>
+			<groupId>org.apache.logging.log4j</groupId>
```
mneethiraj, line 59: `ranger-elasticsearch-plugin-shim.jar` shouldn't require log4j implementation libraries; Sources in this library only refer to `org.slf4j.Logger` and `org.slf4j.LoggerFactory`; so depdency on `slf4j-api` should be the enough. The implementation libraries, log4j for Elasticsearch, should be picked up at runtime from Elasticsearch classpath. It shouldn't be necessary to package them with Ranger plugin.

Also, looking at the stack trace in RANGER-3696, it seems Elastisearch already has class isolation in place for plugins (so that libraries used in the plugin are not visible to Elasticsearch). If this is true, then I suggest getting rid of ranger-elasticsearch-plugin-shim library altogether.

PR #629, opened 2025-08-16
RANGER-5287:Add sort by added user Id and action for the admin audit log

rameeshm:

Ranger admin audits where its stores the history of CRUD operations done on the ranger resource and policies updates needs to be sorted by added user and action done. This is more useful to track the changes done for datasets by sorting the logs.

## How was this patch tested?

tested in the Docker container where ranger admin is brought up.


dineshkumar-yadav (commented):
Please consider using 076  index for patch file across the db, as 075 is already taken.

rameeshm:
@kumaab  @dineshkumar-yadav  Saw your comments now. I shall address this in subsequent patch for the same

PR #640, opened 2025-08-25
RANGER-5292: Add Unit Tests for Ranger Metrics Package in Security Ad…

RakeshGuptaDev:
…min Module

### AI Disclosure   
Added AI-generated Test methods/Test classes for the org.apache.ranger.metrics and org.apache.ranger.metrics.source modules using Cursor AI and manually reviewed.

Added unit tests for classes in the org.apache.ranger.metrics and org.apache.ranger.metrics.source packages within the Security Admin module.

## How was this patch tested?

Build completed successfully along with all test cases using the command: mvn clean compile package install.

As per Jacoco
Apache -> Security-Admin Module

Packages Covered

org.apache.ranger.metrics 	      -> 0% to 100% Coverage

org.apache.ranger.metrics.source -> 0% to 100% Coverage


ChinmayHegde24 (approved):
Everything looks good, It would be better if you just add the code coverage after this patch in description part, Thanks

PR #646, opened 2025-08-28
RANGER-5295: fix GDS policy evaluation handling of row-filters

mneethiraj:

GDS policy evaluation updated to evaluate row-filter policies even when GDS policy does not grant access to all sub-resources of a resource.

## How was this patch tested?

added unit tests to cover this use case; verified that all tests pass

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
@@ -1072,12 +1072,35 @@ public class RangerPolicyEngineImpl implements RangerPolicyEngine {
 
         if (gdsResult != null) {
             if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_ACCESS) {
+                // pick access result from GDS policies only if there is no decision yet
                 if (!result.getIsAccessDetermined() && gdsResult.getIsAllowed()) {
                     result.setIsAllowed(true);
                     result.setIsAccessDetermined(true);
                     result.setPolicyId(gdsResult.getPolicyId());
                     result.setPolicyVersion(gdsResult.getPolicyVersion());
                     result.setPolicyPriority(RangerPolicy.POLICY_PRIORITY_NORMAL);
+                }
+            } else if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_ROWFILTER) {
+                // pick row-filter from GDS policies only if there is no decision yet
+                if (result.getPolicyId() != -1 && CollectionUtils.isNotEmpty(gdsResult.getRowFilters())) {
+                    result.setIsAllowed(true);
+                    result.setIsAccessDetermined(true);
+                    result.setPolicyId(gdsResult.getPolicyId());
+                    result.setPolicyVersion(gdsResult.getPolicyVersion());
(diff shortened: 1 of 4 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPolicyEngineImpl.java
```diff
@@ -1072,13 +1072,36 @@ private void updateFromGdsResult(RangerAccessResult result) {
...
             if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_ACCESS) {
+                // pick access result from GDS policies only if there is no decision yet
                 if (!result.getIsAccessDetermined() && gdsResult.getIsAllowed()) {
                     result.setIsAllowed(true);
                     result.setIsAccessDetermined(true);
                     result.setPolicyId(gdsResult.getPolicyId());
                     result.setPolicyVersion(gdsResult.getPolicyVersion());
                     result.setPolicyPriority(RangerPolicy.POLICY_PRIORITY_NORMAL);
                 }
+            } else if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_ROWFILTER) {
```
vyommani, line 1079: updateFromGdsResult contains lots of repetitive code , particularly handling the different policy type. Can we simplify the logic  by extracting the common code in a helper function and call the method instead. 

 private void updateCommonFields(RangerAccessResult result, GdsAccessResult gdsResult) {
        result.setIsAllowed(true);
        result.setIsAccessDetermined(true);
        result.setPolicyId(gdsResult.getPolicyId());
        result.setPolicyVersion(gdsResult.getPolicyVersion());
        result.setPolicyPriority(RangerPolicy.POLICY_PRIORITY_NORMAL);
    }

    private void    updateFromGdsResult(RangerAccessResult result) {
        LOG.debug("==> updateFromGdsResult(result={})", result);

        RangerAccessRequest request   = result.getAccessRequest();
        GdsAccessResult     gdsResult = request != null ? RangerAccessRequestUtil.getGdsResultFromContext(request.getContext()) : null;

        if (gdsResult != null) {
            if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_ACCESS) {
                // pick access result from GDS policies only if there is no decision yet
                if (!result.getIsAccessDetermined() && gdsResult.getIsAllowed()) {
                   updateCommonFields(result, gdsResult);
                }
            }else if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_ROWFILTER) {
                // pick row-filter from GDS policies only if there is no decision yet
                if (result.getPolicyId() == -1 && CollectionUtils.isNotEmpty(gdsResult.getRowFilters())) {
                    updateCommonFields(result, gdsResult);
                    result.setFilterExpr(gdsResult.getRowFilters().get(0));
                }
            } else if (result.getPolicyType() == RangerPolicy.POLICY_TYPE_DATAMASK) {
                // pick data-mask from GDS policies only if there is no decision yet
                if (result.getPolicyId() == -1 && StringUtils.isNotEmpty(gdsResult.getMaskType())) {
                    updateCommonFields(result, gdsResult);
                    result.setMaskedValue(gdsResult.getMaskedValue());
                    result.setMaskCondition(gdsResult.getMaskCondition());
                }
            }

mneethiraj: Good point. Thanks for the suggestion @vyommani.

PR #650, opened 2025-09-02
RANGER-3965: Handling for DB runtime error when key attribute is very…

vikaskr22:

This PR contains fix for RANGER-3965. If key attribute is very long ( more than 1024, schema defined column length), KMS logs exception (as expected) but same exception is not being propagated. On UI, it was showing as key created but failed to list.  
This fix propagates the exception so that API returns error instead of simply logging the exception and doing nothing.

## How was this patch tested?
-  KMS build and UT
- Created new KMS docker image and manually verified the scenario.

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
@@ -529,6 +529,7 @@ public class RangerKeyStore extends KeyStoreSpi {
             }
         } catch (Exception e) {
             logger.error("dbOperationStore({}) error", rangerKeyStore.getAlias(), e);
+            throw new RuntimeException("Error while storing object in the DB. " + e.getMessage());
         }
 
         logger.debug("<== dbOperationStore({})", rangerKeyStore.getAlias());
--- a/kms/src/main/java/org/apache/ranger/kms/dao/BaseDao.java
+++ b/kms/src/main/java/org/apache/ranger/kms/dao/BaseDao.java
@@ -117,6 +117,7 @@ public abstract class BaseDao<T> {
             logger.error("create({}) failed", tClass.getSimpleName(), e);
 
             rollbackTransaction();
+            throw e;
         }
 
         return ret;


kms/src/main/java/org/apache/hadoop/crypto/key/RangerKeyStore.java
```diff
@@ -529,6 +529,7 @@ public void dbOperationStore(XXRangerKeyStore rangerKeyStore) {
             }
         } catch (Exception e) {
             logger.error("dbOperationStore({}) error", rangerKeyStore.getAlias(), e);
+            throw new RuntimeException("Error while storing object in the DB. " + e.getMessage());
```
vyommani: Please throw the new RuntimeException with the message and cause as follows
       throw new RuntimeException("Error while storing object in the DB.", e);
This will make sure that propagated exception will have original stack trace.

vikaskr22: @vyommani , Agree. I missed that. Thanks. Incorporated the review comments.

kms/src/main/java/org/apache/ranger/kms/dao/BaseDao.java
```diff
@@ -117,6 +117,7 @@ public T create(T obj) {
             logger.error("create({}) failed", tClass.getSimpleName(), e);
 
             rollbackTransaction();
+            throw e;
```
vyommani, line 120: create does not throw Exception (public T create(T obj) {), will this work ?. I suggest you to wrap 'e' to RuntimeException and then throw.

vyommani: In case of any exception, code calls rollbacktransaction() do  you think you need to throw the exception after roll back ?.

vikaskr22: @vyommani ,  Thanks again for your time. You have raised two points:
1. **On throwing the exception as it is instead of wrapping inside RuntimeException:** Yes, the current approach works. And the reason is, there is not any checked exception being thrown from the method or the API being used inside that method. EntityManager is being used for DB operations like commit/rollback etc, if you see this API doesn't throw any checked exception. In some cases it throws only RuntimeExceptions. And also, if there is any API being used throws any checked exceptions, it will be a compile time error. So this approach is working.
2. **Why exception is being thrown after rollbacktransaction() :**  Yes, this is required. And this is the bug. 
The current issue is, Key being created with very long attribute ( beyond column size limits) is being shown as successful on UI but key actually doesn't exist. Actually, due to DB runtime error, it is being rolled back but service layer API is not getting any exception or error code, so it was returning 200 OK from the REST. That is incorrect. here I am trying to propagate the runtime exception (instead of simply logging ) so that API returns 500, internal server error instead of 200 OK.

I hope it's clear now. Please let me know if there are any other concerns.

vikaskr22: Discussed and agreed to resolve the conversation.

PR #659, opened 2025-09-03
RANGER-5309: add authz-api module

mneethiraj:

Introduces authorization API for use by applications that wants to leverage Ranger policies to authorize access to their resources. The API will have at least 2 implementations: 1) embedded policy evaluation 2) remote policy evaluation in Ranger PDP server

## How was this patch tested?

Added unit tests.

### DIFF
--- a/authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
+++ b/authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
@@ -1,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.ranger.authz.util;
+
+import org.apache.ranger.authz.api.RangerAuthzException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.ranger.authz.api.RangerAuthzApiErrorCode.INVALID_RESOURCE_TEMPLATE_UNEXPECTED_MARKER_AT;
+
+public class RangerResourceTemplate {
+    private static final Logger LOG = LoggerFactory.getLogger(RangerResourceTemplate.class);
+
+    private static final char RESOURCE_START_CHAR = '{';
+    private static final char RESOURCE_END_CHAR   = '}';
+    private static final char ESCAPE_CHAR         = '\\';
+
+    private final String   template;
+    private final String[] resources;
+    private final String   prefix;
+    private final String[] separators;
+    private final String   suffix;
+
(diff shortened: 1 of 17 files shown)


authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
```diff
@@ -0,0 +1,184 @@
...
+
+                currentToken.setLength(0);
+            } else {
+                currentToken.append(c);
+            }
+        }
+
+        this.template   = template;
+        this.resources  = resources.toArray(new String[0]);
+        this.separators = separators.toArray(new String[0]);
```
fateh288, line 98: Why can't we directly use ArrayList as the class variable (for resources and separators here) instead of trying to convert List <String> into String[]

authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
```diff
@@ -0,0 +1,184 @@
...
+        return template;
+    }
+
+    public Map<String, String> parse(String resource) {
+        Map<String, String> ret = null;
+
+        if (resource == null || resource.isEmpty()) {
+            LOG.debug("parse(resource='{}', template='{}'): empty or null resource", resource, template);
+        } else if (!resource.startsWith(prefix)) {
+            LOG.debug("parse(resource='{}', template='{}'): resource does not start with prefix {}", resource, template, prefix);
```
fateh288: Won't it be better to throw some exception here (currently null is being returned) if we are unable to parse the resource (e.g. invalid suffix/prefix/any other parsing error) 
Throwing exception will avoid caller method to add null checks which the caller might miss and may result in NPE. Caller is expecting Map<String,String> based on return type of method

authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
```diff
@@ -0,0 +1,184 @@
...
+        } else if (!resource.endsWith(suffix)) {
+            LOG.debug("parse(resource='{}', template='{}'): resource does not end with suffix {}", resource, template, suffix);
+        } else {
+            ret = new HashMap<>();
+
+            int tokenStartPos = prefix.length();
+
+            for (int i = 0; i < this.separators.length; i++) {
+                String sep    = this.separators[i];
+                int    idxSep = resource.indexOf(sep, tokenStartPos);
```
fateh288, line 123: This might be more of a corner case 
Consider a template: {a}@{b}.{c}
Lets consider a resource {abc.def}@{pqr.stu}.{xyz}

resource.indexOf(sep, tokenStartPos); --> This will break I think when we try to get index of "." as we will get index of "." which is between pqr.stu instead of "." between stu}.{xyz

Wouldn't it be better to limit the separator to be a single string in a template instead of a list of strings as we have in our current implementation ? Do we have a use case currently wherein there are 2 or more separators in a template ?

authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
mneethiraj: Accessing entries via array index will likely be a little more efficient than accessing via get(idx) method of a list; hence using array instead of list. Also, arrays make it clear that the size don't change after initialization, but this can be done with unmodifiable collections as well..a little more overhead :-)

authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
mneethiraj: Good idea! It will help the caller to get details of the failure. I will update shortly. Thanks!

authz-api/src/main/java/org/apache/ranger/authz/util/RangerResourceTemplate.java
mneethiraj: > Wouldn't it be better to limit the separator to be a single string in a template instead of a list of strings as we have in our current implementation

Resources like, ADLS-Gen2 paths have format container-name@account-name/path/to/file.txt; hence the template should be flexible enough to handle such resources. 

> Consider a template: {a}@{b}.{c}
> Lets consider a resource {abc.def}@{pqr.stu}.{xyz}

This is not the valid resource string. Given "." is the separator between resources `b` and `c`, resource `b` can not have the separator in its value.

PR #664, opened 2025-09-11
RANGER-5319: Add sleep for passive tagsync to avoid cpu usage

fateh288:
Currently passive tagsync continuously checks whether it the active instance or not in an infinite loop (for quick failover) with 0 seconds delay which causes high CPU usage (150-200% i.e. 1.5 - 2 cores of CPU as noted from the top command).

This needs to be fixed by adding sleep similar to how passive usersync is doing currently

## How was this patch tested?

Manually tested. CPU usage goes down to 0% with periodic usage of 0.3% to check if it active instance or not every 5 seconds

### DIFF
--- a/tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
+++ b/tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
@@ -180,6 +180,18 @@ public class TagAdminRESTSink implements TagSink, Runnable {
 
                     return;
                 }
+            } else {
+                // Only sleep when HA is enabled, similar to user sync
+                if (TagSyncConfig.getInstance().isTagSyncHAEnabled()) {
+                    try {
+                        long sleepInterval = TagSyncConfig.getTagSyncHAPassiveSleepInterval();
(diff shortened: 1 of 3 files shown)


tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
```diff
@@ -180,6 +180,18 @@ public void run() {
 
                     return;
                 }
+            } else {
+                // Only sleep when HA is enabled, similar to user sync
+                if (TagSyncConfig.getInstance().isTagSyncHAEnabled()) {
```
mneethiraj, line 185: Is `isTagSyncHAEnabled()` needed here? `TagSyncConfig.isTagSyncServiceActive()` at line 146 can return `false` only in HA mode.

fateh288: I think there are multiple states in the enum : 
```
public enum ServiceStateValue {
		ACTIVE, PASSIVE, BECOMING_ACTIVE, BECOMING_PASSIVE, MIGRATING
}
```
I think what @mneethiraj might be correct that another method may not be required
@dineshkumar-yadav -- what do you think -- can the state be non active even in non HA mode ?

PR #667, opened 2025-09-13
RANGER-5322: PolicyRefresher can struck with no download thread if exception thrown in startRefresher

vyommani:

Handled the exception thrown in loadRoles() and loadPolicy()

## How was this patch tested?
run mvn clean install and all tests are clean.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/util/PolicyRefresher.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/util/PolicyRefresher.java
@@ -128,16 +128,8 @@ public class PolicyRefresher extends Thread {
     }
 
     public void startRefresher() {
-        try {
-            loadRoles();
-        } catch (Exception e) {
-            LOG.error("Initial roles load failed for serviceName={}: {}, will retry in {}ms via scheduled refresh", serviceName, e, pollingIntervalMs);
-        }
-        try {
-            loadPolicy();
-        } catch (Exception e) {
-            LOG.error("Initial policy load failed for serviceName={}: {}, will retry in {}ms via scheduled refresh", serviceName, e, pollingIntervalMs);
-        }
+        loadRoles();
+        loadPolicy();
         initRefresher();
     }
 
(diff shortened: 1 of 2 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/util/PolicyRefresher.java
```diff
@@ -128,23 +128,41 @@ public void setLastActivationTimeInMillis(long lastActivationTimeInMillis) {
...
-
             LOG.debug("Scheduled policyDownloadRefresher to download policies every {} milliseconds", pollingIntervalMs);
-        } catch (IllegalStateException exception) {
-            LOG.error("Error scheduling policyDownloadTimer:", exception);
+        } catch (IllegalArgumentException | IllegalStateException | NullPointerException e) {
+            LOG.error("Error scheduling policyDownloadTimer:", e);
             LOG.error("*** Policies will NOT be downloaded every {} milliseconds ***", pollingIntervalMs);
-
-            policyDownloadTimer = null;
+            if (policyDownloadTimer != null) {
```
mneethiraj: Will `policyDownloadTimer` be null here, given the initialization in line 152 above?

agents-common/src/main/java/org/apache/ranger/plugin/util/PolicyRefresher.java
```diff
@@ -128,23 +128,41 @@ public void setLastActivationTimeInMillis(long lastActivationTimeInMillis) {
     }
 
     public void startRefresher() {
-        loadRoles();
-        loadPolicy();
-
-        super.start();
+        try {
```
mneethiraj: `loadRoles()` and `loadPolicy()` are called from `run()` method as well. Consider moving `try/catch` blocks inside `loadRoles()` and `loadPolicy()` implementations.

agents-common/src/main/java/org/apache/ranger/plugin/util/PolicyRefresher.java
vyommani: I was cautious and checked for null before calling policyDownloadTimer.cancel(). I think javac will optimize out this unnecessary null check. I'll review this code in my updated pull request.

agents-common/src/main/java/org/apache/ranger/plugin/util/PolicyRefresher.java
vyommani: good point will do.

mneethiraj (approved):

pradeepagrawal8184 (approved):

PR #668, opened 2025-09-15
RANGER-5319: Follow-up: Check for tagsync HA not required

fateh288:

Refactor previous commit https://github.com/apache/ranger/commit/24a47ce8a96005a12a8bf9d8b1f09a48a2b8ed94 as check for HA not required

## How was this patch tested?

mvn clean install

### DIFF
--- a/tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
+++ b/tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
@@ -187,6 +187,8 @@ public class TagAdminRESTSink implements TagSink, Runnable {
                     Thread.sleep(sleepInterval);
                 } catch (InterruptedException interrupted) {
                     LOG.error("Interrupted..: ", interrupted);
+                    // preserve interrupt status for caller of the thread
+                    Thread.currentThread().interrupt();
                     return;
                 }
             }
--- a/tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
+++ b/tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
@@ -181,16 +181,13 @@ public class TagAdminRESTSink implements TagSink, Runnable {
                     return;
                 }
             } else {
-                // Only sleep when HA is enabled, similar to user sync
-                if (TagSyncConfig.getInstance().isTagSyncHAEnabled()) {
-                    try {
-                        long sleepInterval = TagSyncConfig.getTagSyncHAPassiveSleepInterval();
-                        LOG.debug("Sleeping for [{}] milliSeconds as this server is running in passive mode", sleepInterval);
-                        Thread.sleep(sleepInterval);
-                    } catch (InterruptedException interrupted) {
-                        LOG.error("Interrupted..: ", interrupted);
-                        return;
-                    }
+                try {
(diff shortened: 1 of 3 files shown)


mneethiraj (approved):

tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
```diff
@@ -181,16 +181,13 @@ public void run() {
...
-                        Thread.sleep(sleepInterval);
-                    } catch (InterruptedException interrupted) {
-                        LOG.error("Interrupted..: ", interrupted);
-                        return;
-                    }
+                try {
+                    long sleepInterval = TagSyncConfig.getTagSyncHAPassiveSleepInterval();
+                    LOG.debug("Sleeping for [{}] milliSeconds as this server is running in passive mode", sleepInterval);
+                    Thread.sleep(sleepInterval);
+                } catch (InterruptedException interrupted) {
```
vyommani, line 188: Catching InterruptedException resets the thread's interrupted flag. It is always a good idea to preserve the interrupt status, please call Thread.currentThread().interrupt();

tagsync/src/main/java/org/apache/ranger/tagsync/source/atlas/AtlasTagSource.java
```diff
@@ -232,16 +232,13 @@ public void run() {
...
-                        } catch (InterruptedException interrupted) {
-                            LOG.error("Interrupted: ", interrupted);
-                            return;
-                        }
+                    try {
+                        long sleepInterval = TagSyncConfig.getTagSyncHAPassiveSleepInterval();
+                        LOG.debug("Sleeping for [{}] milliSeconds as this server is running in passive mode", sleepInterval);
+                        Thread.sleep(sleepInterval);
+                    } catch (InterruptedException interrupted) {
+                        LOG.error("Interrupted: ", interrupted);
```
vyommani, line 240: Same as above.

tagsync/src/main/java/org/apache/ranger/tagsync/sink/tagadmin/TagAdminRESTSink.java
fateh288: @vyommani 
Can you please elaborate on this -- when we do this and when we don't do this ?
We might also want to change other occurrences where we are catching the exception -- we are not doing explicit interrupt at other places within this class and other places in the code

vyommani: Thanks for the question! My suggestion to call Thread.currentThread().interrupt() when catching InterruptedException follows Java’s concurrency best practices to preserve the thread’s interrupt status, ensuring that higher-level code (e.g., thread pools or executors) can detect interruptions and handle them appropriately (e.g., for graceful shutdown). 

When to call Thread.currentThread().interrupt() When catching InterruptedException in code that should propagate the interrupt signal to callers, such as in a thread pool or long-running thread. Restoring the interrupt status allows frameworks like ExecutorService or calling code to detect the interruption and take action (e.g., exit a loop or terminate a task).

Even I have see in Ranger code we are not doing at all, it can be good start.

fateh288: @vyommani 
Thanks for clarification. Addressed.

vyommani (approved):
LGTM

PR #680, opened 2025-09-22
RANGER-5333:Configurable Master key name for Ranger KMS DB with Luna HSM

ChinmayHegde24:
Master Key name is hard coded for Ranger KMS DB integration with Luna HSM.
Refer : https://github.com/apache/ranger/blob/master/kms/src/main/java/org/apache/hadoop/crypto/key/RangerHSM.java#L50

It is made configurable so CU can provide customised name.

Test:
 Successful creation of new master key with configured name
 Noticed duplicate creation of key error if created with existing Master key name (as expected)
 Successful mvn build

### DIFF
--- a/kms/config/kms-webapp/dbks-site.xml
+++ b/kms/config/kms-webapp/dbks-site.xml
@@ -405,6 +405,12 @@
   	<value>false</value>
   	<description></description>
   </property>
+
+  <property>
+        <name>ranger.ks.hsm.masterkey.alias</name>
+        <value></value>
+        <description>Custom alias for Luna HSM master key</description>
+  </property>
   
   <property>
   	<name>ranger.ks.hsm.partition.name</name>
--- a/kms/config/kms-webapp/dbks-site.xml
+++ b/kms/config/kms-webapp/dbks-site.xml
@@ -408,7 +408,7 @@
 
   <property>
         <name>ranger.ks.hsm.masterkey.alias</name>
-        <value></value>
+        <value>RangerKMSKey</value>
         <description>Custom alias for Luna HSM master key</description>
   </property>
   
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/RangerHSM.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/RangerHSM.java
@@ -48,11 +48,14 @@ public class RangerHSM implements RangerKMSMKI {
     private static final String PARTITION_NAME          = "ranger.ks.hsm.partition.name";
     private static final String HSM_TYPE                = "ranger.ks.hsm.type";
     private static final String MK_KEY_SIZE             = "ranger.kms.hsm.masterkey.size";
-    private static final String ALIAS                   = "RangerKMSKey";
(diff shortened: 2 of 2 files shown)


kms/config/kms-webapp/dbks-site.xml
```diff
@@ -405,6 +405,12 @@
   	<value>false</value>
   	<description></description>
   </property>
+
+  <property>
+        <name>ranger.ks.hsm.masterkey.alias</name>
+        <value></value>
```
vikaskr22: @ChinmayHegde24 , can you please confirm following :
In dbks-site.xml, you are keeping this as empty. And in RangerHSM.java, how will following behave:
this.alias = conf.get(MK_ALIAS, DEFAULT_ALIAS); ?

is it possible that it will initialise this.alias="" ? As per API , if this key doesn't exist, then default will be used. But here key will exist with empty value. 
Pls check and confirm this case.
Other things look good to me.

ChinmayHegde24: @vikaskr22 in that case value will be null not "" hence it will fallback to default alias.
But I'll move that default alias to xml property itself in order to be sync with other properties.
Thank you for pointing it out.

vikaskr22 (approved):
LGTM.

dhavalshah9131 (commented):
Hi @ChinmayHegde24 ,

Ref link : [Thales doc](https://thalesdocs.com/gphsm/luna/7/docs/network/Content/admin_hsm/hsm_roles/naming_reqs.htm#HSMlabel)

We should have some kind validation as per Thales guidelines for alias name string value.

Also we need to considering how KMS will behave in case of invalid name or consider documenting it.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerHSM.java
```diff
@@ -155,7 +158,7 @@ public boolean setMasterKey(String password, byte[] key) {
             try {
                 Key aesKey = new SecretKeySpec(key, MK_CIPHER);
 
-                myStore.setKeyEntry(ALIAS, aesKey, password.toCharArray(), (java.security.cert.Certificate[]) null);
+                myStore.setKeyEntry(alias, aesKey, password.toCharArray(), (java.security.cert.Certificate[]) null);
```
vikaskr22: @ChinmayHegde24 , it's not related to your change. But it would be better to add one check to verify if the key alias already exists or not , the way we check in generateMasterKey() method.

@dhavalshah9131 , is there any reason there is no check for key existence ?

vikaskr22:
@ChinmayHegde24 ,  can you pls update the cases you have tested for this patch ? It would be helpful to review this PR.

ChinmayHegde24:
> @ChinmayHegde24 , can you pls update the cases you have tested for this patch ? It would be helpful to review this PR.

Yeah please check it once now @vikaskr22

vikaskr22 (approved):
LGTM. Once CI pipeline is passed, it can be merged.

vikaskr22:
@kumaab , Above build-11 & plugin-docker-build is failing. Is it a known issue ? Can it be merged ?

PR #683, opened 2025-09-24
RANGER-5259: Add REST API to delete stale entries of Plugin info

pradeepagrawal8184:

We are adding two REST APIs in ranger-admin to delete stale plugin info entries:

Delete based on plugin info entry id : plugins/plugins/info/{id}

Example: if the plugin info entry id is 9 which need to be deleted then below REST shall work. 

` curl -ivk -u admin:Admin123 -H "Accept: application/json" -H "Content-Type: application/json" -X DELETE https://<ranger_host>:6182/service/plugins/plugins/info/9`

Delete based on plugin info entry attributes : plugins/plugins/info

Case-1: if you want to delete all plugin info entries of a ranger service.

`curl -ivk -u admin:Admin123 -H "Accept: application/json" -H "Content-Type: application/json" -X DELETE "https://<ranger_host>:6182/service/plugins/plugins/info?serviceName=${SERVICE_NAME}"`

Case-2: if you want to delete all plugin info entries of a ranger service and host name.

`curl -ivk -u admin:Admin123 -H "Accept: application/json" -H "Content-Type: application/json" -X DELETE "https://<ranger_host>:6182/service/plugins/plugins/info?serviceName=${SERVICE_NAME}&pluginHostName=${HOST_NAME}"`

Case-3: if you want to delete all plugin info entries of a ranger service, host and application name/type.

`curl -ivk -u admin:Admin123 -H "Accept: application/json" -H "Content-Type: application/json" -X DELETE "https://<ranger_host>:6182/service/plugins/plugins/info?serviceName=${SERVICE_NAME}&pluginHostName=${HOST_NAME}&pluginAppType=${APPLICATION_NAME}"`

## How was this patch tested?
 Manually: via calling the above given REST point

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
@@ -2852,6 +2852,48 @@ public class ServiceREST {
         return ret;
     }
 
+    @DELETE
+    @Path("/plugins/info")
+    @Produces("application/json")
+    @PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
+    public void deletePluginsInfo(@Context HttpServletRequest request) {
+        LOG.debug("==> ServiceREST.deletePluginsInfo()");
+
+        RangerPluginInfoList ret    = null;
+        SearchFilter         filter = searchUtil.getSearchFilter(request, pluginInfoService.getSortFields());
+
+        try {
+            PList<RangerPluginInfo> paginatedPluginsInfo = pluginInfoService.searchRangerPluginInfo(filter);
+            if (paginatedPluginsInfo != null) {
+                for (RangerPluginInfo rangerPluginInfo : paginatedPluginsInfo.getList()) {
+                    if (rangerPluginInfo != null) {
+                        deletePluginsInfo(rangerPluginInfo.getId());
+                        LOG.debug("Deleted rangerPluginInfo:[{}]", rangerPluginInfo);
+                    }
+                }
+            }
+        } catch (WebApplicationException excp) {
+            throw excp;
+        } catch (Throwable excp) {
+            LOG.error("deletePluginsInfo() failed", excp);
+
+            throw restErrorUtil.createRESTException(excp.getMessage());
+        }
+
+        LOG.debug("<== ServiceREST.deletePluginsInfo()");
+    }
+
+    @DELETE
+    @Path("/plugins/info/{id}")
+    @PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
+    public void deletePluginsInfo(@PathParam("id") Long id) {
+        LOG.debug("==> ServiceREST.deletePluginsInfo({})", id);
+
+        assetMgr.doDeleteXXPluginInfo(id);
+
+        LOG.debug("<== ServiceREST.deletePluginsInfo() - [id={}]", id);
+    }
+
     public void blockIfGdsService(String serviceName) {
         String serviceType = daoManager.getXXServiceDef().findServiceDefTypeByServiceName(serviceName);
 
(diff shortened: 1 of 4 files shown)


kumaab (commented):
The changes look good, I think it would be good to add unit tests for the new API as well. thanks.

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -2852,6 +2852,48 @@ public RangerPluginInfoList getPluginsInfo(@Context HttpServletRequest request)
...
+            throw restErrorUtil.createRESTException(excp.getMessage());
+        }
+
+        LOG.debug("<== ServiceREST.deletePluginsInfo()");
+    }
+
+    @DELETE
+    @Path("/plugins/info/{id}")
+    @PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
+    public void deletePluginsInfo(@PathParam("id") Long id) {
```
mneethiraj, line 2889: Is it necessary to expose ithis method via REST, given `id` is an internal field (i.e, generated by Ranger/DB)?

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -2852,6 +2852,48 @@ public RangerPluginInfoList getPluginsInfo(@Context HttpServletRequest request)
         return ret;
     }
 
+    @DELETE
+    @Path("/plugins/info")
+    @Produces("application/json")
+    @PreAuthorize("hasRole('ROLE_SYS_ADMIN')")
+    public void deletePluginsInfo(@Context HttpServletRequest request) {
```
mneethiraj, line 2859: I suggest moving this method to AdminREST.

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -2852,6 +2852,48 @@ public RangerPluginInfoList getPluginsInfo(@Context HttpServletRequest request)
...
+
+        try {
+            PList<RangerPluginInfo> paginatedPluginsInfo = pluginInfoService.searchRangerPluginInfo(filter);
+            if (paginatedPluginsInfo != null) {
+                for (RangerPluginInfo rangerPluginInfo : paginatedPluginsInfo.getList()) {
+                    if (rangerPluginInfo != null) {
+                        deletePluginsInfo(rangerPluginInfo.getId());
+                        LOG.debug("Deleted rangerPluginInfo:[{}]", rangerPluginInfo);
+                    }
+                }
```
mneethiraj, line 2873: Add a info level log here:
```
LOG.info("Deleted {} pluginInfo records, for criteria={}", paginatedPluginsInfo.getList().size(), filter);
```

pradeepagrawal8184: Will do in the next patch.

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
pradeepagrawal8184: This can be done.

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
pradeepagrawal8184: This REST is designed keeping in the mind that option can be exposed in future on the RANGER UI to delete the selected row. Currently there is no option in the UI to delete the stale entries. 
If this should not be exposed at all then we can remove the REST part of it and make it private.

if the concern is that such REST APIs should not be there then there are many such APIs which exists in ranger.

mneethiraj: ok. I suggest to introduce this API when we add UI for this.

PR #688, opened 2025-09-26
RANGER-5341: Capture opCode in KMS access logs for EEK_OP operations

ChinmayHegde24:
Currently, when the handleEncryptedKeyOp() method in KMS.java is invoked via the API, there is no visibility into whether the request is performing a decryptKey or reEncryptKey operation as same path gets logged for both operations as it hits the same method.
So after this patch opCode will be included in AccessLog pattern.

e.g In access logs we can see opcode appended
<ip> - - [<DateTime>] "GET /kms/v1/keys/names" 200 3 50 -
<ip> - - [<DateTime>] "GET /kms/v1/key/key1/_eek" 500 288 29 generate
<ip> - - [<DateTime>] "POST /kms/v1/keyversion/test-2@1/_eek" 400 176 16 reencrypt
<ip> - - [<DateTime>] "POST /kms/v1/keyversion/test-2@1/_eek" 400 176 14 decrypt

## How was this patch tested?

Checked Access logs after hitting endpoints through Docker setup
Mvn local build

### DIFF
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSMDCFilter.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSMDCFilter.java
@@ -31,6 +31,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import java.io.IOException;
+import java.net.URLDecoder;
 
 /**
  * Servlet filter that captures context of the HTTP request to be use in the
@@ -39,6 +40,8 @@ import java.io.IOException;
 @InterfaceAudience.Private
 public class KMSMDCFilter implements Filter {
     static final String RANGER_KMS_REST_API_PATH = "/kms/api/status";
+
+    private static final String EEK_OP_CODE = "eek_op";
 
     private static final ThreadLocal<Data> DATA_TL = new ThreadLocal<>();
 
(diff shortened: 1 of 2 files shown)


kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSMDCFilter.java
```diff
@@ -70,15 +78,32 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
 
                 UserGroupInformation ugi         = HttpUserGroupInformation.get();
-                String               method      = ((HttpServletRequest) request).getMethod();
-                StringBuffer         requestURL  = ((HttpServletRequest) request).getRequestURL();
-                String               queryString = ((HttpServletRequest) request).getQueryString();
+                String               method      = req.getMethod();
+                StringBuffer         requestURL  = req.getRequestURL();
+                String               queryString = req.getQueryString();
+
+                // Extract operation from query parameters if present
```
dhavalshah9131, line 91: Can you check for exception and handle it accordingly. E.G. UnsupportedEncodingException.

kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSMDCFilter.java
dhavalshah9131, line 1: Please check for the unit test coverage for new changes.

ChinmayHegde24: Yeah I've added those

kms/src/main/java/org/apache/hadoop/crypto/key/kms/server/KMSMDCFilter.java
ChinmayHegde24: Handled

PR #694, opened 2025-10-01
RANGER-5345: [GDS] Shared resource creation api is failing for oracle db

Sanket-Shelar:

Shared Resource API **/dataset/{id}/resources/{serviceName}** is failing for Oracle database as it is using distinct with CLOB type in query.
Additionally whenever  GDS resource APIs are invoked, there are logs in error as SubResourceMasks and DefaultTagMasks are null.

## How was this patch tested?

Build Successful. API tested in oracle and checked logs and error logs were not present.


agents-common/src/main/java/org/apache/ranger/authorization/utils/JsonUtils.java
```diff
@@ -279,7 +279,7 @@ public static List<RangerGdsMaskInfo> jsonToListGdsMaskInfo(String jsonStr) {
         try {
             return getMapper().readValue(jsonStr, TYPE_LIST_RANGER_GDS_MASK_INFO);
         } catch (Exception e) {
-            LOG.error("Cannot get Map<String, RangerPolicyItemDataMaskInfo> from {}", jsonStr, e);
+            LOG.error("Cannot get List<RangerGdsMaskInfo> from {}", jsonStr, e);
```
rameeshm: Should there be null check on this  "if (StringUtils.isNotEmpty(jsonStr)) " and then do readValue to avoid excetion?

Sanket-Shelar: Update the Patch. kindly review

rameeshm (approved):

PR #729, opened 2025-11-10
RANGER-5394: policyEngine should be volatile to prevent policy update…

vyommani:

Fixed a memory visibility issue in RangerBasePlugin where the policyEngine field was not declared as volatile. This caused a visibility race condition where policy updates made by the policyRefresher thread might not be immediately visible to other threads.

When the policyRefresher thread updates the RangerPolicyEngine reference, the change may not be immediately visible to other threads due to CPU caching and compiler optimizations. This results in threads continuing to use stale policy engine references, which can lead to incorrect authorization decisions and potential access control bypasses.

Without the volatile keyword, there's no happens-before relationship guaranteeing that writes to policyEngine by one thread are visible to reads by other threads. This is a memory visibility issue, not a traditional thread-safety/synchronization issue.

## How was this patch tested?

Added a dedicated concurrency test (TestRangerBasePluginRaceCondition.testVisibilityOfPolicyEngine) that specifically validates the memory visibility fix

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -98,7 +98,7 @@ public class RangerBasePlugin {
     private final List<RangerChainedPlugin>   chainedPlugins;
     private final boolean                     dedupStrings;
     private       PolicyRefresher             refresher;
-    private       RangerPolicyEngine          policyEngine;
+    private volatile RangerPolicyEngine       policyEngine;
     private       RangerAuthContext           currentAuthContext;
     private       RangerAccessResultProcessor resultProcessor;
     private       RangerRoles                 roles;
--- a/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
@@ -99,11 +99,11 @@ public class RangerBasePlugin {
     private final boolean                     dedupStrings;
     private       PolicyRefresher             refresher;
     private volatile RangerPolicyEngine       policyEngine;
-    private       RangerAuthContext           currentAuthContext;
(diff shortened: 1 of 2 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/service/RangerBasePlugin.java
```diff
@@ -98,7 +98,7 @@ public class RangerBasePlugin {
     private final List<RangerChainedPlugin>   chainedPlugins;
     private final boolean                     dedupStrings;
     private       PolicyRefresher             refresher;
-    private       RangerPolicyEngine          policyEngine;
+    private volatile RangerPolicyEngine       policyEngine;
     private       RangerAuthContext           currentAuthContext;
```
mneethiraj: @vyommani  - can you please review if following fields also need to be marked as `volatile`?
- currentAuthContext
- roles
- serviceConfigs

These fields are updated whenever the plugin receives new policies or roles or tags; and are referenced from multiple threads.

vyommani: @mneethiraj I've analyzed the code and agree with your points:
1-> currentAuthContext should be volatile since it's updated in setPolicies() and read in getTagEnricher().
2-> roles is updated in setRoles(), a public method, so making it volatile is a good idea for defensive programming, even if it's mostly initialized or updated via admin operations.
3-> serviceConfigs is updated in setPolicies() and read in other threads, so volatile is necessary for visibility guarantees.
Given the negligible cost of volatile, I agree it's a good practice to make all three fields volatile for API safety and to avoid potential issues.

Your thoughts?

mneethiraj: @vyommani  - yes, please go ahead and mark these fields also as `volatile`.

mneethiraj (approved):

ramackri (approved):

PR #759, opened 2025-12-03
RANGER-5362:Restrict security zone visibility to its owner.

bhaveshamre:

This PR implements enhanced Security Zone visibility and access restrictions in Apache Ranger. The main goal is to ensure that a security zone is visible only to users who are authorized to administer or audit that zone.

Key changes introduced:

Restricted Zone Visibility

A security zone is now visible only to:

Zone Admin users / groups / roles

Zone Auditor users / groups / roles

Super Admin users (global ADMIN)

All other users cannot see non-public zones.

Updated Access Logic

Zone Admins: Can view the zone and fully manage it (create/edit/delete policies).

Zone Auditors: Can view the zone and access its audit logs only. Cannot modify zone/policies.

Unauthorized users: Cannot view the zone, its policies, or its audits.

Public zones continue to be visible to all.

Backend Enforcement

Visibility and filtering logic added/updated in:

PublicAPIsv2 (all zone-fetching API endpoints)

SecurityZoneREST

SecurityZoneDBStore (core business logic)

DAO layer (XXSecurityZoneDao, XXUserDao, RangerDaoManagerBase)

JPA named queries (jpa_named_queries.xml)

UI Data Filtering

Updated zone filtering for:

Security Zone homepage (getAllZones)

Main homepage header zone info (getSecurityZoneHeaderInfoList)

Service homepage (getSecurityZoneHeaderInfoListByServiceId)

## How was this patch tested?

I have verified the changes locally by running mvn clean compile package install and then bringing up the Ranger environment to ensure that the updated security-zone visibility logic works correctly during development testing.


pradeepagrawal8184 (commented):
Have you tested this patch with all db flavors in fresh and upgrade case ?

PR #779, opened 2025-12-31
RANGER-5397: IllegalArgumentException in RangerRESTClient when no ranger.plugin.<service>.policy.rest.url is configured

vyommani:

RangerBasePlugin initialization fails with an IllegalArgumentException when the ranger.plugin..policy.rest.url property is not set.
The issue occurs because getConfiguredURLs() returns an empty list, causing Random.nextInt(0) to throw an exception.

## How was this patch tested?

wrote a new test TestRangerRESTClient.testPluginInit_WithNoUrl_DoesNotCrash() this will fail without fix and pass with fix.


vyommani:
create a new pr from the old approved one(https://github.com/apache/ranger/pull/732). In this PR i migrated the test to use junit 5.

pradeepagrawal8184 (approved):

PR #824, opened 2026-01-27
RANGER-5463: Update userstore version after group-user mapping changes in createOrDeleteXGroupUserList

kimsehwan96:

Added missing updateUserStoreVersion() call in XUserMgr.createOrDeleteXGroupUserList().                                   
                                                                                                                            
Without this fix, when user-group membership changes in LDAP (e.g., a user is added to or removed from a group), the      
UserSync successfully updates the x_group_users table in Ranger Admin, but the RangerUserStore version remains unchanged. 

Other similar methods like createXGroupUserFromMap() and createOrUpdateXGroups() already call updateUserStoreVersion() after modifying user/group data.    

## How was this patch tested?
- Built locally and deployed to a test environment
- Configured UserSync with LDAP source / Added user to a group in LDAP / Verified userstore version increments after sync

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
@@ -2959,10 +2959,14 @@ public class XUserMgr extends XUserMgrBase {
                 }
 
                 for (GroupUserInfo groupUserInfo : groupUserInfoList) {
-                    xGroupUserService.createOrDeleteXGroupUsers(groupUserInfo, usersFromDB);
+                    if (xGroupUserService.createOrDeleteXGroupUsers(groupUserInfo, usersFromDB)) {
+                        updatedGroups++;
+                    }
+                    
                 }
-
-                updatedGroups = groupUserInfoList.size();
+                if (logger.isDebugEnabled()) {
+                    logger.debug("No. of groups actually updated = {}", updatedGroups);
+                }
             }
 			if (updatedGroups > 0) {
 				updateUserStoreVersion("createOrDeleteXGroupUserList(updatedGroups=" + updatedGroups + ")");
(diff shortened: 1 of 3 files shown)


security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
```diff
@@ -2964,6 +2964,9 @@ public int createOrDeleteXGroupUserList(List<GroupUserInfo> groupUserInfoList) {
 
                 updatedGroups = groupUserInfoList.size();
```
mneethiraj: Looking at implementation of `xGroupUserService.createOrDeleteXGroupUsers()`, no update might be performed in some scenarios - depending on contents of parameter `groupUserInfo`. I suggest updating this method to return `true` when  it performs any update and use that to increment `updatedGroups`.

kimsehwan96: @mneethiraj Changed `xGroupUserService.createOrDeleteXGroupUsers()` method to return boolean. If the method early returned with no actual updates then return false and If actual update was performed then return true. 

And increment `updatedGroups` and return its count as `createOrDeleteGroupUserList` return and if at lease one update was really performed then update user store version.

Alos I added test code about it.

kimsehwan96: By the way, returning boolean with name `createOrDeleteXGroupUsers` might be little awkward

kimsehwan96: @mneethiraj Can you review this? Thanks.

mneethiraj (approved):

mneethiraj:
Thank you @kimsehwan96 for reporting this issue and the fix.

Merged in master & ranger-2.9 branches.

horvatic:
Just for anyone using 2.7 or 2.8 trying to find a work around 

We made a DB trigger on `public.x_ugsync_audit_info` to update `x_ranger_global_state` 

We add one to the `RangerUserStore` which can be found under `state_name`

PR #827, opened 2026-01-28
RANGER-5466 Avoid hardcoded delta sync and unnecessary LDAP filter when delta sync is disabled

kimsehwan96:

1. setup.py: Changed hardcoded `ranger.usersync.ldap.deltasync=true` to only set the default value when the property is not already configured in install.properties.

2. LdapUserGroupBuilder: Skip appending delta sync filter conditions (uSNChanged, modifyTimestamp) to LDAP search queries when delta sync is disabled. Applied to both getUsers() and getGroups()

Some LDAP services (e.g., Google Workspace Secure LDAP) do not support the attributes used for delta sync (uSNChanged, modifyTimestamp), which causes delta sync to not work properly. In such cases, users should be able to disable delta sync and run full sync without unnecessary filter conditions being included in the LDAP queries.

## How was this patch tested?

- Built locally and verified LDAP search filters with delta sync enabled and disabled

### DIFF
--- a/ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
+++ b/ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
@@ -512,7 +512,11 @@ public class LdapUserGroupBuilder implements UserGroupSource {
                 deltaSyncUserTimeStamp = dateFormat.format(new Date(0));
             }
 
-            extendedUserSearchFilter = "(objectclass=" + userObjectClass + ")(|(uSNChanged>=" + deltaSyncUserTime + ")(modifyTimestamp>=" + deltaSyncUserTimeStamp + "Z))";
+            if (config.isDeltaSyncEnabled()) {
+                extendedUserSearchFilter = "(objectclass=" + userObjectClass + ")(|(uSNChanged>=" + deltaSyncUserTime + ")(modifyTimestamp>=" + deltaSyncUserTimeStamp + "Z))";
+            } else {
+                extendedUserSearchFilter = "(objectclass=" + userObjectClass + ")";
+            }
 
             if (userSearchFilter != null && !userSearchFilter.trim().isEmpty()) {
                 String customFilter = userSearchFilter.trim();
@@ -753,7 +757,11 @@ public class LdapUserGroupBuilder implements UserGroupSource {
                 deltaSyncGroupTimeStamp = dateFormat.format(new Date(0));
             }
 
-            extendedAllGroupsSearchFilter = "(&" + extendedGroupSearchFilter + "(|(uSNChanged>=" + deltaSyncGroupTime + ")(modifyTimestamp>=" + deltaSyncGroupTimeStamp + "Z)))";
+            if (config.isDeltaSyncEnabled()) {
+                extendedAllGroupsSearchFilter = "(&"  + extendedGroupSearchFilter + "(|(uSNChanged>=" + deltaSyncGroupTime + ")(modifyTimestamp>=" + deltaSyncGroupTimeStamp + "Z)))";
+            } else {
+                extendedAllGroupsSearchFilter = "(&"  + extendedGroupSearchFilter + ")";
+            }
 
             LOG.info("extendedAllGroupsSearchFilter = {}", extendedAllGroupsSearchFilter);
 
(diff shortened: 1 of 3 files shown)


mneethiraj (approved):

ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
spolavarpau1, line 511: Have you checked this? We already are checking if deltaSyncEnabled() and setting the USNChanged to 0 to force full sync. I don't think we need an extra check again.

spolavarpau1 (changes requested):

ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
kimsehwan96: @spolavarpau1 
Thanks for the review. That reset works only on servers that actually have those attributes (AD/OpenLDAP).

Google Secure LDAP (schema) provides neither uSNChanged nor modifyTimestamp, and a filter on an attribute the server doesn't have matches nothing, regardless of the value. 

So with delta sync disabled the search returns 0 entries and no users get synced at all. 

The issue is the presence of the clause, not its value. Verified against Google Secure LDAP: with the clause, 0 entries; without it, all entries. On AD/OpenLDAP the result set is unchanged, so no behavior change.

pradeepagrawal8184:
Seems there are tests failures:
[ERROR] Failures: 
[ERROR]   TestLdapUserGroupBuilder.testW_getGroups_processes_groups_and_memberships:647 (&(objectclass=groupOfNames)(cn=*)): must contain 'modifyTimestamp>=' ==> expected: <true> but was: <false>
[INFO] 
[ERROR] Tests run: 99, Failures: 1, Errors: 0, Skipped: 2

spolavarpau1 (approved):

ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
spolavarpau1: One minor change, can you please update this if condition to if (config.isDeltaSyncEnabled() && (computeDeletes)) ? That way it is much easier to understand the code. In case of delta sync not enabled, anyway we are performing full sync all the time and not using deltaSyncGroupTime or deltaSyncGroupTimeStamp in the filter.

spolavarpau1: One minor change, can you please update this if condition to 
if (!groupUserTable.rowKeySet().isEmpty() || (config.isDeltaSyncEnabled() && (computeDeletes))) 
? That way it is much easier to understand the code. In case of delta sync not enabled, anyway we are performing full sync all the time and not using deltaSyncGroupTime or deltaSyncGroupTimeStamp in the filter.

spolavarpau1 (changes requested):

kimsehwan96:
> Seems there are tests failures: [ERROR] Failures: [ERROR] TestLdapUserGroupBuilder.testW_getGroups_processes_groups_and_memberships:647 (&(objectclass=groupOfNames)(cn=*)): must contain 'modifyTimestamp>=' ==> expected: but was: [INFO] [ERROR] Tests run: 99, Failures: 1, Errors: 0, Skipped: 2

@pradeepagrawal8184 
That test was relying on the delta filter being appended even with delta sync off. I fixed it in https://github.com/apache/ranger/pull/827/commits/74e64097fa7b8599e725072dad9fdec61f803a57
(also added more tests related with this PR changes)

ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
kimsehwan96: @spolavarpau1 Done: https://github.com/apache/ranger/pull/827/commits/19485aa2cbe4498ff2bfbaafee3636a742276812

kimsehwan96: @spolavarpau1 Done: https://github.com/apache/ranger/pull/827/commits/19485aa2cbe4498ff2bfbaafee3636a742276812

spolavarpau1 (approved):

PR #894, opened 2026-03-26
Ranger 2470: Support for native SAML (Okta , keycloak etc) for Ranger Admin Consol

vyommani:
Summary

This PR adds native SAML2 Service Provider (SP) authentication to Ranger Admin using Spring Security SAML2 and OpenSAML 4.x. With this change, Ranger Admin can act as a SAML SP directly, integrating withany standards-compliant Identity Provider (Keycloak, Okta, ADFS, Azure AD).

What Changed

New classes:

1-> RangerSamlRegistrationFactory — builds the RelyingPartyRegistration from configured SP key/cert and IdP metadata, with REDIRECT-binding SLO support
2-> RangerDelegatingAuthenticationEntryPoint — routes unauthenticated requests to the SAML login endpoint or falls back to the existing entry point depending on ranger.authentication.method
3-> RangerDelegatingLogoutSuccessHandler — routes post-logout to the SAML SLO chain or the default handler

Modified classes:

1-> RangerAuthenticationProvider — handles Saml2Authentication tokens; extracts username and group memberships from configurable SAML attributes; auto-provisions users with a default role
2-> SessionMgr — auto-creates SAML users in the Ranger DB on first login (mirrors existing SSO/Spnego behavior)
3-> RangerSecurityContextFormationFilter — returns AUTH_TYPE_SAML when SAML is the active auth method
4-> UserREST — passes authenticationMethod to the frontend so the UI can handle SAML-specific logout flow

Configuration:

security-applicationContext.xml — wires the SAML filter chain (saml2WebSsoAuthenticationRequestFilter, saml2AuthenticationFilter, samlLogoutRequestFilter, samlLogoutResponseFilter); permits SAML endpoint URLs without authentication
ranger-admin-site.xml — introduces SAML-specific properties (see Configuration section below)

Frontend:

login.jsp — auto-redirects to the IdP when SAML is enabled; guards against redirect loops on post-logout landing

XAUtils.js / SideBarBody.jsx — uses full browser navigation (instead of AJAX) for logout, enabling the SLO redirect chain to complete end-to-end

Configuration

When ranger.authentication.method=SAML, the following properties apply in ranger-admin-site.xml:

ranger.saml.entity.id --> SP Entity ID registered with the IdP
ranger.saml.idp.metadata.location --> IdP metadata file path or URL
ranger.saml.sp.key --> SP private key (PKCS8 PEM) path
ranger.saml.sp.cert --> SP public certificate (X.509 PEM) path
ranger.saml.attribute.username	--> SAML attribute to use as the Ranger username	NameID
ranger.saml.attribute.role --> SAML attribute to use for group membership	groups
ranger.saml.default.role --> Default role assigned to new SAML users	ROLE_USER
ranger.saml.admin.group	SAML group that maps to ROLE_SYS_ADMIN	ranger-admins

Design Decisions
Delegating handlers over replacing them — RangerDelegatingAuthenticationEntryPoint and RangerDelegatingLogoutSuccessHandler wrap existing handlers rather than replacing them, so non-SAML auth methods (LDAP, Kerberos, Unix) are unaffected.

Auto-provisioning on first login — consistent with how Ranger handles SSO and Spnego users; no admin action is needed to pre-create SAML users.
No proxy dependency — the SP logic runs entirely within the Ranger Admin JVM.

## How was this patch tested?

 SAML login flow tested against [Keycloak]
 Single Logout (SLO) tested — browser redirected to IdP and back
 Non-SAML auth methods (LDAP/Unix) verified unaffected with ranger.authentication.method set to their respective values
 Auto-provisioned user created in Ranger DB on first SAML login


security-admin/src/main/java/org/apache/ranger/rest/UserREST.java
```diff
@@ -274,6 +274,7 @@ public VXPortalUser getUserProfile(@Context HttpServletRequest request) {
             long                inactivityTimeout = PropertiesUtil.getLongProperty("ranger.service.inactivity.timeout", 15 * 60);
 
             configProperties.put("inactivityTimeout", Long.toString(inactivityTimeout));
+            configProperties.put("authenticationMethod", PropertiesUtil.getProperty("ranger.authentication.method", "NONE"));
```
spolavarpau1, line 277: I am not completely sure why we had to add this?

security-admin/src/main/java/org/apache/ranger/security/web/filter/RangerSecurityContextFormationFilter.java
```diff
@@ -174,7 +174,7 @@ private int getAuthType(Authentication auth, HttpServletRequest request) {
 
         Object  ssoEnabledObj = request.getAttribute("ssoEnabled");
         boolean ssoEnabled    = ssoEnabledObj != null ? Boolean.parseBoolean(String.valueOf(ssoEnabledObj)) : PropertiesUtil.getBooleanProperty("ranger.sso.enabled", false);
-
+        String authMethod = PropertiesUtil.getProperty("ranger.authentication.method", "NONE");
```
spolavarpau1, line 177: Why are we explicitly setting NONE?

security-admin/src/main/resources/conf.dist/ranger-admin-site.xml
```diff
@@ -453,4 +453,46 @@
 		<name>xasecure.audit.jaas.Client.option.principal</name>
 		<value></value>
 	</property>
+	<property>
```
spolavarpau1, line 456: Do we need to specify these properties in our distribution of config file as we already have default values in our code?

security-admin/src/main/resources/conf.dist/ranger-admin-site.xml
```diff
@@ -453,4 +453,46 @@
...
+	</property>
+	<property>
+		<name>ranger.saml.default.role</name>
+		<value>ROLE_USER</value>
+	</property>
+	<property>
+		<name>ranger.saml.admin.group</name>
+		<value>ranger-admins</value>
+	</property>
+	<!-- SAML SP / IdP file paths and identity -->
```
spolavarpau1, line 472: Can you check if we can move the SAML related properties to ranger-admin-default-site.xml

security-admin/src/main/webapp/react-webapp/src/views/SideBar/SideBarBody.jsx
```diff
@@ -220,21 +220,34 @@ export const SideBarBody = (props) => {
...
+
       await fetchApi({
         url: "logout",
         baseURL: "",
         headers: {
           "cache-control": "no-cache"
         }
       });
-      if (checkKnoxSSOVal !== undefined || checkKnoxSSOVal !== null) {
+      if (checkKnoxSSOVal !== undefined && checkKnoxSSOVal !== null) {
```
spolavarpau1, line 250: Can you check if this has any regression impact when KnoxSSO is enabled?

PR #903, opened 2026-04-01
RANGER-5534: added validation logic for username and groupname for usersync process

sneethiraj:

* Added username and groupname validation based on regEx
* Configuration - userName validation enabled [true/false]
* Configuration - groupName validation enabled [true/false] 
* Configuration - RegEx for the username validation [default: "^[a-z_][a-z0-9_-]{0,31}$" ]
* Configuration - RegEx for the groupname validation [default:  "^[a-z][a-z0-9-]{0,30}$" ] 

## How was this patch tested?

Tested locally on Mac


ugsync/src/main/java/org/apache/ranger/unixusersync/process/UnixUserGroupBuilder.java
```diff
@@ -330,6 +340,13 @@ private void buildUnixUserList(String command) throws Throwable {
                     continue;
                 }
 
+                if (validateUserName) {
+                    if (!isValidUserName(userName)) {
+                        LOG.warn("Ignoring Unix Username: [{}]: failed to confirm to validation-pattern: [{}]", userName, regExUserNameValidator);
```
mneethiraj, line 344: @sneethiraj  - this could result in usernames having valid non-English characters (accent characters, CJK character sets) to be skipped with default configuration. This can break environments already have such user names and group names.

sneethiraj: **Good catch**.  We can implement the following regex validations that supports international char sets: 

```
    private static final String DEFAULT_REGEX_USERNAME_VALIDATOR = "^[\\p{L}\\p{Mn}\\p{Nd}\\._-]{3,64}$" ;
    private static final String DEFAULT_REGEX_GROUPNAME_VALIDATOR = "^[\\p{L}\\p{N}._\\-\\s]{1,256}$"    ;

```
However, we need to document this as part of the upgrade to 2.8.1+ release that the username and groupname validations are defaulted to be ON with specific regex and customers are requested to review and modify if needed before deploying newer version of code. 

Can you share your thoughts on the best approach to resolve this?

vyommani: Hi,

I had already addressed this vulnerability in an earlier change by updating the command execution from: {"bash", "-c", "id -G " + userName} 
to the safer:
{"id", "-G", userName}
This change ensures that the shell is never invoked. As a result, metacharacters in the username have no special meaning at the OS execve() level, which fully mitigates the command injection risk.

The validation logic proposed in this PR is a good attempt to add defense-in-depth. However, I'm concerned that the regex-based approach (even with the updated broader patterns) could unintentionally reject many valid usernames — for example, legitimate usernames written in Korean, Japanese, Chinese, or other languages that use non-ASCII characters.

Could you please take a look at the fix in RANGER-5508 and let me know what you think about the above analysis? If this turns out to be a completely unrelated issue, please feel free to let me know as well.

vyommani: I already fixed this issue as part of RANGER-5508   please check this fix and do let me know if it is completely different issue.

sneethiraj:
Closing this PR to resolve this via different approach.

PR #910, opened 2026-04-09
RANGER-5543:Ranger KMS GCP always try to create master key irrespective of its existence

bhaveshamre:

Added missing dependencies (guava, failureaccess, protobuf-java)
Fixed unnecessary master key creation by introducing masterKeyExists()
Implemented role-based logic:
Owner → creates key only if not exists
Viewer → only checks existence, no creation

## How was this patch tested?

Built successfully using mvn clean compile package install. Manual testing was performed across multiple scenarios: when using the Owner role with a new key, the key was created successfully; with an existing key, no creation was attempted. For the Viewer role, the system correctly performed only existence checks for an existing key without attempting creation, and for a new key, no creation was attempted, resulting in a permission denied error as expected.

### DIFF
--- a/kms/pom.xml
+++ b/kms/pom.xml
@@ -104,6 +104,16 @@
             <version>${jsr305.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>failureaccess</artifactId>
+            <version>${google.failureaccess.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>${google.guava.version}</version>
+        </dependency>
+        <dependency>
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
             <version>${gcp.protobuf-java.version}</version>
--- a/kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
+++ b/kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
@@ -68,9 +68,8 @@ public class RangerGoogleCloudHSMProvider implements RangerKMSMKI {
 
     @Override
     public boolean generateMasterKey(String unusedPassword) throws Throwable {
-        if (this.masterKeyExists()) {
-            return false;
-        } else {
+        boolean isMKGenerated = false;
+        if (!this.masterKeyExists()) {
             //The ENCRYPT_DECRYPT key purpose enables symmetric encryption.
             //All keys with key purpose ENCRYPT_DECRYPT use the GOOGLE_SYMMETRIC_ENCRYPTION algorithm.
             //No parameters are used with this algorithm.
@@ -89,15 +88,14 @@ public class RangerGoogleCloudHSMProvider implements RangerKMSMKI {
                 throw new RuntimeCryptoException("Failed to create master key with name '" + this.gcpMasterKeyName + "', Error - " + e.getMessage());
             }
 
-            if (createdKey == null) {
+            if (createdKey != null) {
+                logger.info("Master Key Created Successfully On Google Cloud HSM : {}", this.gcpMasterKeyName);
+                isMKGenerated = true;
+            } else {
                 logger.info("Failed to create master key : {}", this.gcpMasterKeyName);
-                return false;
             }
-
-            logger.info("Master Key Created Successfully On Google Cloud HSM : {}", this.gcpMasterKeyName);
-
-            return true;
         }
+        return isMKGenerated;
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
```diff
@@ -68,37 +68,36 @@ public RangerGoogleCloudHSMProvider(Configuration conf) throws Exception {
...
-        // Create the key.
-        CryptoKey createdKey = null;
-        try {
-            createdKey = client.createCryptoKey(this.keyRingName, this.gcpMasterKeyName, key);
-        } catch (Exception e) {
-            if (e instanceof AlreadyExistsException) {
-                logger.info("MasterKey with the name '{}' already exist.", this.gcpMasterKeyName);
-                return true;
-            } else {
+        if (this.masterKeyExists()) {
```
vikaskr22: @bhaveshamre ,  One minor comment, Can we try to have one single "return" statement, like declare one local boolean var say 
`boolean isMKGenerated = false`
Then set this to true when MK actually gets created and at last return this var.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
```diff
@@ -221,4 +220,24 @@ private static void updateEnv(String name, String val) throws ReflectiveOperatio
...
+        if (this.client == null) {
+            throw new RuntimeCryptoException("Google Cloud KMS client is not initialized; call onInitialization() first.");
+        }
+
+        CryptoKeyName keyName = CryptoKeyName.of(this.gcpProjectId, this.gcpLocationId, this.gcpKeyRingId, this.gcpMasterKeyName);
+
+        try {
+            CryptoKey cryptoKey = this.client.getCryptoKey(keyName);
+            logger.info("Ranger masterKey present with name: {}", cryptoKey.getName());
+            return true;
```
vikaskr22: same review comment as above. Pls try to have one single return statement.

kms/pom.xml
```diff
@@ -103,6 +103,16 @@
             <artifactId>jsr305</artifactId>
             <version>${jsr305.version}</version>
         </dependency>
+        <dependency>
```
vikaskr22, line 106: @bhaveshamre , I want to understand why this new lib is required. As per my knowledge, `com.google.giava:failureaccess` is part of Guava lib. Are we using this anywhere directly in the KMS code ?

bhaveshamre: @vikaskr22 We do not use failureaccess directly in our codebase, however, it is a required dependency used internally by Google Guava. It has been added explicitly to ensure it is available in the packaged runtime.

This was necessary because, after integrating Google Cloud KMS, the runtime environment (Ranger/Hadoop) did not reliably resolve all required dependencies, leading to a NoClassDefFoundError for com/google/common/util/concurrent/internal/InternalFutureFailureAccess.

Including this dependency explicitly ensures all required classes are present at runtime and prevents such classpath-related failures.

vikaskr22: Don't we include transitive or any runtime dependencies in distro module's pom file ?

kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
bhaveshamre: @vikaskr22 ,addressed the comments and made the necessary changes. Please review.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
bhaveshamre: @vikaskr22 addressed the comments and made the necessary changes. Please review.

kms/src/main/java/org/apache/hadoop/crypto/key/RangerGoogleCloudHSMProvider.java
```diff
@@ -68,37 +68,34 @@ public RangerGoogleCloudHSMProvider(Configuration conf) throws Exception {
...
                 throw new RuntimeCryptoException("Failed to create master key with name '" + this.gcpMasterKeyName + "', Error - " + e.getMessage());
             }
-        }
 
-        if (createdKey == null) {
-            logger.info("Failed to create master key : {}", this.gcpMasterKeyName);
-            return false;
+            if (createdKey == null) {
+                logger.info("Failed to create master key : {}", this.gcpMasterKeyName);
+                return false;
```
vikaskr22: This `return false` is not required , instead set it to true if (createdKey != null) and in else part simply log the message.

bhaveshamre: @vikaskr22 , addressed the comments and made the necessary changes. Kindly review.

vikaskr22 (approved):
LGTM now. Thanks @bhaveshamre .

dhavalshah9131 (approved):

PR #911, opened 2026-04-09
RANGER-5549: replace use of com.kstruct.gethostname4j library

mneethiraj:

Replaced use of `com.kstruct.gethostname4j.Hostname.getHostname()` with env variable lookup `HOSTNAME` or `COMPUTERNAME`), with fallback to `InetAddress.getLocalHost().getHostName()`.

## How was this patch tested?

Using docker setup, verified that plugins report correct hostname in audit logs.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/util/RangerRESTUtils.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/util/RangerRESTUtils.java
@@ -145,7 +145,7 @@ public class RangerRESTUtils {
     private static String getHostname() {
         String hostname = System.getenv("HOSTNAME");
 
-        if (StringUtils.isEmpty(hostname)) {
+        if (StringUtils.isBlank(hostname)) {
             hostname = System.getenv("COMPUTERNAME");
 
             if (StringUtils.isBlank(hostname)) {
--- a/agents-common/src/main/java/org/apache/ranger/plugin/util/RangerRESTUtils.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/util/RangerRESTUtils.java
@@ -162,12 +162,11 @@ public class RangerRESTUtils {
                 try {
                     hostname = InetAddress.getLocalHost().getHostName();
                 } catch (Exception e) {
-                    LOG.error("ERROR: Unable to find hostname for the agent ", e);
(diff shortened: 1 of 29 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/util/RangerRESTUtils.java
```diff
@@ -73,7 +72,7 @@ public class RangerRESTUtils {
     private static final Logger LOG                                               = LoggerFactory.getLogger(RangerRESTUtils.class);
     private static final int    MAX_PLUGIN_ID_LEN                                 = 255;
 
-    public static        String hostname;
+    public static final String hostname = getHostname();
```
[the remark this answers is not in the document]
mneethiraj: This variable has been public for many releases; making it private can break any references. Will consider this in next major release - 3.0.0.

agents-common/src/main/java/org/apache/ranger/plugin/util/RangerRESTUtils.java
```diff
@@ -143,12 +142,32 @@ public String getAppIdFromPluginId(String pluginId) {
...
+            } catch (SecurityException excp) {
+                LOG.error("Error in getting environment COMPUTERNAME", excp);
+            }
+
+            if (StringUtils.isBlank(hostname)) {
+                try {
+                    hostname = InetAddress.getLocalHost().getHostName();
+                } catch (Exception e) {
+                    LOG.error("ERROR: Unable to find hostname for the agent ", e);
+                    hostname = "unknownHost";
```
vyommani: Please move the "unknownHost" fallback outside the try-catch and add this check at the end:

// Final safety net
if (StringUtils.isBlank(hostname)) {
    hostname = "unknownHost";
}

Even though InetAddress.getLocalHost().getHostName() never returns null, it can return an empty string in some edge cases (Docker, restrictive SecurityManager, network issues, etc.). This ensures we never return a blank hostname.

mneethiraj: Good suggestion. Thanks @vyommani.

rameeshm (approved):

vyommani (approved):

PR #915, opened 2026-04-12
RANGER-5539: Add Authorisation Check for doAsUser Parameter

ChinmayHegde24:
Currently RangerJwtAuthHandler accepts the doAsUser value directly from the incoming request and uses it to establish the authenticated user identity without performing any validation.
So the user should be validated for impersonation permission on doAsUser parameter.


dineshkumar-yadav (approved):
LGTM

PR #922, opened 2026-04-17
RANGER-5559:  fixes to docker container startup as non-root user and/or hostname as non "ranger"

sneethiraj:
As startup script is trying to move files into /etc/init.d and /usr/bin, it is throwing lot of errors.
Also, having hardcoded hostname as ranger during service creation causes more errors during startup.

Those steps that requires root permission WILL run only if you are a root. This will be required for non-docker setp.
Also, ranger service creation script uses the current hostname to identify the service instead of hardcoded 'ranger'

## How was this patch tested?
Build and deployed as a docker container in Linux Microk8s environment. It deploys fine with no errors.

### DIFF
--- a/dev-support/ranger-docker/scripts/admin/create-ranger-services.py
+++ b/dev-support/ranger-docker/scripts/admin/create-ranger-services.py
@@ -1,8 +1,14 @@
 from apache_ranger.model.ranger_service import RangerService
 from apache_ranger.client.ranger_client import RangerClient
 from json import JSONDecodeError
+import socket
 
-ranger_client = RangerClient('http://ranger:6080', ('admin', 'rangerR0cks!'))
+current_host_name = socket.gethostname()
+
+if not current_host_name:
+	current_host_name = "ranger"
+
+ranger_client = RangerClient('http://' + current_host_name  + ':6080', ('admin', 'rangerR0cks!'))
 
 
 def service_not_exists(service):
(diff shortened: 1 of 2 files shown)


dev-support/ranger-docker/scripts/admin/create-ranger-services.py
```diff
@@ -1,8 +1,22 @@
...
+if skipMe:
+    print("INFO: Skipping sample ranger services ...")
+    sys.exit(0)
+
+current_host_name = socket.gethostname()
+
+if not current_host_name:
+    current_host_name = "ranger"
+
+ranger_client = RangerClient('http://' + current_host_name  + ':6080', ('admin', 'rangerR0cks!'))
```
mneethiraj, line 19: Given this script is executed in the container running ranger admin service, it seem call to  `socket.gethostname()` is not necessary;  simply using `localhost` would work, right?

dev-support/ranger-docker/scripts/admin/create-ranger-services.py
```diff
@@ -1,8 +1,22 @@
 from apache_ranger.model.ranger_service import RangerService
 from apache_ranger.client.ranger_client import RangerClient
 from json import JSONDecodeError
+import socket
+import os
+import sys
 
-ranger_client = RangerClient('http://ranger:6080', ('admin', 'rangerR0cks!'))
+skipMe = os.getenv("SKIP_CREATE_SAMPLE_SERVICES","false").lower() == "true"
```
mneethiraj, line 8: @sneethiraj  - for developer testing, I think it is okay to create services during Ranger admin startup. However, this change would be useful in the official docker image released to docker hub, which is built from https://github.com/apache/ranger-tools/tree/main/release.

PR #936, opened 2026-05-08
RANGER-5577:API to support bulk delete of resources in DataShare

rameeshm:
 To have the API to support bulk delete by getting the resources Id as query param.

## How was this patch tested?
Test in local via with curl call.
` curl -ikv --negotiate -u : -H "Accept: application/json" -H "Content-Type: application/json" -X GET 'https://<ranger-admin-host>:6182/service/gds/resources?Id=2'

curl -ikv --negotiate -u : -H "Accept: application/json" -H "Content-Type: application/json" -X DELETE 'https://<ranger-admin-host>:6182:6182/service/gds/resources?id=9&id=10'
`

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
@@ -1292,12 +1292,16 @@ public class GdsREST {
     @DELETE
     @Path("/resources")
     @PreAuthorize("@rangerPreAuthSecurityHandler.isAPIAccessible(\"" + RangerAPIList.REMOVE_SHARED_RESOURCES + "\")")
-    public void removeSharedResources(List<Long> resourceIds) {
-        LOG.debug("==> GdsREST.removeSharedResources({})", resourceIds);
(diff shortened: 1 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/rest/GdsREST.java
```diff
@@ -1292,12 +1292,16 @@ public void removeSharedResource(@PathParam("id") Long resourceId) {
     @DELETE
     @Path("/resources")
     @PreAuthorize("@rangerPreAuthSecurityHandler.isAPIAccessible(\"" + RangerAPIList.REMOVE_SHARED_RESOURCES + "\")")
-    public void removeSharedResources(List<Long> resourceIds) {
-        LOG.debug("==> GdsREST.removeSharedResources({})", resourceIds);
+    public void removeSharedResources(@QueryParam("resourceIds") List<Long> resourceIds) {
```
mneethiraj: @rameeshm  - given a separate query parameter will be needed for each resource, consider naming the query parameter as `resourceId` or `id`. I suggest `id` as its shorter, hence can handle larger number of resources.

mneethiraj (approved):

PR #940, opened 2026-05-11
RANGER-5586:Ranger Admin UI Fails to Load Access Page Due to Solr Integration Error in Docker Deployment

bhaveshamre:
Updated the audit_solr_urls configuration in Ranger Docker admin property files to use the Docker network hostname ranger-solr.rangernw instead of ranger-solr. This fixes Solr connectivity issues causing the Ranger Admin Access Logs page to fail in Docker deployments.

## How was this patch tested?
Tested manually by deploying Ranger using Docker, accessing the Ranger Admin UI, and verifying that the Access Logs page loaded successfully without Solr errors.


dsatam99 (approved):
Changes looks good to me !

PR #943, opened 2026-05-12
Ranger -5587 modified && condition to OR  to resolve the issue of definition id 

abhiishek26:
synchronization in put /definitions/{id}

 The  PUT api    http://localhost:6080/service/plugins/definitions/{id}   is not throwing error when different id is passed from url and json payload , the issue is because in of a && condition updateServiceDef in ServiceREST which should have been OR ,I have  changed it  . On changing the And condition to or it is proprly throwing error as expected .

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
@@ -359,7 +359,7 @@ public class ServiceREST {
         // if serviceDef.id is null, then set param 'id' into serviceDef Object
         if (serviceDef.getId() == null) {
             serviceDef.setId(id);
-        } else if (StringUtils.isBlank(serviceDef.getName()) && !serviceDef.getId().equals(id)) {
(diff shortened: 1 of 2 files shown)


vyommani:
can you add a  Junit test if possible ?

mneethiraj (approved):

abhiishek26:
> can you add a Junit test if possible ?

Hi Vyom I have added the unit tests . The existing unit tests covered most of the scenarios but 
1>Earlier the case when  json payload and request url ids were not matching but serviceDef name was not blank was    getting passed but it should have failed  . Unit test testUpdateServiceDefWithValidNameButMismatchedId.  handles this case for the changes made .
2> Unit test test2updateServiceDef. was failing with null pointer exception for the changes made  , I have modified it to handle the exception  correctly  .

vyommani (approved):
LGTM

ramackri:
LGTM

ramackri (approved):

PR #945, opened 2026-05-12
RANGER-5592: [GDS] Validity Period is not working for dataset policy

maheshbandal15:
Fixed an issue where the GDS validity period was not enforced for dataset policies. Previously, when a dataset was configured with a validity end date set before the current date, users were still allowed to execute queries (e.g., SELECT) instead of being denied access. The validity period expiration is now correctly evaluated, ensuring that access through dataset policies is denied once the validity period has expired.

## How was this patch tested?
Manual testing is done.
build successful with "mvn clean install"

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/gds/GdsDatasetEvaluator.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/gds/GdsDatasetEvaluator.java
@@ -124,10 +124,9 @@ public class GdsDatasetEvaluator {
                     RangerAccessRequestUtil.setAccessTypeACLResults(datasetRequest.getContext(), null);
 
                     policyEvaluators.forEach(e -> {
-                        if (!e.isApplicable(accessTime)) {
-                            return;
+                        if (e.isApplicable(accessTime)) {
+                            e.evaluate(datasetRequest, datasetResult);
                         }
-                        e.evaluate(datasetRequest, datasetResult);
                     });
                 } finally {
                     RangerAccessRequestUtil.setAccessTypeResults(datasetRequest.getContext(), null);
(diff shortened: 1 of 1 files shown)


agents-common/src/main/java/org/apache/ranger/plugin/policyengine/gds/GdsDatasetEvaluator.java
```diff
@@ -121,7 +123,12 @@ public void evaluate(RangerAccessRequest request, GdsAccessResult result, Collec
                     RangerAccessRequestUtil.setAccessTypeResults(datasetRequest.getContext(), null);
                     RangerAccessRequestUtil.setAccessTypeACLResults(datasetRequest.getContext(), null);
 
-                    policyEvaluators.forEach(e -> e.evaluate(datasetRequest, datasetResult));
+                    policyEvaluators.forEach(e -> {
+                        if (!e.isApplicable(accessTime)) {
```
mneethiraj: For better readability, consider reversing the conition:

```
policyEvaluators.forEach(e -> {
    if (e.isApplicable(accessTime)) {
        e.evaluate(datasetRequest, datasetResult);
    }
});
```

Also, this fix will be needed in `GdsProjectEvaluator.evaluate()` as well.

maheshbandal15: Thanks for reviewing. I have addressed the review comment. Kindly review the updated patch.
 
[RANGER-5597](https://issues.apache.org/jira/browse/RANGER-5597) is created to track required fix in `GdsProjectEvaluator.evaluate()`. Refer https://github.com/apache/ranger/pull/946

mneethiraj (approved):

PR #948, opened 2026-05-15
RANGER-5602: Local build failure due to unit test failure in TestPolicyACLs

ChinmayHegde24:
local maven build i.e mvn clean install
is failing with
TestPolicyACLs.testResourceACLs_dataMask:85->runTestsFromResourceFiles:107->runTests:122->lambda$runTests$0:127 » ConcurrentModification

### DIFF
--- a/agents-common/src/test/java/org/apache/ranger/plugin/policyengine/TestPolicyACLs.java
+++ b/agents-common/src/test/java/org/apache/ranger/plugin/policyengine/TestPolicyACLs.java
@@ -119,7 +119,7 @@ public class TestPolicyACLs {
             RangerPluginContext       pluginContext       = new RangerPluginContext(new RangerPluginConfig(serviceType, null, "test-policy-acls", "cl1", "on-prem", policyEngineOptions));
             RangerPolicyEngine        policyEngine        = new RangerPolicyEngineImpl(testCase.servicePolicies, pluginContext, null);
 
-            testCase.tests.parallelStream().filter(Objects::nonNull).forEach(oneTest -> {
+            testCase.tests.stream().filter(Objects::nonNull).forEach(oneTest -> {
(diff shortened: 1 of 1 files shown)


pradeepagrawal8184 (approved):

agents-common/src/test/java/org/apache/ranger/plugin/policyengine/TestPolicyACLs.java
```diff
@@ -119,7 +119,7 @@ private void runTests(InputStreamReader reader, String testName) {
             RangerPluginContext       pluginContext       = new RangerPluginContext(new RangerPluginConfig(serviceType, null, "test-policy-acls", "cl1", "on-prem", policyEngineOptions));
             RangerPolicyEngine        policyEngine        = new RangerPolicyEngineImpl(testCase.servicePolicies, pluginContext, null);
 
-            testCase.tests.parallelStream().filter(Objects::nonNull).forEach(oneTest -> {
+            testCase.tests.stream().filter(Objects::nonNull).forEach(oneTest -> {
```
vyommani, line 122: Please don't change it from parallelStream to stream. I’ve been investigating this and the parallel execution was an intentional change made as part of RANGER-5115.

We need more analysis to determine the correct fix for this test failure. The test only started failing a couple of hours ago in my local env, so I haven’t finished the root cause analysis yet.

If this is blocking, we can temporarily disable the test, but I recommend not modifying the test logic.

@ChinmayHegde24 If it’s urgent, please disable the test for now. We’ll analyze the failure and push a proper fix

ChinmayHegde24: @vyommani Sure , that will cause only issues in local build, in CI it's fine anyway so it'll not block there

ChinmayHegde24:
Closing this PR as it is handled in https://github.com/apache/ranger/pull/950

PR #950, opened 2026-05-17
RANGER-5602: Local build failure due to unit test failure in TestPolicyACLs

vyommani:

Switched Delegate from lazy computeIfAbsent to eager initialization in the constructor. All internal maps are now fully populated and wrapped with unmodifiableMap before the Delegate is published to the cache.

This eliminates the CME because HashMap.get() is thread-safe for concurrent reads when there are no writes. The previous computeIfAbsent caused concurrent writes on a shared HashMap when multiple threads accessed the same cached Delegate, which parallelStream exposed

## How was this patch tested?

after fix mvn test -Dtest=TestPolicyACLs#testResourceACLs_dataMask is passing.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/model/validation/RangerServiceDefHelper.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/model/validation/RangerServiceDefHelper.java
@@ -507,14 +507,23 @@ public class RangerServiceDefHelper {
 
                         this.hierarchies.put(policyType, Collections.unmodifiableSet(hierarchies));
                         hierarchyKeys.put(policyType, Collections.unmodifiableSet(hierachyKeys));
+                        Map<String, RangerResourceDef> wildcardMap = new HashMap<>(resources.size());
+                        for (RangerResourceDef res : resources) {
+                            if (res != null) {
+                                wildcardMap.put(res.getName(), createWildcardEnabledResourceDef(res));
+                            }
+                        }
+                        wildcardEnabledResourceDefs.put(policyType, Collections.unmodifiableMap(wildcardMap));
                     } else {
                         isValid = false;
                         hierarchies.put(policyType, EMPTY_RESOURCE_HIERARCHY);
                         hierarchyKeys.put(policyType, Collections.emptySet());
+                        wildcardEnabledResourceDefs.put(policyType, Collections.emptyMap());
                     }
                 } else {
                     hierarchies.put(policyType, EMPTY_RESOURCE_HIERARCHY);
                     hierarchyKeys.put(policyType, Collections.emptySet());
+                    wildcardEnabledResourceDefs.put(policyType, Collections.emptyMap());
                 }
             }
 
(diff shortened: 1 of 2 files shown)


vyommani:
@mneethiraj @pradeepagrawal8184 @rameeshm  please review I am not able to add reviewers.

agents-common/src/main/java/org/apache/ranger/plugin/model/validation/RangerServiceDefHelper.java
```diff
@@ -548,6 +557,15 @@ public Delegate(RangerServiceDef serviceDef, boolean checkForCycles) {
             LOG.debug("Found [{}] resource hierarchies for service [{}] update-date[{}]: {}", hierarchies.size(), serviceName, serviceDefFreshnessDate, hierarchies);
         }
 
+        private RangerResourceDef createWildcardEnabledResourceDef(RangerResourceDef resourceDef) {
+            if (resourceDef == null) {
+                return null;
```
pradeepagrawal8184: The constructor loop already guards with if (res != null) before calling createWildcardEnabledResourceDef. The null check here is harmless but redundant;

agents-common/src/main/java/org/apache/ranger/plugin/model/validation/RangerServiceDefHelper.java
```diff
@@ -686,31 +704,11 @@ RangerResourceDef getWildcardEnabledResourceDef(String resourceName, Integer pol
             if (policyType == null) {
```
pradeepagrawal8184, line 701: Should this condition be similar to line 621 ?

agents-common/src/main/java/org/apache/ranger/plugin/model/validation/RangerServiceDefHelper.java
```diff
@@ -507,14 +507,23 @@ public Delegate(RangerServiceDef serviceDef, boolean checkForCycles) {
 
                         this.hierarchies.put(policyType, Collections.unmodifiableSet(hierarchies));
                         hierarchyKeys.put(policyType, Collections.unmodifiableSet(hierachyKeys));
+                        Map<String, RangerResourceDef> wildcardMap = new HashMap<>(resources.size());
+                        for (RangerResourceDef res : resources) {
+                            if (res != null) {
+                                wildcardMap.put(res.getName(), createWildcardEnabledResourceDef(res));
```
pradeepagrawal8184: is there possibility of duplicate resource in resources object:
If resources contains duplicate names, later entries considered in wildcardMap.

vyommani: Good catch. You are correct that if resources contains duplicate names, the current eager init code uses HashMap.put() which would silently overwrite earlier entries with later ones — meaning the last match wins.

However the original lazy code had a different behavior — it iterated the list and broke on the first match

So the eager init introduces a subtle behavioral difference on duplicate resource names.

That said, in practice duplicate resource names in a RangerServiceDef should never occur — the service definition is validated upstream before being stored and a duplicate resource name would indicate a malformed service definition.
However to be defensively correct and to exactly preserve the original behavior (first match wins), we should change put to putIfAbsent

agents-common/src/main/java/org/apache/ranger/plugin/model/validation/RangerServiceDefHelper.java
vyommani: The if (policyType == null) condition at line 704 is pre-existing code — our change only replaced the lazy computeIfAbsent with a simple get.

You are correct that this inconsistency exists, but it is unrelated to the problem we are solving in this PR. Fixing it here would be scope creep and risks introducing unintended side effects.

Will raise a separate Jira to track this inconsistency and fix it in a dedicated PR where it can be properly reviewed and tested in isolation

What do you think about above approach ?

PR #953, opened 2026-05-18
RANGER-5553: Incorrect Tags count reported in SummaryTOTALTAGS metric from /service/metrics/json API

RakeshGuptaDev:

The SummaryTOTALTAGS metric returned by the /service/metrics/json API does not accurately reflect the actual number of Tags present in the system. The value appears to be inconsistent or incorrect when compared against the real tag count.

Steps to Reproduce:

Access the /service/metrics/json API.
Locate the SummaryTOTALTAGS metric in the response.
Compare the reported value with the actual number of Tags in the system.
Expected Result:
SummaryTOTALTAGS should match the actual count of Tags.

Actual Result:
SummaryTOTALTAGS displays an incorrect count.

## How was this patch tested?

Verified that the SummaryTOTALTAGS metric in the /service/metrics/json API correctly reflects the actual count of Tags present in the system.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/metrics/RangerMetricsFetcher.java
+++ b/security-admin/src/main/java/org/apache/ranger/metrics/RangerMetricsFetcher.java
@@ -57,6 +57,7 @@ public class RangerMetricsFetcher {
 
     private final HashMap<String, Long> summaryReusable = new HashMap<>();
     private final HashMap<Integer, Long> summaryPolicy  = new HashMap<>();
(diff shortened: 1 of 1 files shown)


kishorgollapalliwar (approved):
Look good to me. Please make sure to have enough DEV testing before merging the change.

kishorgollapalliwar:
commit: https://github.com/apache/ranger/commit/5ed6267d2021937a092336baeafa83dbec10ff30

PR #965, opened 2026-05-22
RANGER-5603: Upgrade Solr to 9.4.1

pradeepagrawal8184:

Ranger solr plugin install was failing because solr version is not compatible with jdk17; hence trying to upgrade solr to 9.4.1 as it has jdk17 support as per solr's doc.

## How was this patch tested?

I tested this patch in my docker env;  

@reviewers:
if possible please test this patch in your env also to ensure the changes are generic for all env.


ranger-plugin-classloader/src/main/java/org/apache/ranger/plugin/classloader/RangerPluginClassLoader.java
```diff
@@ -40,6 +40,23 @@
 public class RangerPluginClassLoader extends URLClassLoader {
```
rameeshm, line 40: @pradeepagrawal8184 This is bit invasive in doing. If the Jersey libraries are already present in the SOLR can we remove it from the SOLR?

dev-support/ranger-docker/scripts/solr/solr-ranger_audits/solrconfig.xml
```diff
@@ -417,7 +417,7 @@
                       to occupy. Note that when this option is specified, the size
                       and initialSize parameters are ignored.
       -->
-    <filterCache class="solr.FastLRUCache"
+    <filterCache class="solr.CaffeineCache"
```
rameeshm, line 420: How is the upgrade going to be handled?

dev-support/ranger-docker/scripts/hadoop/ranger-hadoop-setup.sh
```diff
@@ -66,6 +66,10 @@ else
     echo "WARNING: Tez directory not found at /opt/tez"
 fi
 
+mkdir -p /etc/ranger/dev_hdfs/policycache /etc/ranger/dev_yarn/policycache
```
rameeshm: Is this change related to Solr upgrade?

plugin-solr/src/main/java/org/apache/ranger/authorization/solr/authorizer/RangerSolrAuditHandler.java
```diff
@@ -31,7 +31,7 @@
 
 public class RangerSolrAuditHandler extends RangerMultiResourceAuditHandler {
     private static final String PROP_SOLR_PLUGIN_AUDIT_EXCLUDED_USERS = "ranger.solr.plugin.audit.excluded.users";
-    private static final String RANGER_AUDIT_COLLECTION               = "ranger_audits";
+    public static final String RANGER_AUDIT_COLLECTION                  = "ranger_audits";
```
rameeshm: Please check the code style on this patch. Many misaligned lines are there like this one

plugin-solr/src/main/java/org/apache/ranger/authorization/solr/authorizer/RangerSolrAuthorizer.java
```diff
@@ -685,6 +710,36 @@ private RangerAccessRequestImpl createBaseRequest(String userName, Set<String> u
         return rangerRequest;
     }
 
+    /**
```
rameeshm: Pluing service user normally authenticates into SOLR to commit into ranger_audits. We have policies for it, so wondering what is the purpose this method. Please check it out

ranger-plugin-classloader/src/main/java/org/apache/ranger/plugin/classloader/RangerPluginClassLoader.java
pradeepagrawal8184: Agreed that changing RangerPluginClassLoader globally is heavy. For Solr 9, Jetty already provides Jersey on the webapp classpath, so we can avoid bundling Jersey/HK2 in ranger-solr-plugin-impl and rely on the existing component-classloader delegation (revert the classloader change).

The classloader isolation was only needed because the plugin tarball still contained Jersey 2.39.1 while ranger-plugins-common also pulled 2.47, and delegation mixed classes across loaders. The fix is to exclude all org.glassfish.jersey.* / org.glassfish.hk2.* from the Solr plugin assembly (not ship duplicate jars), not to remove Jersey from Solr itself.

I’ll validate policy refresh (RangerAdminRESTClient) and Kerberos/docker Solr startup after dropping packaged Jersey and reverting the classloader change.

dev-support/ranger-docker/scripts/solr/solr-ranger_audits/solrconfig.xml
pradeepagrawal8184: Solr has removed FastLRUCache; if this is there then we get ClassNotFoundException regarding same; CaffeineCache is the new alternate in the SOLR 9.4 version.

dev-support/ranger-docker/scripts/hadoop/ranger-hadoop-setup.sh
pradeepagrawal8184: Probably not; i was getting error related to file permission's of these files; may be we can remove it and try again. the issue at my end could be intermittent.

plugin-solr/src/main/java/org/apache/ranger/authorization/solr/authorizer/RangerSolrAuditHandler.java
pradeepagrawal8184: Sure. will fix this.

dev-support/ranger-docker/scripts/hadoop/ranger-hadoop-setup.sh
pradeepagrawal8184: for now I will comment it out rather completely removing it

plugin-solr/src/main/java/org/apache/ranger/authorization/solr/authorizer/RangerSolrAuthorizer.java
```diff
@@ -127,6 +127,12 @@ public AuthorizationResponse authorize(AuthorizationContext context) {
         boolean isDenied = false;
 
         try {
+            if (isRangerAuditCollectionUpdate(context)) {
+                logger.debug("Allowing internal audit write to {}", RangerSolrAuditHandler.RANGER_AUDIT_COLLECTION);
+
+                return new AuthorizationResponse(200);
+            }
```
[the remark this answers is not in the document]
pradeepagrawal8184: Please review again

dev-support/ranger-docker/scripts/solr/solr-security.json
```diff
@@ -1,13 +1,4 @@
 {
```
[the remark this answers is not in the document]
pradeepagrawal8184: By keeping this it will throw below error with solr 9.4.1:
Caused by: java.lang.ClassNotFoundException: solr.KerberosPlugin
	at java.base/java.net.URLClassLoader.findClass(Unknown Source) ~[?:?]
	at java.base/java.lang.ClassLoader.loadClass(Unknown Source) ~[?:?]
	at java.base/java.net.FactoryURLClassLoader.loadClass(Unknown Source) ~[?:?]
	at java.base/java.lang.ClassLoader.loadClass(Unknown Source) ~[?:?]
	at java.base/java.lang.Class.forName0(Native Method) ~[?:?]
	at java.base/java.lang.Class.forName(Unknown Source) ~[?:?]

dev-support/ranger-docker/docker-compose.ranger.yml
```diff
@@ -110,7 +110,12 @@ services:
     volumes:
       - ./dist/keytabs/ranger-solr:/etc/keytabs
       - ./scripts/solr/solr-ranger_audits:/opt/solr/server/solr/configsets/ranger_audits/conf
+      # Core conf is copied once by solr-precreate; mount so Solr 9 cache updates apply without recreating the core
+      - ./scripts/solr/solr-ranger_audits:/var/solr/data/ranger_audits/conf
```
kumaab: please see if the resources in "volumes" can be mounted read-only, thanks.

kumaab (commented):
docker related changes look good.

ranger-plugin-classloader/src/main/java/org/apache/ranger/plugin/classloader/RangerPluginClassLoader.java
rameeshm: @pradeepagrawal8184 what I mean is if we need to remove from the SOLR Plugin and use what SOLR service is bringing in that would be better.

dev-support/ranger-docker/docker-compose.ranger.yml
pradeepagrawal8184: Mounted config, scripts, and keytab dirs as :ro on ranger, ranger-zk, and ranger-solr. Left ranger-kdc’s ./dist/keytabs writable because the KDC creates keytabs there via ktadd.

pom.xml
```diff
@@ -1126,12 +1128,15 @@
                 <module>agents-common</module>
                 <module>agents-cred</module>
                 <module>agents-installer</module>
+                <module>authz-api</module>
                 <module>common-utils</module>
                 <module>credentialbuilder</module>
+                <module>distro</module>
```
mneethiraj, line 1134: Is `distro` module necessary for `ranger-solr-plugin`?

security-admin/pom.xml
```diff
@@ -545,6 +545,11 @@
             <artifactId>unixauthclient</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
```
mneethiraj, line 548: Is it necessary to update the version of Solr client used in modules `security-admin`, `embeddedwebserver` and `agents-audit/dest-solr` modules? Does the older version client (8.11.3) not work with newer Solr server version (9.4.1) used in Docker setup?

PR #971, opened 2026-05-25
RANGER-5613:DockerFile of Ranger Audit Server has issue duplicate tools sets which are already in Ranger Base image

rameeshm:

JIRA is to fix the duplicate  toolset inclusion in the dockerfiles of audit server.

## How was this patch tested?
Tested with Build and running the docker of audit server and verifying.


kumaab (approved):

dev-support/ranger-docker/docker-compose.ranger-audit-ingestor.yml
```diff
@@ -70,177 +70,3 @@ networks:
   ranger:
     name: rangernw
     external: true
```
mneethiraj: `external: true` is seen only in Audit docker compose files; other docker-compose files don't use this flag. Please review and remove if this is not necessary.

rameeshm: @mneethiraj   I have addressed this. Wanted to tell that marking it external prevents it from tearing down the shared network on docker compose down. Since its already done by other dockerfiles we can follow the same, I shall check on the benefits of external and if need to addressed, shall do in another JIRA.

mneethiraj (approved):

kumaab:
Hi @rameeshm, Currently, I see that audit service is not part of CI workflow where all containers are brought up. Are we tracking this as part of a different jira? thanks.

rameeshm:
> Hi @rameeshm, Currently, I see that audit service is not part of CI workflow where all containers are brought up. Are we tracking this as part of a different jira? thanks.

@kumaab  We can take this up as a part of new Jira.

PR #962, opened 2026-05-21
RANGER-3899: Policy creation/updation takes more time when there are more users,groups,roles

ramackri:
## What changes were proposed in this pull request?

Optimized `PolicyRefUpdater` and related DAOs so policy create/update performs fewer database round-trips when a policy references many users, groups, and roles.

### Problem

When creating or updating a policy with a large number of principals (users, groups, roles), Ranger Admin was slow because:

1. **N+1 lookups** — Each user, group, role, resource, access type, condition, and data mask type was resolved with a separate DB query (`findByUserName`, `findByGroupName`, `findByRoleName`, etc.).
2. **Full ref-table rebuild on update** — On policy update, `cleanupRefTables()` deleted all rows from `x_policy_ref_user`, `x_policy_ref_group`, and `x_policy_ref_role` (and other ref tables), then re-inserted every principal even when nothing changed.

### Changes

1. **Batch lookups via named queries** — Added `IN (:names)` JPA named queries and DAO helpers to resolve principals and definitions in bulk:
   - `XXUser.getIdsByUserNames`
   - `XXGroup.getIdsByGroupNames`
   - `XXRole.getIdsByRoleNames`
   - `XXResourceDef.findByNamesAndPolicyId`
   - `XXAccessTypeDef.findByNamesAndServiceId`
   - `XXPolicyConditionDef.findByServiceDefIdAndNames`
   - `XXDataMaskTypeDef.findByNamesAndServiceId`

2. **Pre-resolved principal IDs** — `PolicyPrincipalAssociator` now accepts a pre-fetched principal ID from the bulk map, avoiding per-name lookups when the principal already exists.

3. **Selective cleanup on policy update** — Introduced `cleanupRefTablesForUpdate()` that:
   - Loads existing user/role/group ref rows for the policy
   - Deletes only principals removed from the policy
   - Skips re-insert for principals that are unchanged
   - Still performs full delete/rebuild for resources, access types, conditions, and data mask types (typically far fewer entries)

4. **Unified create/update flow** — `createNewPolMappingForRefTable()` takes a new `isCleanupRefTablesNeeded` flag:
   - **Create:** `false` (no prior ref rows)
   - **Update:** `true` (runs selective cleanup before insert)
   - `ServiceDBStore` no longer calls `cleanupRefTables()` separately before mapping

5. **Batch insert helper** — Added `batchInsert()` with bulk-mode handling and debug timing logs around `dao.batchCreate()`.

### Files changed (15)

| Area | Files |
|------|-------|
[...]

PR #980, opened 2026-05-26
RANGER-5614 : Performance improvement for role create/update with many users, groups, and sub-roles

ramackri:

Optimized `RoleRefUpdater` and related role-ref DAOs so role create/update performs fewer database round-trips when a role references many users, groups, and sub-roles. The design mirrors [RANGER-3899](https://issues.apache.org/jira/browse/RANGER-3899) / PR [#962](https://github.com/apache/ranger/pull/962) (`PolicyRefUpdater`).

### Problem

When creating or updating a role with a large number of principals, Ranger Admin was slow because:

1. **N+1 lookups** — Users, groups, and sub-roles were resolved with repeated per-name database calls during ref-table maintenance.
2. **Full ref-table rebuild on every update** — `createNewRoleMappingForRefTable()` always called `cleanupRefTables()`, deleting all rows from `x_role_ref_user`, `x_role_ref_group`, and `x_role_ref_role`, then re-inserting every principal even when nothing changed.
3. **Monolithic associator** — A single `RolePrincipalAssociator` took three collection parameters and populated one of them inside `doAssociate()`, which was harder to read and inconsistent with the refactored policy path.

#### Performance (`RoleRefUpdater`)

- Added `isRefTableCleanupRequired` to `createNewRoleMappingForRefTable()`.
- **Create** (`RoleDBStore`, `RoleREST`): `isRefTableCleanupRequired = false` — no selective cleanup on empty ref tables.
- **Update** (`RoleDBStore`, `RoleREST`, add/remove users/groups/roles, import update): `true` — `cleanupRefTablesForUpdate()` deletes only principals removed from the role; unchanged principals are not re-inserted.
- Batch principal ID resolution via `getIdsByUserNames` / `getIdsByGroupNames` / `getIdsByRoleNames` and `batchInsert()` for new ref rows.
- `cleanupRoleUsers` / `cleanupRoleGroups` / `cleanupRoleRoles` accept the DAO as a parameter (same pattern as `PolicyRefUpdater`).

#### Associator refactor (aligned with PR #962 review)

Replaced `RolePrincipalAssociator` with three focused inner classes:

| Class | `getRoleRef()` | `createRoleRef(Long)` |
|-------|----------------|------------------------|
| `RoleUserAssociator` | `XXRoleRefUser` for batch insert | Direct `create` on transaction commit |
| `RoleGroupAssociator` | `XXRoleRefGroup` | Direct `create` on transaction commit |
| `RoleRoleAssociator` | `XXRoleRefRole` | Direct `create` on transaction commit |

Call-site pattern (same as `PolicyRoleAssociator` / `PolicyGroupAssociator` / `PolicyUserAssociator`):

```java
Long userId = userNameIdMap.get(userName);
RoleUserAssociator associator = new RoleUserAssociator(userName, userId, roleId);

if (userId != null) {
    XXRoleRefUser userRef = associator.getRoleRef();
    if (userRef != null) {
        xxRoleRefUsers.add(userRef);
    }
} else if (isCreateNonExistentUGRs) {
    rangerTransactionSynchronizationAdapter.executeOnTransactionCommit(associator);
} else {
    throw restErrorUtil.createRESTException(...);
}

### Manual / functional testing (Done)

Environment: Ranger Admin UI at `http://localhost:6080` (minimal Docker: Postgres + Admin per `dev-support/RANGER-LOCAL-BUILD.md`).

#### Setup — create principals

1. Open **Settings** → **Users/Group/Permission** (or navigate to `http://localhost:6080/index.html#/users/roletab`).
2. **Users** tab — create **5 internal users**:
   - `ramk1`
   - `ramk2`
   - `ramackri1`
   - `ramackri2`
   - `ramackri3`
3. **Groups** tab — create **5 groups**:
   - `example`
   - `ramkGroup1`
   - `ramkGroup2`
   - `ramackrGroup1`
   - `ramackriGroup2`
   - `ramackriGroup3`
4. **Roles** tab — create **5 roles** (empty or minimal members):
   - `ramkRole1`
   - `ramkRole2`
   - `ramackrRole1`
   - `ramackriRole2`
   - `ramackriRole3`

#### Test role — `TestRole5` (create / update / incremental member changes)

5. **Roles** tab → **Create** role **`TestRole5`**.
6. Edit **`TestRole5`** and **add all** principals created above:
   - All 5 users  
   - All 5 groups  
   - All 5 sub-roles  
7. **Save** and confirm the role saves successfully and members appear in the UI.
8. **Update `TestRole5` incrementally** (repeat save after each step; verify UI and behavior):
   - Remove **one user**, save → confirm user removed from role, others unchanged.
   - Remove **one group**, save → same.
   - Remove **one sub-role**, save → same.
   - Add back **one** removed user/group/sub-role, save → confirm only new association is added (no duplicate members).
   - Remove **multiple** users/groups/sub-roles in one update, save → confirm selective cleanup (no full ref-table wipe).
9. **Optional stress check**: With `TestRole5` still holding many members, change membership several times (add/remove mix) and confirm:
   - Updates complete in reasonable time vs baseline (before RANGER-5614).
   - No duplicate entries in **Users / Groups / Roles** lists for `TestRole5`.
   - No errors in Ranger Admin logs (`docker logs ranger`).

#### DB sanity (optional)

10. For role id of `TestRole5`, verify ref tables reflect UI after an update:
    - `x_role_ref_user` — only current users  
    - `x_role_ref_group` — only current groups  
    - `x_role_ref_role` — only current sub-roles  

#### Expected results

- Create/update of roles with many users, groups, and sub-roles completes without timeout or error.
- Removing members updates only removed associations; unchanged members are not deleted and re-inserted.
- Adding members after a partial remove does not create duplicates.


## Related

- [RANGER-5637](https://issues.apache.org/jira/browse/RANGER-5637): CI fix — split `plugins-download-archives` job to fix `plugins-docker-build` timeouts.
  - Commit [30ab21a1d](https://github.com/apache/ranger/commit/30ab21a1d): split download job + curl retries in `download-archives.sh`
  - Commit [9b2735726](https://github.com/apache/ranger/commit/9b2735726): tune curl `--retry` to 3 and document flags

### CODE REVIEW

**Comment by ramackri** [comment]
image

**Comment by ramackri** [comment]
@mneethiraj
attaching benchmark testing results
[Apache Ranger Role API Performance Improvements Latency Matrix.pdf](https://github.com/user-attachments/files/28651473/Apache.Ranger.Role.API.Performance.Improvements.Latency.Matrix.pdf)

File: security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java [hunk 1]
```diff
@@ -210,6 +210,9 @@ public void createNewPolMappingForRefTable(RangerPolicy policy, XXPolicy xPolicy
             createPrincipalsIfAbsent = false;
         }
 
+        final boolean policyExists = xPolicy != null && xPolicy.getId() != null
+                && daoMgr.getXXPolicy().getById(xPolicy.getId()) != null;
```
mneethiraj: replace `getById(xPolicy.getId())` with `getCountById(xPolicy.getId()) > 0` - to avoid reading entire XXPolicy into memory.

agents-common/src/main/java/org/apache/ranger/plugin/store/RoleStore.java
```diff
@@ -28,9 +28,9 @@
 public interface RoleStore {
     void init() throws Exception;
 
-    RangerRole createRole(RangerRole role, Boolean createNonExistUserGroup) throws Exception;
+    RangerRole createRole(RangerRole role, Boolean createNonExistUserGroup, Boolean isRefTableCleanupRequired) throws Exception;
```
mneethiraj: Instead of updating signature of existing methods, add new methods with default implementation - to avoid breaking existing implementations.

```
public interface RoleStore {
...
  RangerRole createRole(RangerRole role, Boolean createNonExistUserGroup) throws Exception;

  RangerRole updateRole(RangerRole role, Boolean createNonExistUserGroup) throws Exception;

  default RangerRole createRole(RangerRole role, Boolean createNonExistUserGroup, Boolean isRefTableCleanupRequired) throws Exception {
    // ignore isRefTableCleanupRequired
    return createRole(role, createNonExistUserGroup);
  }

  default RangerRole updateRole(RangerRole role, Boolean createNonExistUserGroup, Boolean isRefTableCleanupRequired) throws Exception {
    // ignore isRefTableCleanupRequired
    return updateRole(role, createNonExistUserGroup);
  }
}

```

security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
```diff
@@ -551,25 +554,23 @@ public void cleanupPolicyRefGroups(Set<String> policyGroups, Long policyId, XXPo
...
-            this.xPolicy = xPolicy;
+        PolicyRoleAssociator(String name, Long roleId, XXPolicy xPolicy, boolean policyExists) {
+            this.name         = name;
+            this.roleId       = roleId;
+            this.xPolicy      = xPolicy;
+            this.policyExists = policyExists;
         }
 
         public XXPolicyRefRole getPolicyRef() {
             Long id = resolveRoleId(false);
```
mneethiraj: Call to `resolveRoleId()` is unnecessary when `policyExists` is `false`. Please review and update - here and other such occurrences.

security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
ramackri: Done

agents-common/src/main/java/org/apache/ranger/plugin/store/RoleStore.java
ramackri: Done

security-admin/src/main/java/org/apache/ranger/biz/PolicyRefUpdater.java
ramackri: Done

distro/src/main/assembly/pdp.xml
```diff
@@ -33,6 +33,7 @@
                 <include>org.apache.ranger:authz-embedded</include>
                 <include>org.apache.ranger:ranger-audit-core</include>
                 <include>org.apache.ranger:ranger-audit-dest-auditserver</include>
+                <include>org.apache.ranger:ranger-audit-dest-solr</include>
```
ramackri: Fixing the blow issue

The job is failing in the services-docker-build step with this critical error:
`
java.lang.ClassNotFoundException: org.apache.ranger.audit.destination.SolrAuditDestination
`

The Ranger PDP (Policy Decision Point) container is unable to initialize because the SolrAuditDestination class is missing from the classpath when the application starts up.

Root Cause
The SolrAuditDestination class is not being included in the built artifacts or Docker image. This typically happens when:

A required dependency is not properly declared in the POM
The jar containing this class is not being packaged into the distribution tarball
The Docker image build is not copying all necessary jars into the container
Solution
Check the PDP module's POM configuration to ensure the audit-related dependencies are properly included:

Verify that ranger-audit-core or the module containing SolrAuditDestination is listed as a dependency in the PDP module's pom.xml
Ensure the dependency is not marked with <scope>provided</scope> when it should be bundled
Check that the assembly/packaging configuration includes audit jars in the final distribution
Example fix in the PDP module's pom.xml:

XML
<dependency>
    <groupId>org.apache.ranger</groupId>
    <artifactId>ranger-audit-core</artifactId>
    <version>${project.version}</version>
</dependency>
Verify the Docker image includes audit jars:

Check the dev-support/ranger-docker/Dockerfile or Docker entrypoint scripts to ensure all required jars are copied into the container's lib directory, particularly audit-related dependencies.

The logs confirm the class is being referenced in configuration but not available at runtime, suggesting a packaging/dependency issue rather than a code logic problem.

mneethiraj: `org.apache.ranger:ranger-audit-dest-solr` shouldn't be packaged in PDP (and other plugin packages); with recent updates in RANGER-5632, all audit destinations except auditserver destination were removed from all plugins. PDP server also should only be sending logs to auditserver destination. I suggest updating `dev-support/ranger-docker/scripts/pdp/ranger-pdp-site.xml` for this.

ramackri: Thanks @mneethiraj  — agreed.
The Solr packaging in pdp.xml was added to unblock CI after ranger-pdp failed with ClassNotFoundException: SolrAuditDestination while Docker still had ranger.authz.audit.destination.solr=true.
Updated as suggested:
distro/src/main/assembly/pdp.xml — reverted the ranger-audit-dest-solr module and Solr-only HTTP client deps; PDP keeps only ranger-audit-dest-auditserver (aligned with RANGER-5632 / master).
dev-support/ranger-docker/scripts/pdp/ranger-pdp-site.xml — disabled Solr audit destination and enabled auditserver (url=http://ranger-audit-ingestor.rangernw:7081, with batch spool dir), matching other Docker plugins.
javax.inject in pdp/pom.xml is unchanged — that addresses the separate Jersey startup issue.

mneethiraj (approved):

PR #982, opened 2026-05-26
RANGER-5610:getResourceACLs for a principal should consider validitySchedule of principal for ACL creation

rameeshm:

getResourceACLs for a principal should consider validitySchedule of principal for ACL creation. Currently getResourceACLs returns a set of ACLs for the principal but if the principal is has a validity period which is expired, it is giving the ACLs which are there which is not correct. 

This scenario occurs in GDS where a principal validity period is expired and ACL still show the access given, even though there is no access.

Show Privileges in RangerHiveAuthorizer uses the getResourceACLs which also result in wrong permission sets shown when the principal validity period expired.

## How was this patch tested?

Tested in local vm + unit test.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerAbstractPolicyEvaluator.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerAbstractPolicyEvaluator.java
@@ -20,8 +20,12 @@
 package org.apache.ranger.plugin.policyevaluator;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.ranger.plugin.conditionevaluator.RangerValidityScheduleConditionEvaluator;
 import org.apache.ranger.plugin.model.RangerPolicy;
+import org.apache.ranger.plugin.model.RangerPolicy.RangerPolicyItem;
+import org.apache.ranger.plugin.model.RangerPolicy.RangerPolicyItemAccess;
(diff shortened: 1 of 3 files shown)


rameeshm:
@mneethiraj  Could you please review this patch and give your comments.

dineshkumar-yadav (approved):
LGTM

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerAbstractPolicyEvaluator.java
```diff
@@ -199,6 +204,7 @@ public void getResourceACLs(RangerAccessRequest request, RangerResourceACLs acls
 
             if (policyType == RangerPolicy.POLICY_TYPE_ACCESS) {
                 updateFromPolicyACLs(isConditionalMatch, acls, targetAccessTypes);
+                filterAclResultsByValiditySchedule(request, acls);
```
mneethiraj, line 207: @rameeshm  - `acls` might have permissions granted by other policies for this user/groups/roles. `filterAclResultsByValiditySchedule()` might remove them, right? That will not be correct. Instead of updating `acls` here, I suggest the caller should check if a permission is granted conditionally - like:

```
AccessResult user1Select =  acls.getUserACLs().get("user1").get("select");

boolean isAllowed = user1Select.getResult() == ACCESS_ALLOWED;
// valid values are:
//   ACCESS_DENIED(-1)
//   ACCESS_UNDETERMINED(0)
//   ACCESS_ALLOWED(1)
//   ACCESS_CONDITIONAL(2)
```

rameeshm: @mneethiraj  Thanks for the review. Will address this by reverting this patch.

PR #984, opened 2026-05-27
RANGER-5607 issue in  put plugins/services/id

abhiishek26:

the api put plugins/services/{id}  implemented in ServiceREST updateService. do not appropriately throws error when the service name field is left  blank in the  payload  or mismatch between ids passed through  payload and request url  . 
I compared this with the put apis present in ServiceREST and other modules there appropriate id mismatch or name absent errors are handled .
I have made changes in the implementation to properly handle those scenarios . 

## How was this patch tested?
All the unit tests were passed locally and the  bulild (mvn clean install  , mvn clean test ) successfully passed 

2 unit tests have been implemented to simulate id mismatch and servicename absent in payload scenarios

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
@@ -734,6 +734,26 @@ public class ServiceREST {
     @PreAuthorize("@rangerPreAuthSecurityHandler.isAPIAccessible(\"" + RangerAPIList.UPDATE_SERVICE + "\")")
     public RangerService updateService(RangerService service, @Context HttpServletRequest request) {
         LOG.debug("==> ServiceREST.updateService(): {}", service);
+        // if service.id and param 'id' are specified, service.id should be same as the param 'id'
+        // if service.id is null, then set param 'id' into service Object
+        if (request != null) {
+            String requestURI = request.getRequestURI();
+            if (requestURI != null) {
+                String[] parts = requestURI.split("/");
+                try {
+                    Long id = Long.parseLong(parts[parts.length - 1]);
+                    if (service.getId() == null) {
+                        service.setId(id);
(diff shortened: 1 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -734,6 +734,26 @@ public RangerService createService(RangerService service) {
     @PreAuthorize("@rangerPreAuthSecurityHandler.isAPIAccessible(\"" + RangerAPIList.UPDATE_SERVICE + "\")")
     public RangerService updateService(RangerService service, @Context HttpServletRequest request) {
         LOG.debug("==> ServiceREST.updateService(): {}", service);
+        // if service.id and param 'id' are specified, service.id should be same as the param 'id'
+        // if service.id is null, then set param 'id' into service Object
+        if (request != null) {
+            String requestURI = request.getRequestURI();
+            if (requestURI != null) {
+                String[] parts = requestURI.split("/");
```
rameeshm, line 742: Can this result in parts being null? Please check

abhiishek26: > Can this result in parts being null? Please check

Since the condition is if (requestURI != null). it will check that the request url is not null . if it is null it will not allow to enter in the if block and will not allow to do the split and form array   . 
and if we consider the case like this http://localhost:6080/service/plugins/services/.  where no id is passed in url although a  blank "" will be present in the  String[] parts array but it will throw NumberFormatException which has been handled .

To test the above cases I had created a build with these changes so if there no id provided which leads to null it throws status 405 . 
and when there is mismatch it throws  status 400 . 

Are my  these right or some change is required ?

pradeepagrawal8184 (approved):

pradeepagrawal8184 (approved):

PR #988, opened 2026-05-29
RANGER-5610:getResourceACLs for a principal should consider validitySchedule of principal for ACL creation

rameeshm:
getResourceACLs for a principal should consider validitySchedule of principal for ACL creation. Currently getResourceACLs returns a set of ACLs for the principal but if the principal is has a validity period which is expired, it is giving the ACLs which are there which is not correct.

This scenario occurs in GDS where a principal validity period is expired and ACL still show the access given, even though there is no access.

Show Privileges in RangerHiveAuthorizer uses the getResourceACLs which also result in wrong permission sets shown when the principal validity period expired. Fix will filter the getResourceACLs for the Prinicipal in RangerHiveAuthorizer if the validitySchedule is not allowing the access for the resource.

## How was this patch tested?

Tested in Local VM and Unit Tests.

### DIFF
--- a/hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
+++ b/hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
@@ -58,6 +58,7 @@ import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.thirdparty.com.google.common.collect.Sets;
 import org.apache.ranger.authorization.hadoop.constants.RangerHadoopConstants;
 import org.apache.ranger.authorization.utils.StringUtil;
+import org.apache.ranger.plugin.conditionevaluator.RangerValidityScheduleConditionEvaluator;
 import org.apache.ranger.plugin.model.RangerPolicy;
 import org.apache.ranger.plugin.model.RangerPolicy.RangerPolicyItemAccess;
 import org.apache.ranger.plugin.model.RangerRole;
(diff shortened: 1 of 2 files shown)


rameeshm:
@mneethiraj  Please review this patch. Thanks.

hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
```diff
@@ -2843,11 +2846,187 @@ private RangerResourceACLs getRangerResourceACLs(HivePrivilegeObject hiveObject,
...
+                break;
+        }
+
+        if (principalACLs != null) {
+            Map<String, RangerResourceACLs.AccessResult> accessMap = principalACLs.get(principal.getName());
+            if (accessMap != null) {
+                Map<String, Collection<String>> impliedGrants    = null;
+                Map<String, Collection<String>> gdsImpliedGrants = null;
+
+                if (hivePlugin != null) {
```
mneethiraj, line 2884: `hivePlugin` will not null here given the caller already references it at line 2847. I suggest removing the `if` at 2884.

hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
```diff
@@ -2843,11 +2846,187 @@ private RangerResourceACLs getRangerResourceACLs(HivePrivilegeObject hiveObject,
...
+                    RangerServiceDef serviceDef = hivePlugin.getServiceDef();
+                    if (serviceDef != null) {
```
mneethiraj, line 2886: Consider replacing this `if` block by initializing `impliedGrants` in line 2881 with:

```
final Map<String, Collection<String>> impliedGrants = hivePlugin.getServiceDefHelper() != null ? hivePlugin.getServiceDefHelper().getImpliedAccessGrants() : Collections.emptyMap();
```

hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
```diff
@@ -2843,11 +2846,187 @@ private RangerResourceACLs getRangerResourceACLs(HivePrivilegeObject hiveObject,
...
+                        impliedGrants = ServiceDefUtil.getExpandedImpliedGrants(serviceDef);
+                    }
+
+                    GdsPolicyEngine gdsPolicyEngine = hivePlugin.getGdsPolicyEngine();
+                    if (gdsPolicyEngine != null && gdsPolicyEngine.getGdsInfo() != null) {
+                        RangerServiceDef gdsServiceDef = gdsPolicyEngine.getGdsInfo().getGdsServiceDef();
```
mneethiraj, line 2892: Consider updating `GdsPolicyEngine` to expose `RangerServiceDefHelper` and use it to initialize `gdsImpliedGrants` in line 2882.

```
final Map<String, Collection<String>> gdsImpliedGrants = gdsPolicyEngine != null && gdsPolicyEngine.getServiceDefHelper() != null ? gdsPolicyEngine.getServiceDefHelper().getImpliedAccessGrants() : Collections.emptyMap();
```

hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
```diff
@@ -2843,11 +2846,187 @@ private RangerResourceACLs getRangerResourceACLs(HivePrivilegeObject hiveObject,
...
+                        if (gdsServiceDef != null) {
+                            gdsImpliedGrants = ServiceDefUtil.getExpandedImpliedGrants(gdsServiceDef);
+                        }
+                    }
+                }
+
+                Map<RangerPolicy.RangerPolicyItem, Boolean> policyItemActiveCache = new HashMap<>();
+                List<String> keysToRemove = new ArrayList<>();
```
mneethiraj, line 2900: `keysToRemove` ==> `permissionsToRemove`

hive-agent/src/main/java/org/apache/ranger/authorization/hive/authorizer/RangerHiveAuthorizer.java
```diff
@@ -2843,11 +2846,187 @@ private RangerResourceACLs getRangerResourceACLs(HivePrivilegeObject hiveObject,
...
+                                        policyItemActiveCache.put(policyItem, isActive);
+                                    }
+                                    if (isActive) {
+                                        hasActiveItem = true;
+                                        break;
+                                    }
+                                }
+                            }
+                            if (!hasActiveItem) {
+                                keysToRemove.add(entry.getKey());
```
mneethiraj, line 2936: @rameeshm  - if multiple policies grant this permission to the principal, removing the permission based on just one policy (line 2904) will result in incorrect ACLs.

rameeshm:
- This feature has to be handled by having a new API for getting all the the allowed ACLs for a given Principal. Changing the showPrivileges() API would be invasive change and not recommended, hence closing this PR

PR #992, opened 2026-06-02
RANGER-5623: Updated XUserMgr.java to solve UI consistency for put groupusers endpoint

krishnamuttevi:
When updating a user-group relationship using the PUT /service/xusers/groupusers API, the backend updates successfully and the UI "Users" tab looks fine. However, the "Groups" tab shows old data—it still lists the user under their old group and doesn't show them in the new one. This happens because the "Groups" view isn't getting the right group reference to trigger a proper refresh.

Solved it by:
Declared XXGroupuser and added validation checks for all mandatory fields to make sure the data is clean.

Explicitly added parentGroupId to vxGroupuser so the UI has the correct group context to update both tabs properly.

Verified that updating a user-group mapping now correctly updates both the "Users" and "Groups" tabs in the UI without leaving any stale data.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
@@ -1678,7 +1678,6 @@ public class XUserMgr extends XUserMgrBase {
 
         vXGroupUser.setParentGroupId(xGroup.getId());
 
-
         return super.updateXGroupUser(vXGroupUser);
     }
 
--- a/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
@@ -1653,6 +1653,10 @@ public class XUserMgr extends XUserMgrBase {
 
         xaBizUtil.blockAuditorRoleUser();
 
+        if (vXGroupUser == null) {
+            throw restErrorUtil.createRESTException(HttpServletResponse.SC_BAD_REQUEST, "Groupuser not found", true);
+        }
+
         XXGroupUser xxGroupUser = null;
 
         if (vXGroupUser.getId() != null) {
--- a/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
@@ -1653,6 +1653,32 @@ public class XUserMgr extends XUserMgrBase {
 
         xaBizUtil.blockAuditorRoleUser();
 
+        XXGroupUser xxGroupUser = null;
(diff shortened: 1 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
```diff
@@ -1653,6 +1653,31 @@ public VXGroupUser updateXGroupUser(VXGroupUser vXGroupUser) {
 
         xaBizUtil.blockAuditorRoleUser();
 
+        XXGroupUser xxGroupUser = null;
+
+        if (vXGroupUser.getId() != null) {
```
pradeepagrawal8184: what if vXGroupUser itself is null.

krishnamuttevi: I've updated the function to include a validation check for invalid inputs. Could you please take a look at the latest commit

security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
```diff
@@ -1653,6 +1653,35 @@ public VXGroupUser updateXGroupUser(VXGroupUser vXGroupUser) {
 
         xaBizUtil.blockAuditorRoleUser();
 
+        if (vXGroupUser == null) {
+            throw restErrorUtil.createRESTException(HttpServletResponse.SC_BAD_REQUEST, "Groupuser not found", true);
+        }
+
+        XXGroupUser xxGroupUser = null;
```
ChinmayHegde24: We can remove this dead variable and better to group checks in single **if** like below
```
if (vXGroupUser.getId() != null &&
        daoManager.getXXGroupUser().getById(vXGroupUser.getId()) == null) {
    throw restErrorUtil.createRESTException(
            "Group-User mapping not found with ID: " + vXGroupUser.getId(),
            MessageEnums.DATA_NOT_FOUND);
}
```

security-admin/src/test/java/org/apache/ranger/biz/TestXUserMgr.java
ChinmayHegde24, line 1: new code has 4 new throw paths, better to cover those in tests as well.

krishnamuttevi: I have updated the unit tests for them

security-admin/src/main/java/org/apache/ranger/biz/XUserMgr.java
krishnamuttevi: Thanks for letting me know have updated the code

ChinmayHegde24 (approved):
Looks good to me now, Thanks

PR #999, opened 2026-06-06
RANGER-5632: update plugins packaging to have audit-server as the only audit destination

mneethiraj:

updated plugins packaging to remove all audit destinations other than audit-server destination.

## How was this patch tested?

Verified the following using docker setup:
- verified successful Ranger build
- verified successful startup of all Ranger services
- verified all plugins successfully download policies from Ranger


kumaab (approved):
LGTM

PR #920, opened 2026-04-16
RANGER-5554: Refactor Ranger Admin bootstrap to use runtime XML config only

kumaab:
## What changes were proposed in this pull request?

Refactors the ranger-admin Docker bootstrap flow to use runtime `XML` configuration and environment variables instead of the legacy installer path (`setup.sh`, `install.properties`, and related setup scripts).

## Key changes:

- Switches admin container startup to runtime-driven config using `ranger-admin-site.xml`
- Adds Python helpers for DB setup, `XML` config handling, password bootstrap, and logging
  - Introduces logging with config to be used for logging across python scripts. All log statements use logger now.
  - Passwords are now driven from env variables in `docker compose`, eliminates hardcoded passwords in `.xml` file.
  - Authn and user/password validations/resets are now handled via apache-ranger python client.
  - Helper methods are grouped in`UserPasswordBootstrap` and DB specific classes in `dba.py` for maintainability and readability.
- Removes dependency on `ranger-admin-install-postgres/mysql/oracle.properties`
- Updates `ranger.sh` to prepare runtime config, wait for DB readiness, seed initial passwords, and bootstrap services
- Updates Docker build/compose to use mounted XML config assets and the new helper scripts
- Added support for MySQL and Oracle DB as well.

## How was this patch tested?
- built the ranger-admin image
- ranger-admin starts with mounted XML config and env-based passwords
- Verified Ranger Admin comes up successfully and UI login works, no errors seen in catalina.out
- Verified default services are created
- Verified fresh install picks up configured admin/usersync/tagsync passwords
- Verified Solr plugin is able to download policies

## Pending
- Handling of Java and DB patches.

PR #1004, opened 2026-06-09
RANGER-5530: Fix for KMS API not working in docker kerberos env

vikaskr22:
## Ranger-KMS APIs (including dev_kms TestConnection) were breaking due to authentication & authorisation error after kerberos implementation in Docker env.

This PR contains fix for following:
1. Enabled KMS to run in kerberos: KMS uses kms-site.xml in it's authentication flow, not core-site.xml. So included kms-site.xml with correct config.
2. Updated KDC entrypoint to generate HTTP.keytab for KMS , It is required for SPNEGO.
3. Corrected hadoop_auth_local rule as per kerberos principal
4. Updated dev_kms KMS URL connection
5. Fixed Class loading errors due to duplicate jars being added in ews/lib & WEB-INF/lib

After kerberos fix, it was throwing following error:
```
2026-06-09 07:19:48,155 ERROR [http-nio-9292-exec-9] o.a.c.c.C.[.[.[.[webservices-driver] (DirectJDKLog.java:170) - Servlet.service() for servlet [webservices-driver] in context with path [] threw exception [org.glassfish.jersey.server.ContainerException: java.lang.NoClassDefFoundError: Could not initialize class org.apache.ranger.plugin.util.JsonUtilsV2] with root cause
java.lang.ExceptionInInitializerError: Exception java.lang.LinkageError: loader constraint violation: loader 'app' wants to load class com.fasterxml.jackson.databind.ObjectMapper. A different class with the same name was previously loaded by org.apache.catalina.loader.ParallelWebappClassLoader @54336c81. (com.fasterxml.jackson.databind.ObjectMapper is in unnamed module of loader org.apache.catalina.loader.ParallelWebappClassLoader @54336c81, parent loader 'app') [in thread "http-nio-9292-exec-3"]
at java.base/java.lang.ClassLoader.defineClass1(Native Method)
at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
………………………………………
………………………………………
at org.apache.ranger.plugin.util.JsonUtilsV2.<clinit>(JsonUtilsV2.java:34)
at org.apache.hadoop.crypto.key.kms.server.KMSJSONWriter.writeTo(KMSJSONWriter.java:61)
at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.invokeWriteTo(WriterInterceptorExecutor.java:242)
at ……………..
…………………………..
at org.glassfish.jersey.internal.Errors.process(Errors.java:244)
```

As per my analysis, it happened due to **duplicate jackson-*:jar** jars available inside distro/kms.xml . Once inside "ews/webapp/WEB-INF/lib/" and once inside "ews/lib".

**RCA:** "ranger-common-utils" was part of "ews/lib" and Jackson jars were part of both moduleSets.  JsonUtilsV2.java tries to load **com.fasterxml.jackson.databind.ObjectMapper** . It's part of Jackson jars and it was already loaded by another class loader. That is, once it  gets loaded by the app loader, then the webapp loader tries to load it again → loader constraint violation → JsonUtilsV2 cannot initialize.

**Fix:** :  Removed ranger-common-utils dependency & Jackson-*.jar from ews/lib and added ranger-common-utils inside "ews/webapp/WEB-INF/lib/".

## How was this patch tested?

-mvn build is working
- setup docker env locally, KMS is UP & running. 
- dev_kms Test connection is working now
- Zone key creation , rollover , delete etc. is working.

### DIFF
--- a/dev-support/ranger-docker/scripts/admin/create-ranger-services.py
+++ b/dev-support/ranger-docker/scripts/admin/create-ranger-services.py
@@ -104,7 +104,7 @@ hbase = RangerService({'name': 'dev_hbase', 'type': 'hbase',
 
 kms = RangerService({'name': 'dev_kms', 'type': 'kms',
                      'configs': {'username': 'keyadmin', 'password': 'rangerR0cks!',
-                                 'provider': 'http://ranger-kms.rangernw:9292',
+                                 'provider': 'kms://http@ranger-kms.rangernw:9292/kms',
                                  'policy.download.auth.users': 'rangerkms',
                                  'tag.download.auth.users': 'rangerkms',
                                  'userstore.download.auth.users': 'rangerkms',
--- a/dev-support/ranger-docker/scripts/kms/ranger-kms.sh
+++ b/dev-support/ranger-docker/scripts/kms/ranger-kms.sh
@@ -37,6 +37,7 @@ then
     if [ "${KERBEROS_ENABLED}" == "true" ]
     then
       cp ${RANGER_SCRIPTS}/core-site.xml ${RANGER_HOME}/kms/ews/webapp/WEB-INF/classes/conf/core-site.xml
+      cp ${RANGER_SCRIPTS}/kms-site.xml ${RANGER_HOME}/kms/ews/webapp/WEB-INF/classes/conf/kms-site.xml
     fi
 
     touch "${RANGER_HOME}"/.setupDone
--- a/distro/src/main/assembly/kms.xml
+++ b/distro/src/main/assembly/kms.xml
@@ -201,6 +201,7 @@
                             <include>com.fasterxml.jackson.core:jackson-core:jar:${fasterxml.jackson.version}</include>
                             <include>com.fasterxml.jackson.core:jackson-databind:jar:${fasterxml.jackson.databind.version}</include>
                             <include>org.javassist:javassist</include>
+                            <include>org.apache.ranger:ranger-common-utils</include>
                         </includes>
                     </dependencySet>
                 </dependencySets>
@@ -227,7 +228,6 @@
                     <include>org.apache.hadoop:hadoop-client-api:jar:${hadoop.version}</include>
                     <include>org.apache.hadoop:hadoop-client-runtime:jar:${hadoop.version}</include>
                     <include>org.apache.solr:solr-solrj:jar:${solr.version}</include>
-                    <include>org.apache.ranger:ranger-common-utils</include>
                     <include>org.apache.ranger:ranger-authz-api</include>
                     <include>org.apache.ranger:ranger-plugins-common</include>
                     <include>org.apache.ranger:credentialbuilder</include>
@@ -240,9 +240,6 @@
                     <include>org.slf4j:slf4j-api</include>
                     <include>ch.qos.logback:logback-classic:jar:${logback.version}</include>
                     <include>ch.qos.logback:logback-core:jar:${logback.version}</include>
-                    <include>com.fasterxml.jackson.core:jackson-annotations:jar:${fasterxml.jackson.version}</include>
-                    <include>com.fasterxml.jackson.core:jackson-core:jar:${fasterxml.jackson.version}</include>
-                    <include>com.fasterxml.jackson.core:jackson-databind:jar:${fasterxml.jackson.version}</include>
                     <include>org.codehaus.woodstox:stax2-api</include>
                     <include>com.fasterxml.woodstox:woodstox-core</include>
                     <include>org.apache.commons:commons-text:jar:${commons.text.version}</include>
--- a/distro/src/main/assembly/kms.xml
+++ b/distro/src/main/assembly/kms.xml
@@ -228,6 +228,7 @@
                     <include>org.apache.hadoop:hadoop-client-api:jar:${hadoop.version}</include>
                     <include>org.apache.hadoop:hadoop-client-runtime:jar:${hadoop.version}</include>
                     <include>org.apache.solr:solr-solrj:jar:${solr.version}</include>
+                    <include>org.apache.ranger:ranger-common-utils</include>
                     <include>org.apache.ranger:ranger-authz-api</include>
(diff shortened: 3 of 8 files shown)


dev-support/ranger-docker/scripts/kms/ranger-kms.sh
```diff
@@ -37,6 +37,7 @@ then
     if [ "${KERBEROS_ENABLED}" == "true" ]
     then
       cp ${RANGER_SCRIPTS}/core-site.xml ${RANGER_HOME}/kms/ews/webapp/WEB-INF/classes/conf/core-site.xml
+      cp ${RANGER_SCRIPTS}/kms-site.xml ${RANGER_HOME}/kms/ews/webapp/WEB-INF/classes/conf/kms-site.xml
```
kumaab, line 40: We're trying to get away from setup.sh installation, for ex: https://github.com/apache/ranger/pull/920 
Could you please try mounting `./scripts/kms/kms-site.xml` to `${RANGER_HOME}/kms/ews/webapp/WEB-INF/classes/conf/kms-site.xml` directly using docker-compose.ranger-kms.yaml.

kumaab (changes requested):

dev-support/ranger-docker/scripts/kms/ranger-kms.sh
vikaskr22: @kumaab , thanks for your review !

Abhishek, Is it okay to merge this PR right now as it is blocking many testing that I do in docker. This PR contains fixes to make KMS run in Kerberos enabled env. There are some other functional test related work going on that is blocked as KMS is not working.

I can create a new JIRA and can work to refactor KMS docker bootstrap to use runtime XML config separately.

What's your thought on this ?

kumaab: Sure @vikaskr22, sounds good.

pradeepagrawal8184 (approved):

kms/scripts/ranger-kms
```diff
@@ -112,7 +112,7 @@ fi
 
 KMS_CONF_DIR=${RANGER_KMS_EWS_DIR}/webapp/WEB-INF/classes/conf
 SERVER_NAME=rangerkms
-cp="-cp ${RANGER_KMS_EWS_CONF_DIR}:${RANGER_KMS_EWS_DIR}/lib/*:${JAVA_HOME}/lib/*:${RANGER_HADOOP_CONF_DIR}/*:$CLASSPATH"
+cp="-cp ${RANGER_KMS_EWS_CONF_DIR}:${RANGER_KMS_EWS_DIR}/lib/*:${RANGER_KMS_EWS_DIR}/webapp/WEB-INF/lib/*:${JAVA_HOME}/lib/*:${RANGER_HADOOP_CONF_DIR}/*:$CLASSPATH"
```
vyommani: Adding WEB-INF/lib/* to the JVM-level classpath fixes the immediate issue but potentially reintroduces the same classloader duplication pattern for any jar that exists in both ews/lib/* and WEB-INF/lib/*. Could you check if there's any overlap between the two directories in the assembled KMS package? Ideally the two classpaths should be disjoint — otherwise we may just be shifting the LinkageError to a different class down the line

distro/src/main/assembly/kms.xml
```diff
@@ -201,6 +201,7 @@
                             <include>com.fasterxml.jackson.core:jackson-core:jar:${fasterxml.jackson.version}</include>
                             <include>com.fasterxml.jackson.core:jackson-databind:jar:${fasterxml.jackson.databind.version}</include>
                             <include>org.javassist:javassist</include>
+                            <include>org.apache.ranger:ranger-common-utils</include>
```
vyommani, line 204: The fix makes sense  moving ranger-common-utils to WEB-INF/lib eliminates the dual classloader loading of Jackson. A couple of things worth verifying:

1--> Are there any other jars remaining in the ews/lib moduleSet that are also present in WEB-INF/lib? If so, we'd have the same loader constraint risk for those classes once WEB-INF/lib/* is added to the JVM classpath (see comment on kms/scripts/ranger-kms).

2--> Is there anything in the KMS startup lifecycle that runs before the webapp classloader is active and depends on ranger-common-utils? Since it's no longer in ews/lib, it won't be available at the JVM level until Tomcat loads the webapp

vyommani: @pradeepagrawal8184  could you weigh in on this? I think you'd have better insight into whether the classpath overlap is actually a concern here. Check my other comment as well.

vikaskr22: @vyommani , I have addressed the concern you raised. I have added the required dependency to be loaded by the same class loader where it is required. Earlier JsonUtils2 was being loaded by one loader and dependent jar was being loaded by another loader. Now this would not be the case.

Pls review one more time. Thanks.

kms/scripts/ranger-kms
vikaskr22: Thanks. I have addressed the concerns you raised. Now change in this file is not required. hence it's not even part of this PR now.

dev-support/ranger-docker/scripts/admin/create-ranger-services.py
```diff
@@ -104,7 +104,7 @@ def service_not_exists(service):
 
 kms = RangerService({'name': 'dev_kms', 'type': 'kms',
                      'configs': {'username': 'keyadmin', 'password': 'rangerR0cks!',
-                                 'provider': 'http://ranger-kms.rangernw:9292',
+                                 'provider': 'kms://http@ranger-kms.rangernw:9292/kms',
```
kumaab, line 107: I'm assuming `Test connection` works after this change.

kumaab (approved):

dev-support/ranger-docker/scripts/admin/create-ranger-services.py
kumaab: I missed the PR details, it's tested.

PR #1011, opened 2026-06-11
RANGER-5635: ranger.admin.allow.unauthenticated.download.access is honored only when Kerberos is enabled

vyommani:

Enforce ranger.admin.allow.unauthenticated.access and ranger.admin.allow.unauthenticated.download.access regardless of whether Ranger Admin runs with Kerberos, so the configured value is applied consistently. 

## How was this patch tested?

unit tests added


ramackri:
LGTM

mneethiraj (approved):

ramackri (approved):

PR #1030, opened 2026-06-23
RANGER-5654: Fix Solr audit dispatcher Kerberos TGT relogin (No key to store) via keytab-aware JAAS relogin

ramackri:
Fixes [RANGER-5654](https://issues.apache.org/jira/browse/RANGER-5654): Solr audit dispatcher stops indexing audits into Kerberos-protected Solr after TGT refresh/relogin when `useTicketCache=true` (the shipped default).

### What changes were proposed in this pull request?

**Alternative B — keytab-aware JAAS relogin** (replaces earlier config-only approach):

| File | Change |
|------|--------|
| `agents-audit/core/.../AbstractKerberosUser.java` | Add `performRelogin()` — in-place `loginContext.login()` for keytab users at TGT renewal (skips `logout()` first) |
| `agents-audit/core/.../KerberosJAASConfigUser.java` | Enable in-place relogin when JAAS `useKeyTab=true` |
| `agents-audit/core/.../KerberosAction.java` | Use `performRelogin()` on SecurityException retry |
| `security-admin/.../solr/krb/AbstractKerberosUser.java` | Same (Admin Solr audit queries share `KerberosAction`) |
| `security-admin/.../solr/krb/KerberosJAASConfigUser.java` | Same |
| `security-admin/.../solr/krb/KerberosAction.java` | Same |
| `security-admin/.../AbstractKerberosUserTest.java` | Unit test: keytab relogin skips `logout()` |

**Site XML reverted** — `useTicketCache` stays **`true`** (shipped default) in both:

- `audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml`
- `dev-support/ranger-docker/scripts/audit-dispatcher/ranger-audit-dispatcher-solr-site.xml`

No config change required for operators; the Java fix addresses relogin with the existing default.

### Problem

The Solr dispatcher consumes from Kafka but eventually stops writing to Solr when Kerberos is enabled. Logs show `Failure in sending audits into Solr` and `No key to store`. Kafka offsets advance; Solr document counts stay flat until dispatcher restart.

**Root cause (causal chain):**

1. `SolrAuditDestination` uses `KerberosJAASConfigUser` + `KerberosAction`.
2. **Every Solr write** → `KerberosAction.execute()` → `checkTGTAndRelogin()`.
3. When the TGT passes ~80% of its lifetime (`TICKET_RENEW_WINDOW = 0.80` in `AbstractKerberosUser`), that method did **`logout(); login()`** — by design in `KerberosAction`, not caused by `useKeyTab=true`.
4. Shipped config has **`useKeyTab=true`** (correct for a keytab daemon) **and** **`useTicketCache=true`** (shipped default).
5. After `logout()`, `login()` with `useTicketCache=true` makes `Krb5LoginModule` use the ticket cache; with no valid cache entry → **"No key to store"** → dispatcher stuck until restart.

**Fix:** For keytab-backed JAAS users, renew via `loginContext.login()` on the existing `LoginContext`/`Subject` without `logout()` first. Non-keytab users and fallback paths keep `logout(); login()`.

#### Unit tests

- `AbstractKerberosUserTest.checkTGTAndRelogin_keytabUserSkipsLogoutWhenLoggedIn` — verifies in-place keytab relogin calls `login()` but not `logout()` at the 80% TGT window.
- `KerberosActionTest` — existing coverage unchanged.

#### Manual testing (Docker Tier 3 audit stack with Kerberos)

Environment: `dev-support/ranger-docker` Tier 3 compose — Admin, KDC, Kafka, ingestor, Solr dispatcher, Hive plugin.

**Verify with shipped `useTicketCache=true` + Alternative B JAR:**

1. Deploy updated `ranger-audit-core` into Solr dispatcher; confirm site XML has `useTicketCache=true`.
2. Clean Kerberos ticket cache (`kdestroy` / remove `/tmp/krb5cc_*`); restart Solr dispatcher.
3. Confirm clean JAAS login (`Successful login for rangerauditserver/...`).
4. Hive beeline SELECT → full pipeline PASS: plugin → ingestor → Kafka → Solr dispatcher → Solr `repo:dev_hive` count increases; Admin `accessAudit` totalCount increases.

**Note:** Flipping `useTicketCache` from `false` to `true` on a running container without clearing stale ticket cache can fail on initial login — that is a lab artifact, not the production relogin bug (which triggers at ~80% TGT lifetime).

### Upgrade note

No site XML change required. Redeploy Solr dispatcher (and Admin if using Kerberized Solr audit queries) with the updated tarball/JARs containing the `AbstractKerberosUser` fix.

---

##  is `useTicketCache` disabled in all Ranger modules?

**Short answer:** No — Ranger uses **two Kerberos auth patterns**. Only the **JAAS outbound client** path has a `useTicketCache` setting. Plugins downloading policies/tags/roles use a different path and do **not** use ticket cache.

#### Pattern 1 — UGI keytab (no JAAS `useTicketCache`)

Most plugins and sync components authenticate with Hadoop `UserGroupInformation.loginUserFromKeytab()` and refresh via `checkTGTAndReloginFromKeytab()`. This path **does not use JAAS `useTicketCache` at all**.

| Component | How it authenticates |
|-----------|---------------------|
| **Plugins → Admin** (policies, tags, roles, user store) | `RangerBasePlugin` UGI keytab login; `RangerAdminRESTClient` → `MiscUtil.executePrivilegedAction()` → `checkTGTAndReloginFromKeytab()` |
| **Plugin → ingestor audits** | `RangerAuditServerDestination` UGI + `checkTGTAndReloginFromKeytab()` |
| **HDFS audit dispatcher** | `UserGroupInformation.loginUserFromKeytabAndReturnUGI()` |
| **TagSync / UserSync** | UGI keytab or `SecureClientLogin.loginUserFromKeytab()` |
| **Service plugins** (Solr, KMS, Storm lookup, etc.) | `SecureClientLogin.loginUserFromKeytab()` |

#### Pattern 2 — JAAS outbound client (has `useTicketCache`)

Used for **outbound Kerberos clients** to Solr/Kafka via `KerberosAction` + `KerberosJAASConfigUser` (proactive relogin at ~80% TGT).

| Component | `useTicketCache` |
|-----------|------------------|
| **Audit ingestor → Kafka** | **`false`** (hardcoded in `AuditServerConstants`) |
| **Kafka plugin → cluster** | **`false`** (hardcoded in `ServiceKafkaClient`) |
| **Schema Registry plugin lookup** | **`false`** (hardcoded JAAS string) |
| **Trino plugin audit** (docker sample) | **`false`** |
| **Ranger Admin → Solr** (Access Audit UI) | **Configurable** via site XML; postgres docker install sets **`false`** |
| **Solr audit dispatcher → Solr** | **`true`** (shipped default; fixed by this PR's Java relogin) |
| **PDP Kerberos auth** | **`false`** |

Admin reads JAAS from site XML at runtime (`SolrMgr` → `InMemoryJAASConfiguration` → `KerberosJAASConfigUser`). Shipped `install.properties` leaves `audit_jaas_client_option_useTicketCache=` blank — operators must set it for Kerberized Solr audit queries. Recommended value for keytab-backed Admin: **`false`**.

### SPNEGO acceptor — `useTicketCache=true` is intentional

`MiscUtil` sets `useTicketCache=true` only for the **SPNEGO acceptor** (`isInitiator=false`) — inbound HTTP Kerberos for the web UI / SPNEGO filter. That is intentional and **unrelated to policy download or Solr indexing**.

### Conclusion for this PR

- **Plugins / TagSync / UserSync / HDFS dispatcher** → UGI keytab; no ticket cache involved.
- **JAAS outbound clients** → keytab daemons often use `useTicketCache=false` (ingestor Kafka, Admin docker), but Solr dispatcher ships `useTicketCache=true`; this PR fixes relogin in Java so that default works.
- **SPNEGO acceptor** → `useTicketCache=true` by design (different role).

---

## Review response: alternate to disabling ticket cache?

Proactive refresh **already exists**. `KerberosAction` calls `checkTGTAndRelogin()` before every Solr write; at ~80% TGT lifetime it relogins by design.

The failure is not missing refresh — it is **`logout()` + `login()` with `useTicketCache=true` on a keytab-only daemon**. After logout the ticket cache is empty, so relogin hits **"No key to store"** instead of re-reading the keytab.

A background refresh thread would **not** fix this: the proactive refresh path already runs; the break is specifically **JAAS relogin after `logout()`** when the login module is told to use an empty ticket cache instead of the keytab.

### Alternatives considered

| | **A: UGI pattern** | **B: AbstractKerberosUser fix (this PR)** | **Config: useTicketCache=false** |
|--|-------------------|------------------------------------------|-----------------------------------|
| **Mechanism** | Hadoop UGI + `checkTGTAndReloginFromKeytab()` | In-place keytab `loginContext.login()` at TGT renewal | Existing `logout(); login()` re-reads keytab |
| **Java changes** | Yes — dispatcher + destination | **Yes — shared krb utils (2 copies)** | No |
| **Touches Admin Solr?** | Only if `SolrAuditDestination` path changes | **Yes** (shared `KerberosAction`) | No |
| **Shipped useTicketCache=true** | N/A | **Works** | Requires config change |

**This PR implements Alternative B** so shipped `useTicketCache=true` works without operator config changes. Setting `useTicketCache=false` remains valid belt-and-suspenders for keytab-only daemons (ingestor Kafka convention).

---

## Test plan

- [ ] `mvn test -pl security-admin -Dtest=AbstractKerberosUserTest,KerberosActionTest`
- [ ] Rebuild / redeploy Solr dispatcher with updated JARs.
- [ ] Confirm `xasecure.audit.jaas.Client.option.useTicketCache=true` in live config (shipped default).
- [ ] Restart Solr dispatcher with clean ticket cache; confirm clean JAAS login in logs.
- [ ] Trigger audits through Tier 3 stack; confirm Solr document count increases.
- [ ] Confirm Admin Access Audit UI shows new events.

### DIFF
--- a/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
+++ b/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
@@ -38,10 +38,12 @@ public class KerberosJAASConfigUser extends AbstractKerberosUser {
 
     private final String        configName;
     private final Configuration config;
+    private final boolean       optionUseKeyTab;
 
     public KerberosJAASConfigUser(final String configName, final Configuration config) {
-        this.configName = configName;
-        this.config     = config;
+        this.configName      = configName;
+        this.config          = config;
+        this.optionUseKeyTab = getBooleanOption(JAAS_USE_KEYTAB);
     }
 
     @Override
@@ -70,25 +72,26 @@ public class KerberosJAASConfigUser extends AbstractKerberosUser {
      */
     @Override
     protected boolean useKeytabRelogin() {
-        return isJaasOptionTrue(JAAS_USE_KEYTAB);
+        return optionUseKeyTab;
     }
 
-    private boolean isJaasOptionTrue(String optionName) {
+    private boolean getBooleanOption(String optionName) {
+        boolean                 ret     = false;
         AppConfigurationEntry[] entries = config.getAppConfigurationEntry(configName);
 
-        if (entries == null) {
-            return false;
-        }
+        if (entries != null) {
+            for (AppConfigurationEntry entry : entries) {
+                Object value = entry.getOptions().get(optionName);
 
-        for (AppConfigurationEntry entry : entries) {
-            Object value = entry.getOptions().get(optionName);
+                if (value != null && Boolean.parseBoolean(value.toString())) {
+                    ret = true;
 
-            if (value != null && "true".equalsIgnoreCase(value.toString())) {
-                return true;
+                    break;
+                }
             }
         }
 
-        return false;
+        return ret;
     }
 
     @Override
--- a/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
+++ b/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
@@ -34,6 +34,8 @@ import javax.security.auth.login.LoginException;
 public class KerberosJAASConfigUser extends AbstractKerberosUser {
     private static final Logger LOG = LoggerFactory.getLogger(KerberosJAASConfigUser.class);
 
+    private static final String JAAS_USE_KEYTAB = "useKeyTab";
+
     private final String        configName;
     private final Configuration config;
 
@@ -58,6 +60,34 @@ public class KerberosJAASConfigUser extends AbstractKerberosUser {
         }
 
         return ret;
+    }
+
+    /**
+     * Solr/Kafka outbound JAAS clients (audit dispatcher, plugin Solr destination) use
+     * {@code useKeyTab=true}. Opt those principals into in-place keytab relogin so shipped
+     * {@code useTicketCache=true} does not fail at TGT renewal with {@code "No key to store"}.
+     */
+    @Override
+    protected boolean useKeytabRelogin() {
+        return isJaasOptionTrue(JAAS_USE_KEYTAB);
+    }
+
+    private boolean isJaasOptionTrue(String optionName) {
+        AppConfigurationEntry[] entries = config.getAppConfigurationEntry(configName);
+
+        if (entries == null) {
+            return false;
+        }
+
+        for (AppConfigurationEntry entry : entries) {
+            Object value = entry.getOptions().get(optionName);
+
+            if (value != null && "true".equalsIgnoreCase(value.toString())) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     @Override
--- a/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
+++ b/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
@@ -63,9 +63,10 @@ public class KerberosJAASConfigUser extends AbstractKerberosUser {
     }
 
     /**
-     * Solr/Kafka outbound JAAS clients (audit dispatcher, plugin Solr destination) use
-     * {@code useKeyTab=true}. Opt those principals into in-place keytab relogin so shipped
-     * {@code useTicketCache=true} does not fail at TGT renewal with {@code "No key to store"}.
+     * Solr/Kafka outbound JAAS clients (audit dispatcher, plugin Solr
+     * destination) use {@code useKeyTab=true}. Opt those principals into
+     * in-place keytab relogin so shipped {@code useTicketCache=true} does not
+     * fail at TGT renewal with {@code "No key to store"}.
      */
     @Override
     protected boolean useKeytabRelogin() {
--- a/audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
+++ b/audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
@@ -175,7 +175,7 @@
 
     <property>
         <name>xasecure.audit.jaas.Client.option.useTicketCache</name>
-        <value>false</value>
+        <value>true</value>
         <description>Allow use of cached Kerberos tickets for Solr authentication</description>
     </property>
 
--- a/audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
+++ b/audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
@@ -175,8 +175,8 @@
 
     <property>
         <name>xasecure.audit.jaas.Client.option.useTicketCache</name>
-        <value>true</value>
-        <description>Allow use of cached Kerberos tickets for Solr authentication</description>
+        <value>false</value>
+        <description>Use keytab for Solr auth (ticket cache breaks relogin with "No key to store")</description>
     </property>
 
     <property>
--- a/audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
+++ b/audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
@@ -176,7 +176,7 @@
     <property>
         <name>xasecure.audit.jaas.Client.option.useTicketCache</name>
         <value>false</value>
-        <description>Use keytab for Solr auth (ticket cache breaks relogin with "No key to store")</description>
+        <description>Allow use of cached Kerberos tickets for Solr authentication</description>
     </property>
 
     <property>
--- a/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/AbstractKerberosUser.java
+++ b/agents-audit/core/src/main/java/org/apache/ranger/audit/utils/AbstractKerberosUser.java
@@ -152,9 +152,14 @@ public abstract class AbstractKerberosUser implements KerberosUser {
     }
 
     /**
-     * Re-login a user from keytab if TGT is expired or is close to expiry.
+     * Proactively renew credentials when the TGT is missing or has passed
+     * {@link #TICKET_RENEW_WINDOW} (80%) of its lifetime.
      *
-     * @throws LoginException if an error happens performing the re-login
+     * <p>Called on every {@link KerberosAction#execute()} before the privileged action runs,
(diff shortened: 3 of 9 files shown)


audit-server/audit-dispatcher/dispatcher-solr/src/main/resources/conf/ranger-audit-dispatcher-solr-site.xml
```diff
@@ -175,7 +175,7 @@
 
     <property>
         <name>xasecure.audit.jaas.Client.option.useTicketCache</name>
-        <value>true</value>
+        <value>false</value>
```
mneethiraj: Is there an alternate to disabling use of ticket cache? Like having a thread refresh kerberos ticket before it expires? Is ticket cache disabled in all Ranger modules, like Ranger admin (to fetch audit logs from Solr), plugins (to download policies/tags/role/..)?

ramackri: Plugins / TagSync / UserSync / HDFS dispatcher → UGI keytab; no ticket cache involved.
JAAS outbound clients (Admin Solr queries, Solr dispatcher, ingestor Kafka, Kafka plugin) → useTicketCache=false is the correct convention for keytab-backed daemons, but it is not hardcoded everywhere; Admin relies on installer/site XML config.
SPNEGO acceptor → useTicketCache=true by design (acceptor role).
The Solr dispatcher was the outlier among long-running audit daemons: JAAS client + proactive KerberosAction relogin with useTicketCache=true. RANGER-5654 aligns it with ingestor Kafka and the recommended Admin/docker postgres setting.

ramackri: alternate to disabling ticket cache?
Proactive refresh already exists — we don't need a separate background thread. The Solr dispatcher uses KerberosAction, which calls checkTGTAndRelogin() before every Solr write. At ~80% of TGT lifetime that path intentionally does logout(); login(). So the issue isn't missing refresh; it's that relogin after logout() fails when useTicketCache=true on a keytab-only daemon: the ticket cache is empty, Krb5LoginModule doesn't fall back to the keytab cleanly, and we see "No key to store" until restart.

Alternatives that avoid setting useTicketCache=false would be Java changes, and they work at different layers:

UGI/keytab pattern (like HDFS dispatcher and plugin→ingestor) — initialize UserGroupInformation from the keytab in AuditSolrDispatcher and run Solr writes via MiscUtil.executePrivilegedAction() / checkTGTAndReloginFromKeytab() instead of KerberosAction. That avoids JAAS logout()/login() on the write path, but it's a broader auth-path change in shared SolrAuditDestination code.
Shared SolrAuditDestination means the Solr write logic lives in one class in the agents-audit library, and more than one component uses it — not only the Solr audit dispatcher.



Change AbstractKerberosUser.checkTGTAndRelogin() — for useKeyTab=true, renew in place (e.g. loginContext.login() without logout() first) instead of always doing logout(); login(). That would affect all KerberosAction callers (Admin Solr queries, plugins writing to Solr, etc.), so the blast radius is wider and behavior is more subtle across JDK/Krb5 versions.
That must be done in both copies of AbstractKerberosUser (agents-audit and security-admin), and it affects Admin Solr, plugins writing to Solr, and the dispatcher — not just the Solr dispatcher.

agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
```diff
@@ -60,6 +62,35 @@ public String getPrincipal() {
...
+     * in-place keytab relogin so shipped {@code useTicketCache=true} does not
+     * fail at TGT renewal with {@code "No key to store"}.
+     */
+    @Override
+    protected boolean useKeytabRelogin() {
+        return isJaasOptionTrue(JAAS_USE_KEYTAB);
+    }
+
+    private boolean isJaasOptionTrue(String optionName) {
+        AppConfigurationEntry[] entries = config.getAppConfigurationEntry(configName);
```
mneethiraj, line 80: To be consistent with most of Ranger code base, prefer a single `return` statement per method:

```
boolean                 ret     = false;
AppConfigurationEntry[] entries = config.getAppConfigurationEntry(configName);

if (entries != null) {
  for (AppConfigurationEntry entry : entries) {
    Object value = entry.getOptions().get(optionName);

    if (value != null && Boolean.parseBoolean(value.toString())) {
      ret = true;

      break;
    }
  }

  return ret;
}

```

agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
```diff
@@ -60,6 +62,35 @@ public String getPrincipal() {
...
+     * destination) use {@code useKeyTab=true}. Opt those principals into
+     * in-place keytab relogin so shipped {@code useTicketCache=true} does not
+     * fail at TGT renewal with {@code "No key to store"}.
+     */
+    @Override
+    protected boolean useKeytabRelogin() {
+        return isJaasOptionTrue(JAAS_USE_KEYTAB);
+    }
+
+    private boolean isJaasOptionTrue(String optionName) {
```
mneethiraj: Consider renaming `isJaasOptionTrue(String optionName)` => `getBooleanOption(String optionName)`

agents-audit/core/src/main/java/org/apache/ranger/audit/utils/KerberosJAASConfigUser.java
```diff
@@ -60,6 +62,35 @@ public String getPrincipal() {
...
 
+    /**
+     * Solr/Kafka outbound JAAS clients (audit dispatcher, plugin Solr
+     * destination) use {@code useKeyTab=true}. Opt those principals into
+     * in-place keytab relogin so shipped {@code useTicketCache=true} does not
+     * fail at TGT renewal with {@code "No key to store"}.
+     */
+    @Override
+    protected boolean useKeytabRelogin() {
+        return isJaasOptionTrue(JAAS_USE_KEYTAB);
```
mneethiraj: Instead of iterating through config entries during every call, consider computing the value in the constructor.

```
public class KerberosJAASConfigUser extends AbstractKerberosUser {
  ...

  private final boolean optionUseKeyTab;

  public KerberosJAASConfigUser(final String configName, final Configuration config) {
    this.configName      = configName;
    this.config          = config;
    this.optionUseKeyTab = getBooleanOption(JAAS_USE_KEYTAB);
  }

  ...

  @Override
  protected boolean useKeytabRelogin() {
    return optionUseKeyTab;
  }

  ...
}
  
```

mneethiraj (approved):

PR #1053, opened 2026-07-07
RANGER-5667: Follow-on updates for RANGER-5628

fmorg-git:

A couple PR review comments came in after the PR associated with [RANGER-5628](https://issues.apache.org/jira/browse/RANGER-5628) was already merged. Essentially, if a Policy Condition has an `"action-matches"` condition, but the request does not supply an `s3Action` in the `RequestContext`, the request must be denied, and this is not being done currently. This ticket will address this issue.

Furthermore, a second comment related to unnecessary object allocation and this ticket also addresses that one.

Separately, it was identified that a non-empty list of actions, if they are all just whitespace, needs to be treated as if there were no action restrictions. This ticket addresses that as well.

Finally, separately, I noticed that for the Ozone service definition, pressing the "+" button to add a new rule under "Allow Conditions” section was causing an error to throw, and I traced it to a change I made in RANGER-5628.  Basically, PolicyPermissionItem.jsx is pushing `undefined` on the `addPolicyItem()` call, and my `usePruneStaleConditions.js` hook wasn't properly checking for that undefined.  This ticket makes that fix as well.

## How was this patch tested?
unit tests and end-to-end smoke tests locally with Ozone and Ranger

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/conditionevaluator/RangerActionMatcher.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/conditionevaluator/RangerActionMatcher.java
@@ -26,7 +26,7 @@ import org.slf4j.LoggerFactory;
 public class RangerActionMatcher extends RangerAbstractConditionEvaluator {
     private static final Logger LOG = LoggerFactory.getLogger(RangerActionMatcher.class);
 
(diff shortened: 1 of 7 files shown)


fmorg-git:
FYI @mneethiraj and @rameeshm

mneethiraj (approved):

mneethiraj (approved):

PR #1055, opened 2026-07-09
RANGER-5624: Fix inconsistent updatedBy masking in groupName API

krishnamuttevi:
Currently, there is an inconsistency in the masking behavior of the updatedBy field in VXGroup responses across group retrieval APIs. When a group is fetched by its ID (GET /xusers/groups/{id}), the field is correctly masked to prevent metadata leakage. However, retrieving the exact same group by its name ( GET /xusers/groups/groupName/ group_name} ) bypasses this masking and exposes the actual user value (e.g., "Admin"). This inconsistency leads to an unreliable API contract and the unintended exposure of sensitive user-related metadata.

This was solved by:

Resolve the group ID using the requested group name directly within the xuserrest layer.

Route the subsequent retrieval call through the existing xUserMgr.getXGroup(id) method.

Update the following unit tests accordingly

Align the underlying logic so both endpoints share the exact same retrieval pipeline, ensuring the updatedBy field is consistently masked across all responses.


ChinmayHegde24 (approved):
LGTM, Add this version in JIRA affects version part.

ramackri (approved):

PR #1072, opened 2026-07-14
RANGER-5628: Gate Ozone action-matches policy behind feature flags

ramackri:
## Summary

Aligns Ozone `action-matches` gating (RANGER-5628) with the **single-property** design from master PR [#1073](https://github.com/apache/ranger/pull/1073) (RANGER-5687), adapted for `ranger-2.9`.

**One property controls everything** (default **off**):

| Property | File |
|----------|------|
| `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition` | `ranger-admin-site.xml` / `ranger-admin-default-site.xml` |

Propagates via service-def option `enableActionMatcherInPoliciesCondition` to:

- **Admin UI** — `RangerServiceDefService.applyActionMatcherInPoliciesConditionHiddenOption()` filters or restores `action-matches` when serving the ozone service-def API
- **Ozone plugin** — `RangerOzoneAuthorizer.isOzoneActionPolicyEnabled()` reads the option from the downloaded service-def
- **STS / inline** — action gating is confined to `plugin-ozone`; generic `RangerInlinePolicyEvaluator` always evaluates actions if present on a grant
- **DB patch J10065** — applies embedded policy conditions and persists `enableActionMatcherInPoliciesCondition` in `def_options`; excludes `action-matches` when flag is off at patch time

### Runtime fix (commit `4107076af`)

After J10065 persists `enableActionMatcherInPoliciesCondition=false` in DB `def_options`, **admin site config is authoritative at runtime** (same fix as [#1073](https://github.com/apache/ranger/pull/1073) `c345373fe`):

- Setting the property to `true` in `ranger-admin-site.xml` and **restarting Admin** takes effect even when DB still has `false`
- `action-matches` is restored from the embedded ozone service-def when enabled but missing from DB (e.g. J10065 removed it earlier)
- No manual `def_options` SQL or re-running J10065 is required for Admin UI / API behavior

**Audit:** When flag is **off**, authorizer sets `request.action` to coarse access type (`read`/`write`) so Solr `access` is populated (pre-5628 behavior). S3 action is used only when flag is **on**.

**Related:** Builds on Ozone action-matches work (#1054 / `PatchForOzoneServiceDefPolicyConditionUpdate_J10065`).

## Ozone action-policy gating (`isOzoneActionPolicyEnabled`)

[...]

PR #1073, opened 2026-07-14
RANGER-5687: Add feature flag to artifacts for RANGER-5628

fmorg-git:

It was requested to add a feature flag for the work in [RANGER-5628](https://issues.apache.org/jira/browse/RANGER-5628). This ticket adds it.

The flag is configured as follows in `ranger-admin-site.xml`:

```
<property>
<name>ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition</name>
<value>true</value>
</property>
```

This code is mostly generated with Cursor with some manual edits by me.

https://issues.apache.org/jira/browse/RANGER-5687

## How was this patch tested?
smoke test locally with ozone and ranger with ranger flag off and on


ramackri (approved):

agents-common/src/main/java/org/apache/ranger/plugin/policyevaluator/RangerInlinePolicyEvaluator.java
```diff
@@ -52,11 +53,13 @@ public class RangerInlinePolicyEvaluator {
     private final RangerInlinePolicy   policy;
     private final RangerPolicyEngine   policyEngine;
     private final List<GrantEvaluator> grants;
+    private final boolean              enableOzoneActionPolicy;
```
mneethiraj: `RangerInlinePolicyEvaluator` is not the appropriate place for a service specific flag like `enableOzoneActionPolicy`. This flag must be confined only within Ozone plugin module.

ramackri: @mneethiraj Fixed this review comment in ranger-2.9 PR :https://github.com/apache/ranger/pull/1072/changes/9be9a47590d6c561cd0db0db16f230a594b0f600
i will update this PR as well

agents-common/src/main/java/org/apache/ranger/plugin/model/RangerServiceDef.java
```diff
@@ -42,6 +42,7 @@ public class RangerServiceDef extends RangerBaseModelObject implements java.io.S
     public static final String OPTION_ENABLE_DENY_AND_EXCEPTIONS_IN_POLICIES = "enableDenyAndExceptionsInPolicies";
     public static final String OPTION_ENABLE_IMPLICIT_CONDITION_EXPRESSION   = "enableImplicitConditionExpression";
     public static final String OPTION_ENABLE_TAG_BASED_POLICIES              = "enableTagBasedPolicies";
+    public static final String OPTION_ENABLE_OZONE_ACTION_POLICY             = "enableOzoneActionPolicy";
```
mneethiraj: 1. I suggest no changes to agents-common module
2. Have `RangerOzoneAuthorizer.isOzoneActionPolicyEnabled()` and `RangerServiceDefService.applyOzoneActionPolicyHiddenOption()` call exisitng `ServiceDefUtil.getBooleanValue(plugin.getServiceDef().getOptions(), "enableOzoneActionPolicy", defaultValue)`

ramackri: @mneethiraj  We have removed changes in agents-common module

fmorg-git:
hi @ramackri - thanks for helping address the comments by @mneethiraj.  I tested these changes locally with Ozone and Ranger smoke tests and they look good to me.  So if they look good to Ranger team and there are no more outstanding comments, please go ahead.  Thanks!

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
```diff
@@ -101,4 +109,52 @@ protected RangerServiceDef mapEntityToViewBean(RangerServiceDef vObj, XXServiceD
...
+        if (serviceDefOptions == null) {
+            serviceDefOptions = new HashMap<>();
+            serviceDef.setOptions(serviceDefOptions);
+        }
+
+        if (serviceDefOptions.get(OPTION_ENABLE_OZONE_ACTION_POLICY) == null) {
+            boolean enableOzoneActionPolicy = config.getBoolean(PROP_ENABLE_OZONE_ACTION_POLICY, false);
+
+            serviceDefOptions.put(OPTION_ENABLE_OZONE_ACTION_POLICY, Boolean.toString(enableOzoneActionPolicy));
+            serviceDef.setOptions(serviceDefOptions);
```
fimugdha: This `enableOzoneActionPolicy` option is getting added to all the serviceDef, even though they don't support : `RangerActionMatcher` policy conditionevaluator.

The `RangerActionMatcher` policy conditionevaluator is currently getting used only for ozone serviceDef, then the option should be enabled true/false for ozone serviceDef using the xml config value and for other serviceDefs it should be set to false.

The name of the option should be kept generic `enableActionMatcherInPoliciesCondition` and the config name for `ranger-admin-site.xml` can be - `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition`

cc @mneethiraj / @rameeshm

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
```diff
@@ -18,6 +18,7 @@
 package org.apache.ranger.patch;
```
fimugdha, line 18: Don't we need a logic to update the serviceDef `options` config here ?

cc @mneethiraj / @rameeshm

ramackri: @fimugdha  yes we need this logic 
it is mentioned in this PR description (Upgrade flow from 2.8.0 )
 :https://github.com/apache/ranger/pull/1072

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
ramackri: This enableOzoneActionPolicy option is getting added to all the serviceDef, even though they don't support : RangerActionMatcher policy conditionevaluator. ---> valid option

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
fimugdha: Discussing about this below global options object and not about the policyConditions object having `action-matches`, could you please confirm how the ozone option going to get handled in case of upgrade scenario for each serviceDefs ?

```
options: {
    "enableDenyAndExceptionsInPolicies": "true",
    "enableTagBasedPolicies": "true",
    "enableOzoneActionPolicy": "false"
}

File: plugin-ozone/src/main/java/org/apache/ranger/authorization/ozone/authorizer/RangerOzoneAuthorizer.java [hunk 5]
```diff
@@ -168,7 +171,11 @@ public boolean checkAccess(IOzoneObj ozoneObject, RequestContext context) {
         rangerResource.setOwnerUser(context.getOwnerName());
         rangerRequest.setResource(rangerResource);
         rangerRequest.setAccessType(accessType);
-        rangerRequest.setAction(context.getS3Action());
+        if (isOzoneActionPolicyEnabled(plugin)) {
```
mneethiraj, line 174: Is it necessary to set `action` field differently based on the feature flag? I suggest setting it to `context.getS3Action()` irrespective of the feature flag.

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
fmorg-git: hi all - just for my understanding - if enableOzoneActionPolicy is added to all serviceDef and they don't support RangerActionMatcher, does it affect any behavior?

ramackri: RangerActionMatcher is only wired in the ozone service def (ranger-servicedef-ozone.json).
Non-ozone service defs never expose action-matches, so enableOzoneActionPolicy on them is misleading noise.

ramackri: We will fix it

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
ramackri: What that code does (J10065)
This runs once when you upgrade to Ranger 2.9 and admin starts up. It updates the ozone service definition in the database.

Step 1 — Read the admin flag
final boolean enableActionMatcherInPoliciesCondition =
    RangerAdminConfig.getInstance().getBoolean(
        RangerServiceDefService.PROP_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION, false);
It reads ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition from admin config. Default is false.

Step 2 — Decide which policy conditions ozone gets
if (enableActionMatcherInPoliciesCondition) {
    updatedPolicyConditions = new ArrayList<>(embeddedPolicyConditions);
} else {
    // copy all conditions EXCEPT "action-matches"
}
Flag ON → ozone gets all embedded conditions, including action-matches (the S3 Action dropdown in the UI).
Flag OFF → ozone gets everything except action-matches.
Step 3 — Save to DB
dbOzoneServiceDef.setPolicyConditions(updatedPolicyConditions);
That list is written to the DB as ozone’s policy condition definitions (what Admin UI and plugins see as available conditions).

What this code does not do 
It only touches policyConditions.

It does not update def_options — the JSON on x_service_def, e.g.:

{
  "enableTagBasedPolicies": "true",
  "enableDenyAndExceptionsInPolicies": "true",
  "enableActionMatcherInPoliciesCondition": "false"
}
That JSON is a separate knob. The Ozone plugin reads it to know whether to enforce S3 actions (GetObject, etc.) or only coarse read/write.

yes @fimugdha  we need this logic” means: extend J10065 (or a follow-on patch) to manage def_options on upgrade, not only policyConditions.
@fmorg-git  Please add if i missed anything here

kumaab:
@fmorg-git, please ensure to add the feature flag in dev-support/ranger-docker and test it in docker based environments.

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
ramackri: Pre-upgrade (ozone DB)
Stock Ranger 2.9 / unreleased 3.0 (feature not shipped):

{
  "enableTagBasedPolicies": "true",
  "enableDenyAndExceptionsInPolicies": "true"
}
No enableOzoneActionPolicy. No enableActionMatcherInPoliciesCondition until upgrade or first admin sync.

Flag OFF (false in site.xml)
Ozone DB def_options after J10065:

{
  "enableTagBasedPolicies": "true",
  "enableDenyAndExceptionsInPolicies": "true",
  "enableActionMatcherInPoliciesCondition": "false"
}
Ozone API options (GET service-def):

{
  "enableTagBasedPolicies": "true",
  "enableDenyAndExceptionsInPolicies": "true",
  "enableActionMatcherInPoliciesCondition": "false"
}
J10065 adds enableActionMatcherInPoliciesCondition: "false".

Flag ON (true in site.xml)
Ozone DB def_options after J10065:

{
  "enableTagBasedPolicies": "true",
  "enableDenyAndExceptionsInPolicies": "true",
  "enableActionMatcherInPoliciesCondition": "true"
}
Ozone API options (GET service-def):

{
  "enableTagBasedPolicies": "true",
  "enableDenyAndExceptionsInPolicies": "true",
  "enableActionMatcherInPoliciesCondition": "true"
}
J10065 adds enableActionMatcherInPoliciesCondition: "true".

| Option | Pre-upgrade (stock) | Flag OFF after J10065 | Flag ON after J10065 |
|--------|---------------------|------------------------|----------------------|
| `enableTagBasedPolicies` | `"true"` | `"true"` | `"true"` |
| `enableDenyAndExceptionsInPolicies` | `"true"` | `"true"` | `"true"` |
| `enableActionMatcherInPoliciesCondition` | absent | `"false"` | `"true"` |
| `enableOzoneActionPolicy` | absent | absent | absent |

Only ozone service def will get impacted 
@fimugdha Are you looking for this sort this behaviour ?

ramackri: @fimugdha  Addressed this

ramackri:
> @fmorg-git, please ensure to add the feature flag in dev-support/[ranger-docker](https://issues.apache.org/jira/browse/RANGER-docker) and test it in docker based environments.

@kumaab  Addressed this

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
fimugdha: @ramackri,

 All the serviceDefs currently in Ranger have `enableDenyAndExceptionsInPolicies` & `enableTagBasedPolicies` options. 

The updated fix in the pull-request file > RangerServiceDefService.java  for ozone-sts feature flag is doing two things :
 
1st - setting `enableActionMatcherInPoliciesCondition` now as false for all the serviceDef except for ozone it is setting the value for this option depending upon the xml config : `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition`.

2nd - Add the actionMatcher policyCondtion only to ozone serviceDef depending upon the xml config.

The J10065 java patch should add this option `enableActionMatcherInPoliciesCondition` to false for all the serviceDefs except for ozone. For ozone, it will be checking the boolean from the xml config for this option.

This J10065 java patch should not update touch the existing options : `enableDenyAndExceptionsInPolicies` and `enableTagBasedPolicies` for any serviceDefs (including ozone).

fmorg-git:
hi all, something appears to be wrong. I just tried my Ozone/Ranger smoke tests with the feature flag set to true, but it is not working now (this used to work before).  So something with the changes between the last time I said it was working and now seems to be different.  I updated with the new name `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition` but it still doesn't work.

plugin-ozone/src/main/java/org/apache/ranger/authorization/ozone/authorizer/RangerOzoneAuthorizer.java
```diff
@@ -384,4 +395,36 @@ private static String toRangerPermission(ACLType acl) {
 
         return null;
     }
+
+    private static boolean isOzoneActionPolicyEnabled(final RangerBasePlugin plugin) {
+        if (plugin == null || plugin.getServiceDef() == null) {
```
mneethiraj: I suggest replacing lines 400 to 410 with:
```
final RangerServceDef     serviceDef = plugin != null ? plugin.getServiceDef() : null;
final Map<String, String> options    = serviceDef != null ? serviceDef.getOptions() : null;

if (options == null) {
  return false;
}

return ServiceDefUtil.getBooleanValue(options, OPTION_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION, false) ||
       ServiceDefUtil.getBooleanValue(options, LEGACY_OPTION_ENABLE_OZONE_ACTION_POLICY, false);

File: security-admin/src/main/webapp/react-webapp/src/views/PolicyListing/AddUpdatePolicyForm.jsx [hunk 7]
```diff
@@ -511,15 +511,18 @@ export default function AddUpdatePolicyForm() {
...
+            policyConditionUpdatedJSON(serviceCompData?.policyConditions || []),
             function (m) {
               if (m.name == val.type) {
                 return m;
               }
             }
           );
 
-          if (!isEmpty(conditionObj.uiHint)) {
+          // Service-def conditions can be feature-flagged off (e.g. ozone action-matches).
```
mneethiraj: I suggest removing references to "feature-flag" here. From the UI perspective, it will simply ignore conditions in policies that are not specified in the service-def.

Same for the comment in line 715 as well.

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
```diff
@@ -131,7 +136,21 @@ private void updateOzoneServiceDef() {
                 return;
             }
 
-            dbOzoneServiceDef.setPolicyConditions(embeddedPolicyConditions);
+            final boolean enableActionMatcherInPoliciesCondition = RangerAdminConfig.getInstance().getBoolean(RangerServiceDefService.PROP_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION, false);
```
mneethiraj: Are changes in `PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java` necessary, given `RangerServiceDefService.java` handles service-def updates based on feature-flag while retrieving the service-def from the database? If no additional cases are handled here, I suggest reverting changes in `PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java`.

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
```diff
@@ -101,4 +109,112 @@ protected RangerServiceDef mapEntityToViewBean(RangerServiceDef vObj, XXServiceD
...
+
+        Map<String, String> serviceDefOptions = serviceDef.getOptions();
+
+        if (serviceDefOptions == null) {
+            serviceDefOptions = new HashMap<>();
+            serviceDef.setOptions(serviceDefOptions);
+        }
+
+        if (!StringUtils.equalsIgnoreCase(serviceDef.getName(), EmbeddedServiceDefsUtil.EMBEDDED_SERVICEDEF_OZONE_NAME)) {
+            serviceDefOptions.put(OPTION_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION, Boolean.FALSE.toString());
```
mneethiraj: Automatically disabling `action-matcher` condition in all service-defs doesn't make sense. This makes it impossible for a service-def to include this custom condition. This doesn't make sense.

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
```diff
@@ -21,17 +21,25 @@
...
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
 @Service
 @Scope("singleton")
 public class RangerServiceDefService extends RangerServiceDefServiceBase<XXServiceDef, RangerServiceDef> {
+    public static final String PROP_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION = "ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition";
```
mneethiraj: `PROP_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION` - rename this to include Ozone in the name, to make it clear that this is about Ozone service-def.

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
fimugdha: @mneethiraj, Good point. Since the action-matcher is currently being enabled for Ozone, the suggestion was to keep OPTION_ENABLE_ACTION_MATCHER_IN_POLICIES_CONDITION set to false for all other serviceDefs. 

Alternatively, we could make this option dynamic for each serviceDef XML configuration by using `ranger.servicedef.<service-def-hard-coded-name>.enableActionMatcherInPoliciesCondition`. Could you please share your thoughts on whether this approach makes sense ?

ramackri:
> @fmorg-git, please ensure to add the feature flag in dev-support/[ranger-docker](https://issues.apache.org/jira/browse/RANGER-docker) and test it in docker based environments.

Done

ramackri:
> hi all, something appears to be wrong. I just tried my Ozone/Ranger smoke tests with the feature flag set to true, but it is not working now (this used to work before). So something with the changes between the last time I said it was working and now seems to be different. I updated with the new name `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition` but it still doesn't work.
@fmorg-git 
You this issue got fixed   (admin config wins over stale DB),

ramackri:
> hi all, something appears to be wrong. I just tried my Ozone/Ranger smoke tests with the feature flag set to true, but it is not working now (this used to work before). So something with the changes between the last time I said it was working and now seems to be different. I updated with the new name `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition` but it still doesn't work.

Fixed

plugin-ozone/src/main/java/org/apache/ranger/authorization/ozone/authorizer/RangerOzoneAuthorizer.java
ramackri: @fmorg-git  Can you check this ?

fmorg-git: hi @mneethiraj @ramackri - in the code prior to RANGER-5628, it was setting action to `accessType`.  So when the feature flag is off, I believe it should match that prior behavior, otherwise the audit logs will differ when the feature flag is off when compared to the pre-RANGER-5628 behavior.

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
fmorg-git: updated - 3fc1b1508038e6b33d13332aff08374bc8702a02.  I think with this approach, service definitions can opt-in in the future to include the condition

security-admin/src/main/webapp/react-webapp/src/views/PolicyListing/AddUpdatePolicyForm.jsx
fmorg-git: updated - 3b0d85cf3c664d52656161772dea53392b67c093.  I also updated `AccessGrantForm.jsx` as well.

security-admin/src/main/java/org/apache/ranger/patch/PatchForOzoneServiceDefPolicyConditionUpdate_J10065.java
fmorg-git: updated to revert - e220c42e2dc0261848c056ce82b97d2f35d3ddfb.  Also updated the README.md accordingly.

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
fmorg-git: updated - a3757542468bcdf992ad4b65fa28ce10b866268a

security-admin/src/main/java/org/apache/ranger/service/RangerServiceDefService.java
mneethiraj: @fimugdha - using a service-def specific configuration to enable/disable action-matcher condition is a better solution. +1 for this approach.

plugin-ozone/src/main/java/org/apache/ranger/authorization/ozone/authorizer/RangerOzoneAuthorizer.java
mneethiraj: @fmorg-git - ok. Let's retain the updates in this PR.

mneethiraj (commented):
I suggest replacing changes in `dev-support/ranger-docker` with an update in Ranger admin setup processing of `install.properties`, to support configuration
`FF_ENAB_OZONE_ACTION_MATCHES_CONDITION` mapped to `ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition` in the generated `ranger-admin-site.xml`.

fmorg-git:
> I suggest replacing changes in dev-support/ranger-docker with an update in Ranger admin setup processing of install.properties, to support configuration
FF_ENAB_OZONE_ACTION_MATCHES_CONDITION mapped to ranger.servicedef.ozone.enableActionMatcherInPoliciesCondition in the generated ranger-admin-site.xml.

updated - d4343e5a49b7d5e5688ce58d85c37b833993c8df

mneethiraj (approved):

fimugdha (approved):
LGTM. Lets wait to get a review vote from @mneethiraj.

fimugdha:
> LGTM. Lets wait to get a review vote from @mneethiraj.

PR was not refreshed, we already have a vote from him.

fmorg-git:
Thanks all for the review and merge.

PR #1075, opened 2026-07-15
RANGER-5652: Update threat model to v1.1

kumaab:

Add user/group/role lookup by an authenticated user as NON-FINDING

## How was this patch tested?

NA


ramackri (approved):
LGTM

PR #1076, opened 2026-07-15
RANGER-5689: Standardize TLS hostname verification across Ranger HTTP…

vyommani:
… clients using a uniform verifier

used the RangerDefaultHostnameVerifier in all four clients

## How was this patch tested?

added new tests and mvn clean install is successful.


ramackri (approved):
LGTM

PR #1081, opened 2026-07-16
RANGER-5693: Stop logging full JWT bearer tokens on validation failure

ramackri:
## Summary

- Remove logging of the full serialized JWT bearer token (`header.payload.signature`) when `RangerJwtAuthHandler` validation fails.
- Log only non-sensitive metadata: `subject`, `audience`, `issuer`, `keyId`, `jwtId`.

**JIRA:** https://issues.apache.org/jira/browse/RANGER-5693

## Problem

On validation failure, `RangerJwtAuthHandler.authenticate()` logged:

```
RangerJwtAuthHandler.authenticate(): Validation failed for JWT token: [<full serialized JWT>]
```

`validateToken()` checks claims in order: **expiration → signature → audience → issuer**. A cryptographically valid, unexpired token rejected only for **audience mismatch** still hit this WARN path. The full bearer credential in logs could be replayed against another Ranger JWT consumer that trusts the same signing material and accepts the token's audience.

## Fix

**Before:**

```java
LOG.warn("... Validation failed for JWT token: [{}] ", jwtToken.serialize());
```

**After:**

```java
LOG.warn("JWT validation failed ({})", safeJwtLogContext(jwtToken));
```

`safeJwtLogContext()` emits diagnostic fields only, for example:

```
JWT validation failed (subject=f015_replay_user, audience=service-a, issuer=test-issuer, keyId=kid-abc, jwtId=jti-123)
```

The raw bearer string is never written to operational logs. `jwtId` (`jti` claim) provides log correlation when present.

### Null-safety

- `validateAudiences()`: guard `tokenAudienceList != null` before iterating (avoids NPE when `aud` claim is absent).
- `safeJwtLogContext()`: guard `jwtToken.getHeader()` before reading `keyId`; use `StringUtils.join()` for audience list (JDK 8 compatible, tolerates null list entries).

## Files changed (2)

| File | Change |
|------|--------|
| `ranger-authn/.../RangerJwtAuthHandler.java` | Safe logging, `safeJwtLogContext()`, audience null guard |
| `ranger-authn/.../TestRangerJwtAuthHandler.java` | Unit tests for safe context and missing audience |

### Unit tests (automated)

```bash
mvn -pl ranger-authn test -Dtest=TestRangerJwtAuthHandler
mvn verify -pl ranger-authn -Dtest=TestRangerJwtAuthHandler -Drat.skip=true
```

### DIFF
--- a/ranger-authn/src/main/java/org/apache/ranger/authz/handler/jwt/RangerJwtAuthHandler.java
+++ b/ranger-authn/src/main/java/org/apache/ranger/authz/handler/jwt/RangerJwtAuthHandler.java
@@ -40,9 +40,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.text.ParseException;
 import java.util.Arrays;
 import java.util.Date;
@@ -130,7 +127,7 @@ public abstract class RangerJwtAuthHandler implements RangerAuthHandler {
                         }
                         return userName;
                     } else {
-                        LOG.warn("JWT validation failed ({})", safeJwtLogContext(jwtToken, serializedJWT));
+                        LOG.warn("JWT validation failed ({})", safeJwtLogContext(jwtToken));
                     }
                 } catch (ParseException pe) {
                     LOG.warn("RangerJwtAuthHandler.authenticate(): Unable to parse the JWT token", pe);
@@ -313,14 +310,12 @@ public abstract class RangerJwtAuthHandler implements RangerAuthHandler {
 
     /**
      * Build non-sensitive JWT metadata for operational logs.
-     * Never log the raw bearer token; a one-way hash is included for
-     * correlation across log lines.
+     * Never log the raw bearer token.
      *
      * @param jwtToken parsed JWT used to extract claim metadata
-     * @param serializedJwt original bearer string, hashed for correlation only
      * @return safe diagnostic string for log output
      */
-    protected String safeJwtLogContext(final SignedJWT jwtToken, final String serializedJwt) {
+    protected String safeJwtLogContext(final SignedJWT jwtToken) {
         try {
             JWTClaimsSet claims = jwtToken.getJWTClaimsSet();
             JWSHeader header = jwtToken.getHeader();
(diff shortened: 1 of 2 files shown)


ranger-authn/src/main/java/org/apache/ranger/authz/handler/jwt/RangerJwtAuthHandler.java
```diff
@@ -304,6 +311,43 @@ protected boolean validateIssuer(final SignedJWT jwtToken) {
...
+            List<String> tokenAudiences = claims.getAudience();
+            String audience = tokenAudiences == null || tokenAudiences.isEmpty() ? null : StringUtils.join(tokenAudiences, ",");
+
+            return String.format("subject=%s, audience=%s, issuer=%s, keyId=%s, jwtId=%s, tokenHash=%s", claims.getSubject(), audience, claims.getIssuer(), keyId, claims.getJWTID(), sha256Hex(serializedJwt));
+        } catch (ParseException pe) {
+            return "claims_unparseable";
+        }
+    }
+
+    private static String sha256Hex(final String value) {
```
kumaab: What value does `sha256Hex` add here? I suggest dropping sha256/tokenHash entirely.

Agreed it enables correlation, but `jti` (claims.getJWTID()) already covers that whenever it's present, and when it's absent the opaque hash is non-actionable for normal debugging — so it's not worth the added code for this fix.

ramackri: Sure i will remove it

ranger-authn/src/test/java/org/apache/ranger/authz/handler/jwt/TestRangerJwtAuthHandler.java
```diff
@@ -106,4 +116,37 @@ void validateIssuerFalse_whenJwtClaimsCannotBeParsed() throws Exception {
...
+
+        String context = handler.callSafeJwtLogContext(SignedJWT.parse(serialized));
+
+        assertNotNull(context);
+        assertTrue(context.contains("subject=f015_replay_user"));
+        assertTrue(context.contains("audience=service-a"));
+        assertTrue(context.contains("issuer=test-issuer"));
+        assertTrue(context.contains("keyId=kid-abc"));
+        assertTrue(context.contains("jwtId=jti-123"));
+        assertFalse(context.contains("tokenHash="));
```
kumaab: L146 needs to be removed.

ramackri: @kumaab Removed assertFalse(context.contains("tokenHash="));

dhavalshah9131 (approved):

PR #1096, opened 2026-07-20
RANGER-5700: Support SPIFFE-based authn via HTTP headers

kumaab:
- Adds SPIFFE-based service-to-service authentication as a second trusted-header path, alongside the existing username-header (user identity) authentication, in Ranger Admin and the PDP server.
- When the username header is absent, the server reads a configured SPIFFE header and extracts the trailing service-account segment from the SPIFFE ID (spiffe://<trust-domain>/ns/<ns>/sa/<service-account> → <service-account>), authenticating the caller as that service account.
- Adds a shared SpiffeIdUtil helper in common-utils so Admin and PDP share identical parsing semantics.
- New config: 
     - ranger.admin.authn.header.spiffe (Admin)
     - ranger.pdp.authn.header.spiffe (PDP).
- By design the config accepts a comma-separated list of SPIFFE header names while also fully supporting a single header name (a single value is just a list of one).
- Precedence: the username header always wins; among SPIFFE headers, the first one carrying a well-formed SPIFFE ID is used, and malformed values are skipped.
- Backward compatible: the existing *.authn.header.enabled flag continues to gate the entire trusted-header path.

## How was this patch tested?
- Added unit tests for SpiffeIdUtil, the Admin RangerHeaderPreAuthFilter, and the PDP HttpHeaderAuthNHandler.
- Verified end-to-end in the dev-support/ranger-docker stack (Admin + PDP).
- Single-ID scenarios: no header → 401; valid SPIFFE header → authenticated as the service account; malformed SPIFFE ID → 401.
- Multiple-ID scenarios: falls through to a later header when an earlier one is absent; the first well-formed header wins even when a later header would resolve to a different valid identity; a malformed leading header is skipped so resolution continues to the next configured header.

### DIFF
--- a/common-utils/src/main/java/org/apache/ranger/plugin/util/SpiffeIdUtil.java
+++ b/common-utils/src/main/java/org/apache/ranger/plugin/util/SpiffeIdUtil.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.ranger.plugin.util;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Helper for extracting a service-account identity from a SPIFFE ID.
+ *
+ * <p>A SPIFFE ID has the form {@code spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>};
+ * for service-to-service authentication the receiving server trusts the workload
+ * identity carried in a header (e.g. {@code x-service-spiffe-id}) and uses
+ * the trailing service-account segment as the authenticated principal.
+ */
+public final class SpiffeIdUtil {
+    private static final String SPIFFE_SCHEME     = "spiffe://";
+    private static final String HEADER_NAMES_SEP  = ",";
+
+    private SpiffeIdUtil() {
+        // to block instantiation
+    }
+
+    /**
+     * Parses a configured, comma-separated list of SPIFFE header names into an ordered list of
+     * trimmed, non-empty names. Supports a single header name as a list of length one.
+     *
+     * @param configValue the raw config value (e.g. {@code x-awc-source-workload-id, x-other-id})
+     * @return an ordered, immutable list of header names; empty when the input is blank
+     */
+    public static List<String> parseHeaderNames(String configValue) {
+        if (StringUtils.isBlank(configValue)) {
+            return Collections.emptyList();
+        }
+
+        List<String> ret = new ArrayList<>();
+
+        for (String name : configValue.split(HEADER_NAMES_SEP)) {
+            String trimmed = StringUtils.trimToNull(name);
+
+            if (trimmed != null) {
+                ret.add(trimmed);
+            }
+        }
+
+        return Collections.unmodifiableList(ret);
+    }
+
+    /**
+     * Extracts the service-account name (the last non-empty path segment) from a
+     * SPIFFE ID such as {@code spiffe://my-cluster/ns/service-namespace/sa/service-sa}.
+     *
+     * @param spiffeId the raw SPIFFE ID value from the request header
+     * @return the trailing service-account segment (e.g. {@code service-sa}), or
+     *         {@code null} if the input is blank or not a well-formed SPIFFE ID
+     */
+    public static String extractServiceAccount(String spiffeId) {
+        if (StringUtils.isBlank(spiffeId)) {
+            return null;
+        }
+
+        String trimmed = spiffeId.trim();
+
+        if (!trimmed.startsWith(SPIFFE_SCHEME)) {
+            return null;
+        }
+
+        String path = StringUtils.stripEnd(trimmed, "/");
+        int    idx  = path.lastIndexOf('/');
+
+        return idx >= 0 && idx < path.length() - 1 ? path.substring(idx + 1) : null;
+    }
+}
--- a/common-utils/src/main/java/org/apache/ranger/plugin/util/SpiffeIdUtil.java
+++ b/common-utils/src/main/java/org/apache/ranger/plugin/util/SpiffeIdUtil.java
@@ -39,16 +39,20 @@ public final class SpiffeIdUtil {
     private static final String  HEADER_NAMES_SEP  = ",";
 
     /**
-     * Matches {@code spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>}.
-     *
-     * <p>The trust-domain is a DNS-style label set (lowercase alphanumerics, {@code -} and {@code .});
-     * the namespace and service-account are RFC-1123 labels (lowercase alphanumerics and {@code -},
-     * not starting or ending with {@code -}), matching Kubernetes naming. This deliberately excludes
-     * uppercase, whitespace and control characters so the extracted principal is well-formed.
-     * The service-account is exposed via the named group {@code sa}.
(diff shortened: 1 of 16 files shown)


dev-support/ranger-docker/scripts/pdp/ranger-pdp-site.xml
```diff
@@ -115,6 +115,12 @@
     <value>X-Forwarded-User</value>
   </property>
 
+  <!-- service-to-service authentication via SPIFFE ID (comma-separated header names) -->
+  <property>
+    <name>ranger.pdp.authn.header.spiffe</name>
+    <value>x-awc-source-workload-id</value>
```
ramackri: x-awc-source-workload-id
x-awc-username ---> can you change it to some other name

common-utils/src/main/java/org/apache/ranger/plugin/util/SpiffeIdUtil.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Matches {@code spiffe://<trust-domain>/ns/<namespace>/sa/<service-account>}.
+     *
+     * <p>The trust-domain is a DNS-style label set (lowercase alphanumerics, {@code -} and {@code .});
+     * the namespace and service-account are RFC-1123 labels (lowercase alphanumerics and {@code -},
+     * not starting or ending with {@code -}), matching Kubernetes naming. This deliberately excludes
+     * uppercase, whitespace and control characters so the extracted principal is well-formed.
+     * The service-account is exposed via the named group {@code sa}.
+     */
+    private static final Pattern SPIFFE_ID_PATTERN = Pattern.compile(
```
ramackri, line 56: The regex only allows lowercase letters and digits (a-z, 0-9). Any uppercase letter (A-Z) fails validation.

That applies to all three parts of the ID:

trust domain → my-cluster ✅, My-Cluster ❌
namespace → prod ✅, Prod ❌
service account → service-sa ✅, Service-SA ❌
So a SPIFFE ID like spiffe://My-Cluster/ns/Prod/sa/Service-SA is rejected, even though the SPIFFE spec allows mixed case.
Example: this is rejected even though SPIFFE IDs are case-sensitive in the spec:

spiffe://My-Cluster/ns/Prod/sa/Service-SA

kumaab: Good catch, thanks, I've fixed it now.

dev-support/ranger-docker/scripts/pdp/ranger-pdp-site.xml
kumaab: This is only used in dev-support/ranger-docker, so should be fine.

common-utils/src/main/java/org/apache/ranger/plugin/util/SpiffeIdUtil.java
kumaab: Strictly following the spec described here: https://github.com/spiffe/spiffe/blob/main/standards/SPIFFE-ID.md

dev-support/ranger-docker/scripts/pdp/ranger-pdp-site.xml
mneethiraj: @kumaab  - I suggest using a generic header name as `X-Spiffe-Id`

security-admin/src/test/java/org/apache/ranger/security/web/filter/TestRangerHeaderPreAuthFilter.java
```diff
@@ -151,6 +152,183 @@ public void doFilter(ServletRequest req, ServletResponse res) {
         filter.doFilter(request, response, chain);
     }
 
+    @Test
+    public void testDoFilter_enabled_withSpiffeHeader_setsServiceAccountAuthentication() throws Exception {
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_HEADER_AUTH_ENABLED, "true");
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_USERNAME_HEADER_NAME, "x-awc-username");
```
mneethiraj: `x-awc-username` => `X-Forwarded-User`
`x-awc-source-workload-id` => `X-Spiffe-Id`

security-admin/src/test/java/org/apache/ranger/security/web/filter/TestRangerHeaderPreAuthFilter.java
```diff
@@ -151,6 +152,183 @@ public void doFilter(ServletRequest req, ServletResponse res) {
...
+            }
+        };
+
+        filter.doFilter(request, response, chain);
+    }
+
+    @Test
+    public void testDoFilter_enabled_usernameHeaderTakesPrecedenceOverSpiffe() throws Exception {
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_HEADER_AUTH_ENABLED, "true");
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_USERNAME_HEADER_NAME, "x-awc-username");
```
mneethiraj: `x-awc-username` => `X-Forwarded-User`
`x-awc-source-workload-id` => `X-Spiffe-Id`

security-admin/src/test/java/org/apache/ranger/security/web/filter/TestRangerHeaderPreAuthFilter.java
```diff
@@ -151,6 +152,183 @@ public void doFilter(ServletRequest req, ServletResponse res) {
...
+            }
+        };
+
+        filter.doFilter(request, response, chain);
+    }
+
+    @Test
+    public void testDoFilter_enabled_malformedSpiffeHeader_passesThrough() throws Exception {
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_HEADER_AUTH_ENABLED, "true");
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_SPIFFE_HEADER_NAME, "x-awc-source-workload-id");
```
mneethiraj: `x-awc-source-workload-id` => `X-Spiffe-Id`

security-admin/src/test/java/org/apache/ranger/security/web/filter/TestRangerHeaderPreAuthFilter.java
```diff
@@ -151,6 +152,183 @@ public void doFilter(ServletRequest req, ServletResponse res) {
...
+
+        verify(chain).doFilter(request, response);
+        verify(userMgr, never()).getRolesByLoginId(anyString());
+        assertNull(SecurityContextHolder.getContext().getAuthentication());
+    }
+
+    @Test
+    public void testDoFilter_enabled_specValidButNonConformingSpiffeHeader_passesThrough() throws Exception {
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_HEADER_AUTH_ENABLED, "true");
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_SPIFFE_HEADER_NAME, "x-awc-source-workload-id");
```
mneethiraj: `x-awc-source-workload-id` => `X-Spiffe-Id`

security-admin/src/test/java/org/apache/ranger/security/web/filter/TestRangerHeaderPreAuthFilter.java
```diff
@@ -151,6 +152,183 @@ public void doFilter(ServletRequest req, ServletResponse res) {
...
+
+        verify(chain).doFilter(request, response);
+        verify(userMgr, never()).getRolesByLoginId(anyString());
+        assertNull(SecurityContextHolder.getContext().getAuthentication());
+    }
+
+    @Test
+    public void testDoFilter_enabled_spiffeHeaderWithIllegalChars_passesThrough() throws Exception {
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_HEADER_AUTH_ENABLED, "true");
+        PropertiesUtil.getPropertiesMap().put(RangerHeaderPreAuthFilter.PROP_SPIFFE_HEADER_NAME, "x-awc-source-workload-id");
```
mneethiraj: `x-awc-source-workload-id` => `X-Spiffe-Id`

dev-support/ranger-docker/scripts/pdp/ranger-pdp-site.xml
kumaab: Sure, fixed all header names. thanks.

ramackri (approved):

kumaab:
Testing evidence: 
[spiffe-id-test-cases-validation-docker.log](https://github.com/user-attachments/files/30250161/spiffe-id-test-cases-validation-docker.log)

PR #1107, opened 2026-07-24
Convert Tagsync into webapp and add metrics

Sanket-Shelar:
Converted tagsync into webapp, made changes to start tagsync in docker with webapp and added metrics for tagsync

## How was this patch tested?

Local build passed and tested tags are syncing in docker

-----------------------------------------------

[RANGER-4153](https://issues.apache.org/jira/browse/RANGER-4153): Ranger Tagsync - Convert to Web Application
[RANGER-5706](https://issues.apache.org/jira/browse/RANGER-5706): Support for Tagsync as Webapp in Docker
[RANGER-5707](https://issues.apache.org/jira/browse/RANGER-5707): Add/ Update metric details for Ranger TagSync


tagsync/src/main/java/org/apache/ranger/tagsync/server/EmbeddedServer.java
```diff
@@ -0,0 +1,733 @@
...
+import java.security.PrivilegedAction;
+import java.security.SecureRandom;
+import java.security.UnrecoverableKeyException;
+import java.security.cert.CertificateException;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+public class EmbeddedServer {
```
pradeepagrawal8184, line 71: Try to use existing embedded webserver rather copying and introducing one more.

pradeepagrawal8184 (changes requested):

tagsync/src/main/java/org/apache/ranger/tagsync/source/atlasrest/AtlasRESTTagSource.java
```diff
@@ -174,6 +177,7 @@ public void run() {
         LOG.debug("==> AtlasRESTTagSource.run()");
 
         while (true) {
+            StopWatch eventStopWatchRest = new StopWatch().start();
```
pradeepagrawal8184: it seems updateTotalUploadsTime(timeElapsed) runs every loop iteration and includes the full sleep interval, not just upload work. Metric name suggests upload time but measures cycle time.

tagsync/src/main/java/org/apache/ranger/tagsync/model/AbstractTagSource.java
```diff
@@ -70,6 +71,12 @@ protected void updateSink(final ServiceTags toUpload) throws Exception {
                 if (LOG.isDebugEnabled()) {
                     LOG.debug("Uploaded serviceTags={}", JsonUtils.objectToJson(uploaded));
                 }
+                if (!uploaded.getTags().isEmpty()) {
+                    RangerTagSyncMetricsSourceTags.updateTotalUploads();
```
pradeepagrawal8184: TotalUploads increments only when !uploaded.getTags().isEmpty(). Uploads with service resources but no tags are not counted.

tagsync/src/main/java/org/apache/ranger/tagsync/source/atlasrest/AtlasRESTTagSource.java
```diff
@@ -183,6 +187,11 @@ public void run() {
                     LOG.debug("==> This server is running passive mode");
```
pradeepagrawal8184, line 185: is this should be part of else block ?

tagsync/src/main/java/org/apache/ranger/tagsync/source/atlasrest/AtlasRESTTagSource.java
Sanket-Shelar: Removed loop-level StopWatch from run(); updateTotalUploadsTime() now records only the updateSink() duration in synchUp(), and sleep runs in finally outside the timer.

tagsync/src/main/java/org/apache/ranger/tagsync/source/atlasrest/AtlasRESTTagSource.java
Sanket-Shelar: Passive-mode debug log moved to the else branch when the server is not active.

pradeepagrawal8184 (approved):

PR #1112, opened 2026-07-27
RANGER-5678: Concurrent policy-engine rebuild on shared serviceDef causes CME/NPE for delegated-admin User

vyommani:

Under concurrent delegated-admin V2 API load, multiple threads rebuild the policy engine for the same service — or for services that share a tag serviceDef — at the same time. Each rebuild calls ServiceDefUtil.normalize() on the shared, cached RangerServiceDef, which mutates it in place, causing ConcurrentModificationException/NullPointerException under concurrent access.
The fix makes ServiceDefUtil.normalize() return a deep copy of the RangerServiceDef instead of mutating its input.

Initially, I assumed adding copy constructors and changing normalize() to copy-then-mutate would be enough on its own. However, ServiceDefUtil.normalize() was both mutating its input and returning it — which made its original intent ambiguous: was the return value meant to be authoritative, or was it just a convenience alongside the real effect (the in-place mutation)? Rather than guess, I took the conservative approach: treat the return value as authoritative, and update every call site to propagate the returned (now-copied) RangerServiceDef back to wherever the caller had originally expected the mutation to be visible. This keeps the change behavior-preserving — nothing that depended on seeing the normalized state after calling normalize() loses that guarantee — while fully resolving the CME/NPE.

## How was this patch tested?

Existing unit tests updated and passing.
New test testNormalizeDoesNotMutateOrShareStateWithSource verifies identity and content isolation.
Local concurrency exercises (same serviceDef + cross-service shared tag serviceDef) no longer reproduce the CME/NPE

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/model/RangerBaseModelObject.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/model/RangerBaseModelObject.java
@@ -57,6 +57,17 @@ public class RangerBaseModelObject implements java.io.Serializable {
         setIsEnabled(null);
     }
 
+    public RangerBaseModelObject(RangerBaseModelObject other) {
+        setId(other.getId());
+        setGuid(other.getGuid());
+        setIsEnabled(other.getIsEnabled());
+        setCreatedBy(other.getCreatedBy());
+        setUpdatedBy(other.getUpdatedBy());
+        setCreateTime(other.getCreateTime());
+        setUpdateTime(other.getUpdateTime());
+        setVersion(other.getVersion());
+    }
+
     public static <T> List<T> nullSafeList(List<T> coll) {
         return nullSafeSupplier.toList(coll);
     }
--- a/agents-common/src/main/java/org/apache/ranger/plugin/model/RangerServiceDef.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/model/RangerServiceDef.java
@@ -32,6 +32,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.UnaryOperator;
 
 @JsonAutoDetect(fieldVisibility = Visibility.ANY)
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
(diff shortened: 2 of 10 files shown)


vyommani:
@mneethiraj , @pradeepagrawal8184 , can you please review this PR when you get time.

mneethiraj (commented):
@vyommani  - how about adding `normalize()` method in `RangerServiceDef` which will perform the operation in thread-safe way. This will isolate the details within `RangerServiceDef`.

vyommani:
> @vyommani - how about adding `normalize()` method in `RangerServiceDef` which will perform the operation in thread-safe way. This will isolate the details within `RangerServiceDef`.

Good suggestion, and it's true that today's ServiceDefUtil.normalizeDataMaskDef()/normalizeRowFilterDef() are technically callable directly on an un-copied instance, which is a real gap worth closing eventually. One clarification on the "thread-safe" framing: the current fix doesn't need any synchronization at all - normalize() returns a fresh copy with no shared mutable state, so there's nothing left to lock. Moving it onto RangerServiceDef wouldn't change that; it'd still be copy-then-mutate, just called as serviceDef.normalize() instead of ServiceDefUtil.normalize(serviceDef).

We'd rather not fold this into the current PR. Making it worthwhile (not just a thin wrapper) means normalizeDataMaskDef()/normalizeRowFilterDef() becoming private instance methods only reachable through the new normalize(), which means re-touching every call site we just finished auditing and testing , plus a decision on whether normalizeAccessTypeDefs()/getMarkerAccessTypes()/getDataMaskType() move too (leaving them as static utilities would create a half-migrated, inconsistent API). None of that is hard, but it's real scope on a patch that just went green.

We'll file this as a follow-up and take a pass at moving the whole normalize-family onto the model class in one coherent change, rather than partially migrating it here.

mneethiraj:
> > @vyommani - how about adding `normalize()` method in `RangerServiceDef` which will perform the operation in thread-safe way. This will isolate the details within `RangerServiceDef`.
>
... 
> 
> We'll file this as a follow-up and take a pass at moving the whole normalize-family onto the model class in one coherent change, rather than partially migrating it here.

@vyommani  - instead of taking this fix and then subsequent patch to move normalization to RangerServiceDef, why not skip the current fix and directly go with the final/desired fix?

vyommani:
> @vyommani - how about adding `normalize()` method in `RangerServiceDef` which will perform the operation in thread-safe way. This will isolate the details within `RangerServiceDef`.

Good suggestion, and it's true that today's ServiceDefUtil.normalizeDataMaskDef()/normalizeRowFilterDef() are technically callable directly on an un-copied instance, which is a real gap worth closing eventually. One clarification on the "thread-safe" framing: the current fix doesn't need any synchronization at all - normalize() returns a fresh copy with no shared mutable state, so there's nothing left to lock. Moving it onto RangerServiceDef wouldn't change that; it'd still be copy-then-mutate, just called as serviceDef.normalize() instead of ServiceDefUtil.normalize(serviceDef).

We'd rather not fold this into the current PR. Making it worthwhile (not just a thin wrapper) means normalizeDataMaskDef()/normalizeRowFilterDef() becoming private instance methods only reachable through the new normalize(), which means re-touching every call site we just finished auditing and testing , plus a decision on whether normalizeAccessTypeDefs()/getMarkerAccessTypes()/getDataMaskType() move too (leaving them as static utilities would create a half-migrated, inconsistent API). None of that is hard, but it's real scope on a patch that just went green.

We'll file this as a follow-up and take a pass at moving the whole normalize-family onto the model class in one coherent change, rather than partially migrating it here.

> @vyommani - instead of taking this fix and then subsequent patch to move normalization to RangerServiceDef, why not skip the current fix and directly go with the final/desired fix?

Update on the normalize()-as-instance-method idea - started the migration on a separate branch and found something worth flagging before we go further.

normalizeDataMaskDef()/normalizeRowFilterDef() call ServiceDefUtil.mergeResourceDef()/mergeAccessTypeDef(), which are private static on ServiceDefUtil. These aren't trivial - they're real merge logic (field-by-field precedence rules, a couple of special-cased conditions, manual per-key map merging). Moving the normalize family onto RangerServiceDef means either making these private helpers public (leaking internal merge logic as permanent public API for one caller) or cascading the move further into RangerServiceDef. Neither is great  RangerServiceDef is a plain data model everywhere else in the codebase, and this would be the first real business logic living on it.

Given that design cost, I don't think the full migration is worth pursuing as originally scoped. If we still want the nicer call syntax, a thin delegating wrapper gets it for free with none of the downside:

public RangerServiceDef normalize() {
    return ServiceDefUtil.normalize(this);
}

Everything else - normalizeDataMaskDef, normalizeRowFilterDef, mergeResourceDef, mergeAccessTypeDef - stays exactly where it is, exactly as private as it is today. Zero cascade, zero visibility changes, zero risk.

Open to your read on whether even the thin wrapper is worth adding, or whether we just leave ServiceDefUtil.normalize() as the one true entry point and close this out.

mneethiraj:
@vyommani  - I suggest considering the approach in [RANGER-5678.patch](https://github.com/user-attachments/files/30689786/RANGER-5678.patch)

PR #1120, opened 2026-07-31
RANGER-5717: Introduce health check REST APIs in Ranger Admin

kumaab:

Adds two new REST APIs:
- `/service/actuator/health/readiness`
- `/service/actuator/health/liveness`

readiness check runs with`healthcheck` user only , the check by-passes db updates while maintaining a session.

## How was this patch tested?

- Tested using `header based authentication` in docker with `multiple cURL calls`: 
  ```
  curl -X GET \
  'http://localhost:6080/service/actuator/health/readiness' \
  -H 'Accept: application/json' \
  -H 'X-Forwarded-User: healthcheck'

  {"status":"UP","details":{"components":{"service-defs": ["atlas","elasticsearch","gds","hbase","hdfs","hive","kafka","kms","knox","kudu","kylin","nestedstructure","nifi","nifi-registry","ozone","polaris","presto","schema-registry","solr","sqoop","storm","tag","trino","yarn"]}}}%


  curl -X GET \
  'http://localhost:6080/service/actuator/health/readiness' \
  -H 'Accept: application/json' \
  -H 'X-Forwarded-User: healthcheck1'

  {"statusCode":403,"msgDesc":"Only the healthcheck user may query service-def names via this path."}%
   ```
- Records are seen in access logs but not in 'Login Sessions' in Ranger UI -> Audits tab
image


security-admin/src/main/java/org/apache/ranger/rest/RangerHealthREST.java
```diff
@@ -54,4 +62,26 @@ public RangerServerHealth getRangerServerHealth() {
 
         return rangerServerHealthUtil.getRangerServerHealth(dbVersion);
     }
+
+    @GET
+    @Path("/health/readiness")
+    @Produces("application/json")
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public RangerServerHealth getRangerServerReadiness(@Context HttpServletRequest request) {
+        List<RangerServiceDef> serviceDefs = serviceREST.getServiceDefs(request).getServiceDefs();
```
mneethiraj: Instead of retrieving `RangerServiceDef` objects from the database, I suggest to retrieve only service-def names from the database using a new JPA query.

Also update `RangerServerHealthUtil.serviceUpWithAvailableServiceDefs()` to return the name list instead of sertializing entire service-defs.

security-admin/db/postgres/optimized/current/ranger_core_db_postgres.sql
```diff
@@ -1997,6 +1997,11 @@ INSERT INTO x_portal_user(CREATE_TIME,UPDATE_TIME,FIRST_NAME,LAST_NAME,PUB_SCR_N
 INSERT INTO x_portal_user_role(CREATE_TIME,UPDATE_TIME,USER_ID,USER_ROLE,STATUS)VALUES(current_timestamp,current_timestamp,getXportalUIdByLoginId('rangertagsync'),'ROLE_SYS_ADMIN',1);
 INSERT INTO x_user(CREATE_TIME,UPDATE_TIME,user_name,status,descr)VALUES(current_timestamp,current_timestamp,'rangertagsync',0,'rangertagsync');
 
+-- Built-in health-check user for the /service/actuator/health/readiness endpoint. It authenticates via trusted header (Istio/trusted proxy) only; no usable password is set, so password login is effectively disabled. It is granted read access only.
```
mneethiraj: "It authenticates via trusted header (Istio/trusted proxy) only;" => this is not correct. The user can still authenticate via Kerberos or JWT. I suggest updating the comment to only retain the first sentence.

security-admin/db/postgres/optimized/current/ranger_core_db_postgres.sql
```diff
@@ -1997,6 +1997,11 @@ INSERT INTO x_portal_user(CREATE_TIME,UPDATE_TIME,FIRST_NAME,LAST_NAME,PUB_SCR_N
...
+INSERT INTO x_portal_user(CREATE_TIME,UPDATE_TIME,FIRST_NAME,LAST_NAME,PUB_SCR_NAME,LOGIN_ID,PASSWORD,EMAIL,STATUS)VALUES(current_timestamp,current_timestamp,'healthcheck','','healthcheck','healthcheck','*disabled-header-auth-only*','healthcheck',1);
```
mneethiraj: Database scripts for other flavors (example: MySQL, Oracle) should be updated as well.

ramackri: Do we need to consider for upgrade ?

mneethiraj: @kumaab - why is it necessary to add an entry for `healthcheck` user in the database? If authentication is to be handled by header or kerberos, there shouldn't be any need.

kumaab: Yes, I thought that too initially. `SessionMgr.processSuccessLogin()` still requires a portal user (`x_portal_user` table ) with roles and module permissions. 
Readiness calls `getServiceDefNames()`, which is protected by @PreAuthorize and `hasModuleAccess()`. Without a DB user, `processSuccessLogin()` returns null and readiness fails. 
The `healthcheck` user is a minimal built-in account for probe identity only — not for password login — consistent with other system users in ranger_core_db_*.sql.

security-admin/db/postgres/optimized/current/ranger_core_db_postgres.sql
kumaab: Sure, updated!

security-admin/src/main/java/org/apache/ranger/rest/RangerHealthREST.java
kumaab: Yes, I realized later as well the JSON response is quite heavy and not really required. Updated.

security-admin/db/postgres/optimized/current/ranger_core_db_postgres.sql
ramackri: Readiness fetches all service-def names every time
Readiness calls getServiceDefNames() →:

SELECT obj.name FROM XXServiceDef obj ORDER BY obj.name
Readiness only needs “any service def exists?” — not the full sorted list returned in the JSON body.

Impact: Unnecessary DB work + larger response on every probe (30+ service types in many deployments).

Fix: Use SELECT COUNT(*) or SELECT 1 ... LIMIT 1. Return { "status": "UP" } only — omit the full service-defs list from probe responses.

ramackri: Readiness reuses heavy ServiceREST authorization
getServiceDefNames() is wired like a normal API:

@PreAuthorize(... GET_SERVICE_DEFS ...)
hasModuleAccess(MODULE_RESOURCE_BASED_POLICIES)
refreshPermissionsIfNeeded() inside isAPIAccessible() (can re-hit DB on interval)
Impact: Health probe pays the same authorization cost as GET /service/definitions.

Fix: Use a dedicated DAO method on the readiness path without @PreAuthorize / module checks (auth already handled at the edge, or use unauthenticated internal check).

security-admin/src/main/java/org/apache/ranger/rest/RangerHealthREST.java
```diff
@@ -54,4 +59,26 @@ public RangerServerHealth getRangerServerHealth() {
...
+    @Path("/health/readiness")
+    @Produces("application/json")
+    @Transactional(propagation = Propagation.NOT_SUPPORTED)
+    public RangerServerHealth getRangerServerReadiness() {
+        List<String> serviceDefNames = serviceREST.getServiceDefNames();
+
+        if (serviceDefNames != null && !serviceDefNames.isEmpty()) {
+            return rangerServerHealthUtil.serviceUpWithAvailableServiceDefs(serviceDefNames);
+        } else {
+            return rangerServerHealthUtil.serviceDown();
```
mneethiraj: "down" doesn't seem appropriate here; consider "initializing" or "initialization_failure".

security-admin/src/main/java/org/apache/ranger/biz/SessionMgr.java
```diff
@@ -507,10 +507,15 @@ protected boolean validateUserSession(UserSessionBase userSession, String curren
 
     @Transactional(readOnly = false, propagation = Propagation.REQUIRES_NEW)
     protected XXAuthSession storeAuthSession(XXAuthSession gjAuthSession) {
-        // daoManager.getEntityManager().getTransaction().begin();
+        /*
+        Recording an x_auth_sess row for every liveness/readiness probe is not required
+         */
+        if (gjAuthSession != null && RangerConstants.HEALTH_CHECK_USERNAME.equalsIgnoreCase(gjAuthSession.getLoginId())) {
```
mneethiraj: Replace repeated references to `RangerConstants.HEALTH_CHECK_USERNAME.equalsIgnoreCase` with a method like `RangerBizUtil.isHealthCheckUser(`.

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -607,6 +610,46 @@ public RangerServiceDefList getServiceDefs(@Context HttpServletRequest request)
         return ret;
     }
 
+    public List<String> getServiceDefNames() {
+        LOG.debug("==> ServiceREST.getServiceDefNames()");
+
+        if (!RangerConstants.HEALTH_CHECK_USERNAME.equalsIgnoreCase(resolveAuthenticatedLoginId())) {
+            throw restErrorUtil.createRESTException(HttpServletResponse.SC_FORBIDDEN,
```
mneethiraj: If `ServiceREEST.getServiceDefNames()` meant to be called only from `RangerHealthREST`, why not directly have this implementation in `RangerHealthREST`, and eliminate the updates in ServiceREST, RangerServiceDefService and related classes?

kumaab: Sure, updated.

security-admin/src/main/java/org/apache/ranger/util/RangerServerHealthUtil.java
```diff
@@ -98,4 +136,44 @@ private Map<String, Object> getDbStatus(String dbVersion) {
...
+
+        return ret;
+    }
+
+    /**
+     * Header-based auth sets {@link RangerAuthenticationToken} before Ranger
+     * {@link UserSessionBase} is available. Fall back to that token when the Ranger session
+     * has not been materialized yet (e.g. healthcheck user not in DB).
+     */
+    private String resolveAuthenticatedLoginId() {
```
mneethiraj: To be consistent with rest of Ranger code base, I suggest having a single return statement in a method:

```
if (loginId == null) {
  Object authentication = SecurityContextHolder.getContext().getAuthentication();

  if (authentication instanceof RangerAuthenticationToken) {
    RangerAuthenticationToken token = (RangerAuthenticationToken) authentication;

    if (token.getAuthType() == XXAuthSession.AUTH_TYPE_TRUSTED_PROXY) {
      loginId = token.getName();;
    }
  }
}

return loginId;
```

security-admin/src/main/java/org/apache/ranger/util/RangerServerHealthUtil.java
```diff
@@ -98,4 +136,44 @@ private Map<String, Object> getDbStatus(String dbVersion) {
 
         return ret;
     }
+
+    public List<String> getServiceDefNames() {
+        LOG.debug("==> RangerServerHealthUtil.getServiceDefNames()");
+
+        if (!bizUtil.isHealthCheckUser(resolveAuthenticatedLoginId())) {
```
mneethiraj: Is this `if` necessary? I suggest removing this.

security-admin/src/main/java/org/apache/ranger/common/RangerConstants.java
```diff
@@ -46,6 +46,9 @@ public class RangerConstants extends RangerCommonEnums {
     public static final String RANGER_ADMIN_SUPER_USERS  = "ranger.admin.super.users";
     public static final String RANGER_ADMIN_SUPER_GROUPS = "ranger.admin.super.groups";
 
+    // login id used by the Ranger Admin health readiness endpoint
+    public static final String HEALTH_CHECK_USERNAME = "healthcheck";
```
mneethiraj: I suggest reading the name of healthcheck user from configuration - perhaps in RangerBizUtil.

```
public static final String HEALTHCHECK_USERNAME = RangerAdminConfig.getInstance().get("ranger.admin.healthcheck.username", "healthcheck");
```

security-admin/src/main/java/org/apache/ranger/util/RangerServerHealthUtil.java
kumaab: Yes, this restricts any other user (except `healthcheck` user) from querying this endpoint. (added an example in PR description for `healthcheck1` user).

security-admin/src/main/java/org/apache/ranger/util/RangerServerHealthUtil.java
kumaab: I see, added a cursor rule as well.

security-admin/src/main/java/org/apache/ranger/common/RangerConstants.java
kumaab: done

security-admin/src/main/java/org/apache/ranger/util/RangerServerHealthUtil.java
mneethiraj: ok. A better place for this validation is `RangerHealthREST.getRangerServerReadiness()`.

mneethiraj (approved):

PR #1123, opened 2026-08-03
RANGER-5724: Grant API returns 200 OK for null payload

Sanket-Shelar:
With null payload grant Apis were returning 200 OK status, added null check in Api

## How was this patch tested?
Local build passed, tested on docker and added new unit tests

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
@@ -1149,48 +1149,170 @@ public class ServiceREST {
     public RESTResponse grantAccess(@PathParam("serviceName") String serviceName, GrantRevokeRequest grantRequest, @Context HttpServletRequest request) throws Exception {
         LOG.debug("==> ServiceREST.grantAccess({}, {})", serviceName, grantRequest);
 
+        if (grantRequest == null) {
+            throw restErrorUtil.createRESTException(HttpServletResponse.SC_BAD_REQUEST, "Grant request object is null or missing body in grant access api", false);
+        }
+
         RESTResponse     ret  = new RESTResponse();
         RangerPerfTracer perf = null;
 
-        if (grantRequest != null) {
-            if (serviceUtil.isValidateHttpsAuthentication(serviceName, request)) {
-                try {
-                    bizUtil.failUnauthenticatedIfNotAllowed();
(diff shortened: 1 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
mneethiraj: `if (grantRequest != null) {` can be removed since this condition will not be true due to the `if` added above.

mneethiraj: `if (grantRequest != null) {` can be removed since this condition will not be true due to the `if` added above.

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -1149,6 +1149,10 @@ public List<String> lookupResource(@PathParam("serviceName") String serviceName,
     public RESTResponse grantAccess(@PathParam("serviceName") String serviceName, GrantRevokeRequest grantRequest, @Context HttpServletRequest request) throws Exception {
         LOG.debug("==> ServiceREST.grantAccess({}, {})", serviceName, grantRequest);
 
+        if (grantRequest == null) {
+            throw restErrorUtil.createRESTException(HttpServletResponse.SC_BAD_REQUEST, "Grant request object is null or missing in grant access api", false);
```
mneethiraj: "null or missing body"

security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -1271,6 +1275,10 @@ public RESTResponse grantAccess(@PathParam("serviceName") String serviceName, Gr
     public RESTResponse secureGrantAccess(@PathParam("serviceName") String serviceName, GrantRevokeRequest grantRequest, @Context HttpServletRequest request) throws Exception {
         LOG.debug("==> ServiceREST.secureGrantAccess({}, {})", serviceName, grantRequest);
 
+        if (grantRequest == null) {
+            throw restErrorUtil.createRESTException(HttpServletResponse.SC_BAD_REQUEST, "Grant request object is null or missing in grant access api", false);
```
mneethiraj: "null or missing body"

mneethiraj (approved):

PR #1125, opened 2026-08-04
RANGER-5729: Unable to import exported GDS policies via Service Manag…

Sanket-Shelar:
…er due to 403 Error

Governed Data Sharing (GDS) policies are designed to be managed exclusively through GDS APIs. However, they were inadvertently being included when exporting policies from the Report page. When users attempted to import these files back through the Service Manager, it resulted in a 403 error.

This patch resolves the issue by aligning the export and import behaviors:

**Export Updates**: GDS policies are now automatically excluded when exporting policies (JSON, CSV, or Excel).

**Import Updates**: If a user imports a legacy file that still contains GDS policies, the system will now safely skip the GDS policies (logging a helpful warning) rather than failing the entire import. This ensures that non-GDS policies, like Hive, continue to import successfully.

## How was this patch tested?

Created both GDS objects (datashares, datasets, policies) and non-GDS policies (Hive).

Exported policies from the Report page and verified that GDS policies are no longer included in the file.

Imported the newly exported file via Service Manager and confirmed it succeeds without a 403 error.

Imported a legacy export file containing GDS policies to verify that non-GDS policies import correctly while GDS policies are cleanly skipped and logged.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
@@ -3488,6 +3488,12 @@ public class ServiceREST {
                 RangerPolicy policy = entry.getValue();
 
                 if (policy != null) {
+                    if (isGdsPolicy(policy)) {
+                        LOG.warn("Skipping GDS policy '{}' during import - GDS policies must be managed via GDS APIs", policy.getName());
+
+                        continue;
+                    }
+
                     if (!CollectionUtils.isEmpty(serviceNameList)) {
                         for (String service : serviceNameList) {
                             if (StringUtils.isNotEmpty(service.trim()) && StringUtils.isNotEmpty(policy.getService().trim())) {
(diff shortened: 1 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/rest/ServiceREST.java
```diff
@@ -3833,24 +3839,26 @@ private List<RangerPolicy> getAllFilteredPolicyList(SearchFilter filter, HttpSer
...
-            }
+            policyLists.clear();
+            policyLists.addAll(orderedPolicies.values());
         }
 
         return policyLists;
     }
 
+    private boolean isGdsPolicy(RangerPolicy policy) {
+        return EMBEDDED_SERVICEDEF_GDS_NAME.equals(policy.getServiceType()) || ServiceDBStore.GDS_SERVICE_NAME.equals(policy.getService());
```
mneethiraj: Must this condition look for service name match - `ServiceDBStore.GDS_SERVICE_NAME.equals(policy.getService())`? Why not only look for serviceType match?

Sanket-Shelar: It was added as an additional check, in case serviceType is not present. I have udpated the condition to only check for serviceType match.

mneethiraj (approved):

PR #1137, opened 2026-08-06
RANGER-5719: Shared partition-plan library

ramackri:
## Summary

First PR in the [RANGER-5655](https://issues.apache.org/jira/browse/RANGER-5655) series — Admin-managed Kafka audit partition plan.

**JIRA:** [RANGER-5719](https://issues.apache.org/jira/browse/RANGER-5719)

###  Partition plan model

JSON representation of the audit topic name, plan version, per-plugin partition assignments, buffer pool, and per-service allow-list metadata.
Plan lifecycle helpers

Allocation and validation when services are onboarded, offboarded, or reconfigured.
Append-only partition growth, buffer promotion, and version increments.
Routing utilities

Mapping from Admin-assigned logical partition IDs to Kafka producer partition indices, including safe behavior when live topic metadata differs from the plan.
Allow-list normalization

Parsing and normalization of per-repo allow-list values stored in the plan.
Allow-list usernames are sourced from service.admin.users (not policy.download.auth.users).
Admin configuration helpers

Reading partition-count defaults and per-plugin overrides used when building or expanding a plan.

### Partition plan library (`agents-common`)

Canonical JSON model stored in Admin `x_audit_config` (`audit.partition.plan`):

| Component | Purpose |
|-----------|---------|
| `PartitionPlan`, `PluginEntry`, `BufferEntry` | JSON model: topic, version, plugin → partition map, buffer slots |
| `PartitionPlanAllocator` | Assigns Kafka partition numbers when plugins are promoted from buffer |
| `PartitionPlanValidator` | Validates plan shape, version, and partition ranges |
| `PartitionPlanRoutingUtils` | Resolves plugin ID → Kafka partition at ingestor runtime |
| `ServiceAllowedUsersUtil` | Normalizes per-repo allow-lists in the partition plan JSON (`service.admin.users` values populated by Admin in RANGER-5721) |
| `AuditPartitionPlanAdminConfig` | Reads Admin partition-count settings (`partitions.per.plugin`, per-plugin overrides) |
| `AuditPartitionPlanConstants` | Shared defaults (`ranger_audits` topic, initial version, property names) |

## Test plan

- [x] `mvn -pl agents-common -Dtest='org.apache.ranger.audit.partition.**' test` — partition plan unit tests pass
- [ ] CI: `build-17`, `plugins-docker-build`, `services-docker-build` (after push)

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanAllocator.java
+++ b/agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanAllocator.java
@@ -152,7 +152,7 @@ public final class PartitionPlanAllocator {
             throw new PartitionPlanException("updatedBy is required");
         }
 
-        Map<String, List<String>> normalized = PolicyDownloadAuthUsersUtil.normalizeServiceAllowedUsers(serviceAllowedUsers);
+        Map<String, List<String>> normalized = ServiceAllowedUsersUtil.normalizeServiceAllowedUsers(serviceAllowedUsers);
         if (Objects.equals(current.getServiceAllowedUsers(), normalized)) {
             return current;
         }
--- a/agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanAllocator.java
+++ b/agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanAllocator.java
@@ -1,0 +1,297 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.ranger.audit.partition;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.ranger.audit.partition.exception.PartitionPlanException;
+import org.apache.ranger.audit.partition.model.BufferEntry;
+import org.apache.ranger.audit.partition.model.PartitionPlan;
+import org.apache.ranger.audit.partition.model.PluginEntry;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+/** Append-only plan updates for Admin-managed audit partition routing. */
+public final class PartitionPlanAllocator {
+    private PartitionPlanAllocator() {
+    }
+
+    /**
+     * Onboard a Ranger service repo under a plugin type. Promotes the plugin from buffer when needed,
+     * otherwise adds the service to an existing plugin entry.
+     */
+    public static PartitionPlan onboardService(PartitionPlan current, String pluginId, String serviceName, int partitionCount, String updatedBy) {
+        requireMutationInputs(current, pluginId, partitionCount, updatedBy);
+        if (StringUtils.isBlank(serviceName)) {
+            throw new PartitionPlanException("serviceName is required");
+        }
+        String trimmedService = serviceName.trim();
+        PluginEntry existing = current.getPlugins().get(pluginId);
+        if (existing != null) {
+            return addServiceToPlugin(current, pluginId, trimmedService, updatedBy);
+        }
+        return promotePlugin(current, pluginId, partitionCount, updatedBy, trimmedService);
+    }
+
+    /** Adds a service repo to an already-promoted plugin without changing partition assignment. */
+    public static PartitionPlan addServiceToPlugin(PartitionPlan current, String pluginId, String serviceName, String updatedBy) {
+        if (current == null) {
+            throw new PartitionPlanException("Current plan is required");
+        }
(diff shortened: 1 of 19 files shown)


agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanValidator.java
```diff
@@ -0,0 +1,144 @@
...
+
+        Set<Integer> assigned = new HashSet<>();
+        registerPartitions(plan.getBuffer().getPartitions(), assigned, true);
+        for (Map.Entry<String, PluginEntry> entry : plan.getPlugins().entrySet()) {
+            if (StringUtils.isBlank(entry.getKey())) {
+                throw new PartitionPlanException("Plugin id is required");
+            }
+            registerPartitions(entry.getValue().getPartitions(), assigned, false);
+        }
+        if (assigned.size() != plan.getTopicPartitionCount()) {
```
rameeshm, line 66: If topicPartitionCount = 3 and partition array [2, 4, 6] has be injected into partition plan then we may miss a partition in the hash map. Please check this out.

agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanAllocator.java
```diff
@@ -0,0 +1,257 @@
...
+
+    private static List<Integer> takeFromBuffer(List<Integer> bufferIds, int count) {
+        List<Integer> taken = new ArrayList<>(Math.min(count, bufferIds.size()));
+        while (taken.size() < count && !bufferIds.isEmpty()) {
+            taken.add(bufferIds.remove(0));
+        }
+        return taken;
+    }
+
+    private static int appendTailPartitions(List<Integer> target, int topicPartitionCount, int count, Set<Integer> assigned) {
```
rameeshm: Can we simplify this like the following as it is strictly append-only and partition IDs are a continuous 1..topicPartitionCount
private static int appendTailPartitions(List<Integer> target, int topicPartitionCount, int count) {
    int nextId = topicPartitionCount + 1;
    for (int i = 0; i < count; i++) {
        target.add(nextId++);
    }
    return topicPartitionCount + count;
}

agents-common/src/main/java/org/apache/ranger/audit/partition/PolicyDownloadAuthUsersUtil.java
```diff
@@ -0,0 +1,74 @@
...
+    public static Map<String, List<String>> normalizeServiceAllowedUsers(Map<String, List<String>> serviceAllowedUsers) {
+        if (serviceAllowedUsers == null || serviceAllowedUsers.isEmpty()) {
+            return Collections.emptyMap();
+        }
+        Map<String, List<String>> normalized = new LinkedHashMap<>();
+        for (Map.Entry<String, List<String>> entry : serviceAllowedUsers.entrySet()) {
+            if (StringUtils.isBlank(entry.getKey())) {
+                continue;
+            }
+            List<String> users = entry.getValue() == null ? Collections.emptyList() : parseUsers(String.join(",", entry.getValue()));
```
rameeshm: can be done like 
List<String> users = entry.getValue() == null ? Collections.emptyList() : 
    entry.getValue().stream()
         .flatMap(s -> parseUsers(s).stream())
         .collect(Collectors.toList());

agents-common/src/main/java/org/apache/ranger/audit/partition/PartitionPlanValidator.java
```diff
@@ -0,0 +1,144 @@
...
+            List<Integer> before = entry.getValue().getPartitions();
+            PluginEntry afterEntry = proposed.getPlugins().get(pluginId);
+            if (afterEntry == null) {
+                throw new PartitionPlanException("Append-only violation for plugin '" + pluginId + "'");
+            }
+            List<Integer> after = afterEntry.getPartitions();
+            if (after.size() < before.size()) {
+                throw new PartitionPlanException("Append-only violation for plugin '" + pluginId + "'");
+            }
+            for (int i = 0; i < before.size(); i++) {
```
rameeshm: Can this for loop and check be replaced by 
if (!after.subList(0, before.size()).equals(before)) {
    throw new PartitionPlanException("Append-only violation for plugin '" + pluginId + "': existing partitions reshuffled");
}

agents-common/src/main/java/org/apache/ranger/audit/partition/PolicyDownloadAuthUsersUtil.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/** Parses {@code policy.download.auth.users} service config for audit ingestor authorization. */
+public final class PolicyDownloadAuthUsersUtil {
+    public static final String CONFIG_NAME = "policy.download.auth.users";
```
rameeshm: Do we need to rely only on "policy.download.auth.users"?  I feel we should have a new param "service.access.audit.user" which will be a keytab user (by default service user can be there e.g hive, rangeraz, hdfs, etc) and if its not there then check for  Service admin and if no there then look for "policy.download.auth.users". What do you say? @mneethiraj need your input on this. Thanks.

mneethiraj: @ramackri  - if this is to determine the users authorized to report audits for a service, we should only recognize `service.admin.users` and `service.admin.groups`.  Picking up users from`policy.download.auth.users` is not correct.

agents-common/src/main/java/org/apache/ranger/audit/partition/ServiceAllowedUsersUtil.java
```diff
@@ -0,0 +1,67 @@
...
+                .filter(StringUtils::isNotBlank)
+                .filter(user -> !"*".equals(user))
+                .collect(Collectors.toList());
+    }
+
+    public static Map<String, List<String>> normalizeServiceAllowedUsers(Map<String, List<String>> serviceAllowedUsers) {
+        if (serviceAllowedUsers == null || serviceAllowedUsers.isEmpty()) {
+            return Collections.emptyMap();
+        }
+        Map<String, List<String>> normalized = new LinkedHashMap<>();
```
rameeshm, line 53: Its better to use a LinkedHashSet to avoid duplicate entries here? Just a corner case

PR #1142, opened 2026-08-06
Ranger 5730: Restrict role import createNonExist flag to roles only

eoinmcdonnell113:

RANGER-5730: Restrict createNonExistUserGroupRole during role import so it creates missing nested roles only, and does not create missing users or groups.

Previously, importRolesFromFile with createNonExistUserGroupRole=true treated users, groups, and nested roles the same way. That forced creation of missing users/groups during role import, which is undesirable for large synced environments.

This change:

Splits create-missing behavior in RoleRefUpdater into separate controls for users/groups vs nested roles
Updates RoleREST.importRolesFromFile so createNonExistUserGroupRole=true means create missing nested roles only (createNonExistUserGroup=false)
Leaves existing create/update role API behavior (createNonExistUserGroup) unchanged

## How was this patch tested?

Unit tests:

TestRoleRefUpdater — verifies missing users are not created when only role creation is enabled, and missing nested roles are scheduled for creation
TestRoleREST — verifies import passes (createNonExistUserGroup=false, createNonExistRole=true) when the import flag is true
TestRoleDBStore — existing create/update role coverage
Focused run result: TestRoleRefUpdater, TestRoleDBStore, and TestRoleREST — 102 tests passed (0 failures)

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/store/RoleStore.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/store/RoleStore.java
@@ -40,6 +40,26 @@ public interface RoleStore {
         return updateRole(role, createNonExistUserGroup);
     }
 
+    /**
+     * Create role with separate controls for creating missing users/groups vs nested roles.
+     * Default implementation combines the flags for stores that do not support the split.
+     */
+    default RangerRole createRole(RangerRole role, Boolean createNonExistUserGroup, Boolean createNonExistRole, Boolean isRefTableCleanupRequired) throws Exception {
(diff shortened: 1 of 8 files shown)


ramackri:
@mneethiraj  / @pradeepagrawal8184 
Import with createNonExistUserGroupRole=true used to auto-create missing users/groups referenced in the JSON.
After this PR, that workflow breaks by design. Anyone who relied on import to seed users/groups from a role export file must instead:
Sync/create users & groups first (UserSync / UI / API), then import roles, or
Use POST /roles / PUT /roles/{id} with createNonExistUserGroup=true for individual role create/update (not bulk import)
Is this okay ?

pradeepagrawal8184 (approved):

PR #1143, opened 2026-08-10
RANGER-5728: Update stress_kms script to run against Keberos enabled .

vikaskr22:
Existing stress_kms.py utility file is being updated to run against Kerberos enabled server. Now Ranger services run with Kerberos by default in Docker, so updated this script to run with Kerberos.

This script can still run with SIMPLE AUTH .

This script has been very helpful to quickly verify any sanity/tests, updating this to run with Kerberos will be helpful.

## How was this patch tested?

- mvn build with UT passed
- By running the script against Kerberos enabled server.

### DIFF
--- a/ranger-tools/src/main/python/stress/stress_kms.py
+++ b/ranger-tools/src/main/python/stress/stress_kms.py
@@ -29,27 +29,56 @@ import logging
 # This script requires Python package apache_ranger to be present.
 # The package can be installed using following command:
 #    pip3 install --upgrade apache_ranger
-#
+# If KMS is running with Kerberos, then 'requests-kerberos' package is required and can be installed using following command:
+#    pip3 install requests-kerberos
 
 ##
 ## This script calls KMS APIs from multiple-threads
 ##
 ## parameters:
 ##  kms_url:             URL to Apache Ranger admin server
-##  kms_auth:            authentication to connect to Apache Ranger admin server
+##  kms_auth_type:       authentication mode used to connect to Ranger KMS
+##                         'simple'   - Hadoop simple auth; identity is taken
+##                                      from `kms_user`.
+##                         'kerberos' - SPNEGO/Kerberos auth; identity is taken
+##                                      from the caller's Kerberos ticket cache
+##                                      (run `kinit` before starting the script).
+##                                      Requires the `requests-kerberos` package.
+##  kms_user:            user name to pass when kms_auth_type == 'simple'.
+##                         Ignored in kerberos mode.
 ##  key_count            number of keys to create
 ##  rollover_key_count:  number of times each key to be rolled over
 ##  encrypted_key_count: number of encrypted keys to generate per key
 ##  thread_count:        number of threads to call Apache Ranger APIs from
 ##
-kms_url             = 'http://localhost:9292'
-kms_auth            = HadoopSimpleAuth('keyadmin')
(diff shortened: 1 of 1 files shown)


ranger-tools/src/main/python/stress/stress_kms.py
```diff
@@ -29,21 +29,37 @@
...
+##  kms_user:            user name to pass when kms_auth_type == 'simple'.
+##                         Ignored in kerberos mode.
 ##  key_count            number of keys to create
 ##  rollover_key_count:  number of times each key to be rolled over
 ##  encrypted_key_count: number of encrypted keys to generate per key
 ##  thread_count:        number of threads to call Apache Ranger APIs from
 ##
-kms_url             = 'http://localhost:9292'
-kms_auth            = HadoopSimpleAuth('keyadmin')
+kms_auth_type       = 'kerberos'          # 'simple' | 'kerberos'
```
mneethiraj: Is use of `kms_auth_type` necessary? To keep this script simple, I suggest directly setting `kms_auth` in line 81 below: 

```
# use Hadoop simple authentication
kms_auth = HadoopSimpleAuth(kms_user)

# uncomment following 2 lines to use Kerberos authentication
# from requests_kerberos import HTTPKerberosAuth, DISABLED
# kms_auth = HTTPKerberosAuth(mutual_authentication=DISABLED)
```

vikaskr22: thanks for the review @mneethiraj .

Actually I simply tried to keep all the parameters that can be modified at one place, this way it remains consistent with other params like kms_user, kms_url etc. 
With explicit kms_auth_type along with description , it would be clear for the new user that this part is configurable. Making inline changes is also fine, but to me it seems making changes in the execution logic.

It's only about convention, please let me know your thoughts, I will change it.

mneethiraj: @vikaskr22 - the comments have become too noisy, hence my suggestion to keeps the code simple for humans to read. I would suggest only the following 2 changes:

1. add following at the top of the file (not much value in "optimizing" to import only when using Kerbeors
```
from requests_kerberos import HTTPKerberosAuth, DISABLED
```

2.  Change `kms_auth` initialization to following:
```
kms_auth = HadoopSimpleAuth(kms_user)
# for kerberos:
# kms_auth = HTTPKerberosAuth(mutual_authentication=DISABLED)

```

vikaskr22: incorporated the review comments. Please check. Thanks.

ranger-tools/src/main/python/stress/stress_kms.py
```diff
@@ -29,7 +30,8 @@
 # This script requires Python package apache_ranger to be present.
 # The package can be installed using following command:
 #    pip3 install --upgrade apache_ranger
-#
+# If KMS is running with Kerberos, then 'requests-kerberos' package is required and can be installed using following command:
```
mneethiraj: I suggest removing the comment in line 33, as requests-kerberos package is no more optional due to reference in line 23.

vikaskr22: Removed that comment @mneethiraj . Please check now.

mneethiraj (approved):

PR #1152, opened 2026-08-13
RANGER-5742: Inconsistent handling of service config fields by name v…

vyommani:

Service config masking (read) and encryption (write) previously only matched the literal key "password". Service-defs can declare other config items with type="password" under different names (e.g. NiFi's nifi.ssl.keystorePassword /truststorePassword) - those were left unmasked/unencrypted.

This change treats any config item as a secret if its key is "password" OR the service-def declares it with type="password", applied consistently across RangerServiceService (read/view) and ServiceDBStore (create/update). updateService now preserves the correct per-key value on the hidden-sentinel case instead of a single shared variable. Shared classification logic lives in one place (ServiceDBStore) to avoid drift.

## How was this patch tested?

Added/updated unit tests in TestRangerServiceService and TestServiceDBStore covering masking, encryption, and per-key update behavior for service-def-declared secret keys.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
@@ -233,6 +233,7 @@ public class ServiceDBStore extends AbstractServiceStore {
     public static final     String                        RANGER_PLUGINS_CONFIG_CONF_PREFIX = "ranger.plugins.conf.";
     public static final     String                        HIDDEN_PASSWORD_STR               = "*****";
     public static final     String                        CONFIG_KEY_PASSWORD               = "password";
+    public static final     String                        CONFIG_TYPE_PASSWORD              = "password";
     public static final     String                        ACCESS_TYPE_DECRYPT_EEK           = "decrypteek";
     public static final     String                        ACCESS_TYPE_GENERATE_EEK          = "generateeek";
     public static final     String                        ACCESS_TYPE_GET_METADATA          = "getmetadata";
@@ -983,6 +984,7 @@ public class ServiceDBStore extends AbstractServiceStore {
 
         XXService             xCreatedService = daoMgr.getXXService().getById(service.getId());
         XXServiceConfigMapDao xConfMapDao     = daoMgr.getXXServiceConfigMap();
+        Set<String> passwordConfigKeys = getPasswordConfigKeys(service.getType());
 
         for (Entry<String, String> configMap : validConfigs.entrySet()) {
             String configKey   = configMap.getKey();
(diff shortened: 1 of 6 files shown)


security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -233,6 +233,7 @@ public class ServiceDBStore extends AbstractServiceStore {
     public static final     String                        RANGER_PLUGINS_CONFIG_CONF_PREFIX = "ranger.plugins.conf.";
     public static final     String                        HIDDEN_PASSWORD_STR               = "*****";
     public static final     String                        CONFIG_KEY_PASSWORD               = "password";
+    public static final     String                        CONFIG_TYPE_PASSWORD              = "password";
```
pradeepagrawal8184, line 236: How its different from above "CONFIG_KEY_PASSWORD"

vyommani: They both currently equal "password", but they're matched against completely different fields — the shared value is coincidental, not a duplicate:
	1-> CONFIG_KEY_PASSWORD — the literal config key name. Compared against configKey (e.g. svcConfMap.getConfigkey()), to catch the legacy case where a config item is literally named "password", regardless of what the service-def says about it.
	2-> CONFIG_TYPE_PASSWORD — the service-def declared config type. Compared against svcConfDef.getType(), to catch config items with any key name (e.g. nifi.ssl.keystorePassword) that the service-def marks as type="password".

vyommani:
@pradeepagrawal8184  @mneethiraj @ramackri  @dineshkumar-yadav please review this PR.

security-admin/src/main/java/org/apache/ranger/biz/ServiceDBStore.java
```diff
@@ -4245,6 +4250,28 @@ private static List<RangerPolicyDelta> compressDeltas(List<RangerPolicyDelta> de
         return ret;
     }
 
+    private Set<String> getPasswordConfigKeys(String serviceType) {
+        List<XXServiceConfigDef> svcConfDefList = serviceType != null ? daoMgr.getXXServiceConfigDef().findByServiceDefName(serviceType) : null;
```
mneethiraj: It will be efficient to retrieve only config names from the database with filter `type = password` (instead of all XXServiceConfigDef entries).

```
public class XXServiceConfigDefDao ... {
  ...
  public List<String> getConfigNamesByType(String serviceType, String configType) {
    return getEntityManager()
            .createNamedQuery("XXServiceConfigDef.getConfigNamesByType", String.class)
            .setParameter("serviceType", serviceType)
            .setParameter("type", configType).getResultList();
    }
}
```

vyommani: addressed

mneethiraj (approved):

PR #1156, opened 2026-08-14
RANGER-5746: Improvement in Ranger policy engine

maheshbandal15:

Align case-insensitive folding in RangerResourceTrie with the folding used by resource matchers.

When ignoreCase is enabled, the resource trie indexes child nodes using getLookupChar(), while policy resource matching uses StringUtils.equalsIgnoreCase() (equivalent to String.regionMatches(true, …)). Those two paths did not use the same Unicode case fold:

Trie lookup (before): Character.toLowerCase(ch)
Matchers: upper-case, then lower-case (equalsIgnoreCase semantics)
For some Unicode characters these folds diverge (for example i / ı, s / ſ, and other BMP equivalents). The trie could route a request to a different branch than the matcher would treat as equal, so candidate policy evaluators returned from trie pre-filtering could be incomplete relative to full matcher evaluation.

This change updates getLookupChar() to use Character.toLowerCase(Character.toUpperCase(ch)) when ignoreCase is true, matching the matcher’s ignore-case behavior. Case-sensitive resources are unchanged.

## How was this patch tested?

New unit tests (TestRangerResourceTrieCaseFolding):
* Documents Unicode pairs where equalsIgnoreCase and plain toLowerCase differ
* Confirms RangerDefaultResourceMatcher treats equivalent Unicode variants as matches
* Verifies trie lookup returns the expected evaluators for case-variant resources at branch boundaries
* Covers single-character resources, sibling-branch splits, wildcard + exact-resource combinations, ASCII regression, compressed trie nodes, and bidirectional lookup (i <--> ı)

Manual verification: Built ranger locally using mvn clean install; all targeted and related tests pass.

### DIFF
--- a/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerResourceTrie.java
+++ b/agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerResourceTrie.java
@@ -679,7 +679,7 @@ public class RangerResourceTrie<T extends RangerResourceEvaluator> {
     }
 
     private Character getLookupChar(char ch) {
-        return optIgnoreCase ? Character.toLowerCase(ch) : ch;
+        return optIgnoreCase ? Character.toLowerCase(Character.toUpperCase(ch)) : ch;
     }
 
     private Character getLookupChar(String str, int index) {
--- a/agents-common/src/test/java/org/apache/ranger/plugin/policyengine/TestRangerResourceTrieCaseFolding.java
+++ b/agents-common/src/test/java/org/apache/ranger/plugin/policyengine/TestRangerResourceTrieCaseFolding.java
@@ -42,14 +42,28 @@ import java.util.Set;
 import java.util.stream.Stream;
 
 class TestRangerResourceTrieCaseFolding {
-    private static final char DOTLESS_I = '\u0131'; // LATIN SMALL LETTER DOTLESS I
-    private static final char LONG_S    = '\u017F'; // LATIN SMALL LETTER LONG S
+    private static final char DOTLESS_I       = '\u0131'; // LATIN SMALL LETTER DOTLESS I
+    private static final char LONG_S          = '\u017F'; // LATIN SMALL LETTER LONG S
+    private static final char DOTTED_I        = '\u0130'; // LATIN CAPITAL LETTER I WITH DOT ABOVE
+    private static final char KELVIN_SIGN     = '\u212A'; // KELVIN SIGN
+    private static final char ANGSTROM_SIGN   = '\u212B'; // ANGSTROM SIGN
+    private static final char LATIN_A_RING    = '\u00E5'; // LATIN SMALL LETTER A WITH RING ABOVE
+    private static final char SHARP_S         = '\u00DF'; // LATIN SMALL LETTER SHARP S
+    private static final char CAPITAL_SHARP_S = '\u1E9E'; // LATIN CAPITAL LETTER SHARP S
 
     private static final RangerResourceDef TABLE_RESOURCE_DEF = getTableResourceDef();
 
     // folding sanity
     static Stream<Arguments> caseFoldingMismatchPairs() {
         return Stream.of(Arguments.of("i vs dotless-i", 'i', DOTLESS_I, true, false), Arguments.of("s vs long-s", 's', LONG_S, true, false), Arguments.of("I vs dotless-i", 'I', DOTLESS_I, true, false), Arguments.of("S vs long-s", 'S', LONG_S, true, false), Arguments.of("Sigma vs final-sigma", '\u03A3', '\u03C2', true, false), Arguments.of("ASCII self", 'a', 'A', true, true));
+    }
+
+    static Stream<Arguments> doubleFoldRequiredPairs() {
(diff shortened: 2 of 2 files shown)


dineshkumar-yadav (approved):
LGTM

vyommani (approved):
This fix correctly mitigates the reported issue — folding with Character.toLowerCase(Character.toUpperCase(ch)) instead of a plain toLowerCase matches the equivalence classes equalsIgnoreCase actually uses, so case-variant resources (e.g. Turkish dotless i, long s) no longer get routed to a trie branch missing a DENY evaluator that the real matcher would have matched. Test coverage looks solid.

One thing worth flagging, not for this PR: there's a narrower, related gap for supplementary-plane (astral) characters — scripts like Deseret or Adlam encoded as surrogate pairs — where a sibling policy can force the trie to split mid-pair and reintroduce the same class of issue one level deeper. This is very unlikely in production (resource names are essentially always ASCII/BMP), and properly closing it needs more than a localized fix, so I'd rather track it in a follow-up JIRA than extend this PR.

Recommend merging as-is.

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerResourceTrie.java
```diff
@@ -679,7 +679,7 @@ private int getMaxDepth() {
     }
 
     private Character getLookupChar(char ch) {
-        return optIgnoreCase ? Character.toLowerCase(ch) : ch;
+        return optIgnoreCase ? Character.toLowerCase(Character.toUpperCase(ch)) : ch;
```
mneethiraj, line 682: @maheshbandal15  - would using `Character.toUpperCase(ch)`, instead of two level conversion - `Character.toLowerCase(Character.toUpperCase(ch))`, address the reported issue?

vyommani: sigle toUpperCase(ch) alone fixes this specific reported case (ı vs i), but not others like İ vs i, Kelvin sign vs k, or Ångström sign vs å, which still diverge under single toUpperCase. The double-fold (toLowerCase(toUpperCase(ch))) is needed to cover those too, since it matches equalsIgnoreCase's actual behavior rather than just this one character. Please run the attached java program for detail.

[FoldProof.java](https://github.com/user-attachments/files/31111770/FoldProof.java)

mneethiraj: Thank you @vyommani  for the details and the program to demonstrate the issue.

@maheshbandal15  - can you please make sure that unit tests cover these cases?

maheshbandal15: Thank you all for the review. Thank you @mneethiraj for highlighting the current test coverage was limited to two-level conversion.
To give some context, the core issue revolves around how certain Unicode characters are routed and evaluated in the trie during case-insensitive matching. Using `Character.toLowerCase(Character.toUpperCase(ch))` is the most robust solution. It perfectly mirrors the matcher’s full two-step case folding process, resulting in zero BMP mismatches.
Relying solely on toUpperCase(ch) is a narrower fix. While it successfully closes the originally reported edge cases, it leaves other Unicode ignore-case pairs potentially mis-routed in the trie, as @vyommani pointed out.

Across the BMP, folding equivalence vs equalsIgnoreCase looks like this:
| Strategy | Mismatched pairs |
|--------|--------|
| Character.toLowerCase(ch) (old trie) | 50 |
| Character.toUpperCase(ch) | 13 |
| Character.toLowerCase(Character.toUpperCase(ch)) | 0 |

So toUpperCase alone still leaves 13 BMP pairs where the matcher says “equal” but the trie would use different keys.

There are roughly 60 BMP codes. To keep the test class lean and maintainable, I have opted to include only a representative few that adequately cover both case-folding scenarios rather than adding all of them to the test.

Request you to review the patch.

vyommani: Your latest tests look good, but they're a bit complex and don't exercise the public API (RangerBasePlugin.isAccessAllowed). I spent some time and wrote a test that fails without the fix and passes with it — please take a look at the attached test, and feel free to include it in the PR if you find it valuable.
[TrieCaseFoldDenySkipTest.java](https://github.com/user-attachments/files/31150021/TrieCaseFoldDenySkipTest.java)

vyommani (approved):
LGTM, please check my test if it is use full add it in PR

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerResourceTrie.java
maheshbandal15: @vyommani TestRangerResourceTrieCaseFolding is a component-level unit test that exercises trie and matcher behavior in isolation rather than exercising the public API RangerBasePlugin.isAccessAllowed. It mainly covers getLookupChar(), Trie build/insert, traverse / branch lookup through getEvaluatorsForResource(). RangerDefaultResourceMatcher — isMatch() with ignoreCase=true and wildCard=true.

I reviewed your Java code to see if I can integrate it into the existing test class.
Refactored for readability. Renamed denySkip-related helpers and policy-engine test methods to neutral names, and updated related constants and assertion helpers for consistency. Updated assertion line to match the rest of the test and to avoid false positives: variant.getIsAllowed() defaults to false (see lines 96 and 83/87 of your code). Resolved Checkstyle errors.

Please review the addendum patch which includes test coverage for angerBasePlugin.isAccessAllowed().

vyommani (approved):
LGTM

PR #1162, opened 2026-08-17
RANGER-5110: Prevent entity soft deletion when LDAP DN is updated

krishnamuttevi:
Fixes an issue in LDAP/AD sync where users or groups are incorrectly skipped or marked for deletion when their DN changes (e.g., due to OU moves). This PR correctly detects, maps, and updates DN changes efficiently using the existing cache.

**Key Changes:**
Allow DN Updates: Added an isLdapAdSync check in delta computation (computeUserDelta/computeGroupDelta). If a DN changes, the update is now explicitly allowed rather than skipped, and old DN mappings are cleaned up.

**Smart Deletion Check:** 
Before marking an entity as deleted due to a missing DN (computeDeletedUsers/computeDeletedGroups), we now verify if the base name still exists in the source. If it does, deletion is bypassed so the delta sync can naturally correct the DN.

**Why this approach (vs. PR #516)**
Better Performance: Avoids the heavy REST reloads and cache wiping introduced in #516. We utilize the existing cache.

Complete Fix: _Actually updates the entity with the new DN across cycles (PR #516 only attempts a soft-delete and fails to map the new DN)._

Cleaner: _Highly efficient fix with a significantly smaller LOC footprint._

**Testing:**

Verified LDAP/AD users and groups update correctly during OU moves.

Confirmed no extra cache-clearing loads are triggered.

Verified non-LDAP/AD sync behavior remains unaffected.

**Please check the below the below doc for the testcaes which was verified by runtime:** :

_[Test cases for the RANGER-5110_.pdf](https://github.com/user-attachments/files/31945942/Test.cases.for.the.RANGER-5110_.pdf)_


ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -747,14 +747,26 @@ private void computeGroupDelta(Map<String, Map<String, String>> sourceGroups) {
                     noOfModifiedGroups++;
                     groupNameMap.put(groupDN, groupName);
                 } else {
-                    if (MapUtils.isNotEmpty(curGroupAttrs) && !StringUtils.equalsIgnoreCase(groupDN, curGroupDN)) { // skip update
-                        LOG.debug("[{}]: SyncSource update skipped, current group DN = {} new user DN  = {}", groupName, curGroupDN, groupDN);
-
-                        if (StringUtils.equalsIgnoreCase(curGroupAttrsStr, newGroupAttrsStr)) {
-                            groupNameMap.put(groupDN, groupName);
+                    boolean isLdapAdSync = StringUtils.equalsIgnoreCase(newSyncSource, "LDAP/AD")
```
spolavarpau1: This will cause regressions in case where a user left the company and another person with same name joined in a different OU. In this case, we cannot assume user is same and update the DN of the user right?

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -747,14 +747,26 @@ private void computeGroupDelta(Map<String, Map<String, String>> sourceGroups) {
...
 
-                        continue;
+                            if (StringUtils.equalsIgnoreCase(curGroupAttrsStr, newGroupAttrsStr)) {
+                                groupNameMap.put(groupDN, groupName);
+                            }
+                            continue;
+                        }
                     }
 
                     if (StringUtils.isEmpty(curSyncSource) || (!StringUtils.equalsIgnoreCase(curGroupAttrsStr, newGroupAttrsStr) && StringUtils.equalsIgnoreCase(curSyncSource, newSyncSource))) { // update
```
spolavarpau1: When sync source is same and group OU is updated, then the groupAttrsStr will be different isn't it? As one of the groupAttrs is groupFullName. For this feature, can you please add a new config flag like "ranger.usersync.dn.validation.enabled" and in this if block at the end we need to add below change
if (isDNValidationEnabled) {
     groupNameMap.remove(curGroupDN);
}
Please check if this works.

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -838,15 +850,25 @@ private void computeUserDelta(Map<String, Map<String, String>> sourceUsers) {
                     noOfModifiedGroups++;
                     userNameMap.put(userDN, userName);
                 } else {
-                    if (MapUtils.isNotEmpty(curUserAttrs) && !StringUtils.equalsIgnoreCase(userDN, curUserDN)) { // skip update
-                        // Same username with different DN already exists
-                        LOG.debug("[{}]: SyncSource update skipped, current user DN = {} new user DN  = {}", userName, curUserDN, userDN);
-
-                        if (StringUtils.equalsIgnoreCase(curUserAttrsStr, newUserAttrsStr)) {
-                            userNameMap.put(userDN, userName);
+                    boolean isLdapAdSync = StringUtils.equalsIgnoreCase(newSyncSource, "LDAP/AD")
```
spolavarpau1: Same comments as above apply here as well

spolavarpau1 (changes requested):

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
krishnamuttevi: Thanks for the niche observation, I have updated the hardcoded boolean check and made it a custom valve property, to let the user decide and avoid the regression

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
krishnamuttevi: Thanks, I have modified and implemented the above suggestion before the conditional block to reduce the computional overhead.

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
krishnamuttevi: Have updated in the following commit.

krishnamuttevi: Please check the below the below doc for the testcaes which was verified by runtime

[Test cases for the RANGER-5110_.pdf](https://github.com/user-attachments/files/31269816/Test.cases.for.the.RANGER-5110_.pdf)

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -738,6 +740,18 @@ private void computeGroupDelta(Map<String, Map<String, String>> sourceGroups) {
                 String              curGroupDN       = MapUtils.isEmpty(curGroupAttrs) ? groupName : curGroupAttrs.get(UgsyncCommonConstants.FULL_NAME);
                 String              newSyncSource    = newGroupAttrs.get(UgsyncCommonConstants.SYNC_SOURCE);
 
+                // Universal DN Validation Check (Executes first for both startup & runtime)
+                if (MapUtils.isNotEmpty(curGroupAttrs) && !StringUtils.equalsIgnoreCase(groupDN, curGroupDN)) {
```
spolavarpau1: How about the case where line 755 condition is true? i.e., Customer has initially configured usersync with Unix as sync source and all the users/groups have the sync source as Unix and now they changed the sync source as AD/LDAP and they had isSyncSourceValidationEnabled is set to false and restarted ranger usersync. Here isDnValidationEnabled is set to false. Then the username/groupname DN is not updated isn't it? It will not come to line 755 at all right?

krishnamuttevi: Yes, As per the current version of the code if  isDnValidationEnabled = false, means blocking the domain name change and in this though inter sync source change was accepted by setting the issyncsourcevalidation enabled = false, the dn would be changed from Unix to dn of LDAP/Ad which would be eventually blocked as per the property isDnValidationEnabled.

So to prevent code duplication for both incremental and at startup (restart) as the check is necessary placed the dn validation block upfront.

If this 
Dn = Unix to Dn = {ou=xxx,dc=xx} of LDAP is also considered as dn change and currently it is blocked.
Should we allow this condition let it change at startup?

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -754,18 +756,21 @@ private void computeGroupDelta(Map<String, Map<String, String>> sourceGroups) {
...
                     groupNameMap.put(groupDN, groupName);
-                } else {
-                    if (MapUtils.isNotEmpty(curGroupAttrs) && !StringUtils.equalsIgnoreCase(groupDN, curGroupDN)) { // skip update
-                        LOG.debug("[{}]: SyncSource update skipped, current group DN = {} new user DN  = {}", groupName, curGroupDN, groupDN);
 
-                        if (StringUtils.equalsIgnoreCase(curGroupAttrsStr, newGroupAttrsStr)) {
-                            groupNameMap.put(groupDN, groupName);
-                        }
+                    if (isDnValidationEnabled) {
+                        groupNameMap.remove(curGroupDN);
```
spolavarpau1, line 761: I don't think we need to add this check for the existing case.

ugsync/src/main/java/org/apache/ranger/unixusersync/process/PolicyMgrUserGroupBuilder.java
```diff
@@ -777,6 +782,10 @@ private void computeGroupDelta(Map<String, Map<String, String>> sourceGroups) {
                         deltaGroups.put(groupName, curGroup);
                         noOfModifiedGroups++;
                         groupNameMap.put(groupDN, groupName);
+
+                        if (isDnValidationEnabled) {
```
spolavarpau1, line 786: I think you should check allowSyncSourceOverwrite instead of isDnValidationEnabled isn't it?

PR #1103, opened 2026-07-22
RANGER-4458: Usersync Webapp, metrics and Docker changes

Sanket-Shelar:
## What changes were proposed in this pull request?

Converted usersync to webapp module, added docker changes for usersync webapp and added metrics for usersync.

## How was this patch tested?

Manual build passing and tested on docker.

--------------------------------------------
[RANGER-4458](https://issues.apache.org/jira/browse/RANGER-4458): Ranger Usersync - Convert to Web Application
[RANGER-5701](https://issues.apache.org/jira/browse/RANGER-5701): Support for Usersync as WebApp in docker containers
[RANGER-5702](https://issues.apache.org/jira/browse/RANGER-5702): Add/ Update metric details for Ranger UserSync

PR #1171, opened 2026-08-21
RANGER-5698 : Disable Unix Authentication Service by default

fimugdha:

The Unix Authentication service is rarely utilized for Ranger authentication, as the vast majority of deployments have migrated to modern alternatives like PAM or LDAP.

To optimize default startup behavior, we are removing the Unix Authentication service from the Ranger Usersync default startup sequence.

Moving forward, this service will run only if explicitly enabled via the `ranger.usersync.unix.auth.enabled` property within the `ranger-ugsync-site.xml` configuration file.

## How was this patch tested?

1. Brought up ranger core services: ranger, usersync, tagsync, pdp and kms in containers.
2. Created a user on ranger & usersync containers.
3. Set the password for the user.
4. User got synced successfully on Ranger Admin.
5. Login was successfully with the user password on Ranger Admin with PAM authentication.

### DIFF
--- a/ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
+++ b/ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
@@ -195,12 +195,14 @@ public class UserGroupSyncConfig {
     /* SSL Configs */
     private static final String SSL_KEYSTORE_FILE_TYPE_PARAM       = "ranger.keystore.file.type";
     private static final String SSL_TRUSTSTORE_FILE_TYPE_PARAM     = "ranger.truststore.file.type";
-    private static final String SSL_KEYSTORE_PATH_PARAM            = "ranger.usersync.keystore.file";
-    private static final String SSL_KEYSTORE_PATH_PASSWORD_PARAM   = "ranger.usersync.keystore.password";
+    private static final String SSL_KEYSTORE_PATH_PARAM            = "ranger.usersync.service.https.attrib.keystore.file";
+    private static final String SSL_KEYSTORE_PATH_PASSWORD_PARAM   = "ranger.usersync.service.https.attrib.keystore.pass";
     private static final String SSL_TRUSTSTORE_PATH_PARAM          = "ranger.usersync.truststore.file";
     private static final String SSL_TRUSTSTORE_PATH_PASSWORD_PARAM = "ranger.usersync.truststore.password";
-    private static final String SSL_KEYSTORE_PATH_PASSWORD_ALIAS   = "usersync.ssl.key.password";
+    private static final String SSL_KEYSTORE_PATH_PASSWORD_ALIAS   = "ranger.usersync.service.https.attrib.keystore.pass";
     private static final String SSL_TRUSTSTORE_PATH_PASSWORD_ALIAS = "usersync.ssl.truststore.password";
+
+    private static final String UGSYNC_UNIX_AUTH_ENABLED = "ranger.usersync.unix.auth.enabled";
 
     private static volatile UserGroupSyncConfig me;
     private final           Properties          prop = new Properties();
@@ -409,6 +411,14 @@ public class UserGroupSyncConfig {
         }
 
         return prop.getProperty(SSL_TRUSTSTORE_PATH_PASSWORD_PARAM);
+    }
+
+    public boolean getUserSyncUnixAuth() {
+        boolean enableUnixAuth = false;
+        if (prop.containsKey(UGSYNC_UNIX_AUTH_ENABLED)) {
+            enableUnixAuth = Boolean.parseBoolean(prop.getProperty(UGSYNC_UNIX_AUTH_ENABLED));
+        }
+        return enableUnixAuth;
     }
 
     public long getUpdateMillisMin() {
--- a/unixauthservice/conf.dist/ranger-ugsync-default.xml
+++ b/unixauthservice/conf.dist/ranger-ugsync-default.xml
@@ -19,7 +19,7 @@
 <configuration>
 	<property>
 		<name>ranger.usersync.port</name>
-                <value>5151</value>
+		<value>5151</value>
 	</property>
 	<property>
 		<name>ranger.usersync.ssl</name>
--- a/unixauthservice/scripts/install.properties
+++ b/unixauthservice/scripts/install.properties
@@ -28,6 +28,11 @@ POLICY_MGR_URL =
 # defaults to unix
 SYNC_SOURCE = unix
 
+# Enable unix auth service, only when sync source is set to unix along with unix backend is set to passwd.
+# Also, Unix authentication is enabled in Ranger Admin.
+# defaults to false
+ENABLE_UNIX_AUTH = false
+
 #
 # Minimum Unix User-id to start SYNC.
 # This should avoid creating UNIX system-level users in the Policy Manager
@@ -61,10 +66,12 @@ hadoop_conf=/etc/hadoop/conf
 #
 CRED_KEYSTORE_FILENAME=/etc/ranger/usersync/conf/rangerusersync.jceks
 
-# SSL Authentication
+# SSL Authentication configuration to be enabled when unix authentication is enabled.
 AUTH_SSL_ENABLED=false
 AUTH_SSL_KEYSTORE_FILE=/etc/ranger/usersync/conf/cert/unixauthservice.jks
 AUTH_SSL_KEYSTORE_PASSWORD=UnIx529p
+
+# Usersync SSL Truststore configuration
 AUTH_SSL_TRUSTSTORE_FILE=
 AUTH_SSL_TRUSTSTORE_PASSWORD=
 
--- a/unixauthservice/src/main/java/org/apache/ranger/authentication/UnixAuthenticationService.java
+++ b/unixauthservice/src/main/java/org/apache/ranger/authentication/UnixAuthenticationService.java
@@ -112,6 +112,7 @@ public class UnixAuthenticationService {
     }
 
     public void run() {
+        boolean unixAuthServiceStarted = false;
         try {
             UserGroupSyncConfig config = UserGroupSyncConfig.getInstance();
             enableUnixAuth = config.getUserSyncUnixAuth();
@@ -123,6 +124,7 @@ public class UnixAuthenticationService {
             if (enableUnixAuth && "passwd".equalsIgnoreCase(unixBackend)) {
                 LOG.info("Enabling Unix Auth Service!");
                 init();
+                unixAuthServiceStarted = true;
                 startService();
             } else {
                 LOG.info("Unix Auth Service Disabled!");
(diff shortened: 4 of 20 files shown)


dev-support/ranger-docker/Dockerfile.ranger
```diff
@@ -26,12 +26,15 @@ ARG TARGETARCH
...
 COPY ./scripts/admin/ranger.sh                        ${RANGER_SCRIPTS}/
 COPY ./scripts/admin/create-ranger-services.py        ${RANGER_SCRIPTS}/
+COPY ./scripts/setup-ranger-shadow-group.sh           ${RANGER_SCRIPTS}/
 
 RUN    tar xvfz /home/ranger/dist/ranger-${RANGER_VERSION}-admin.tar.gz --directory=${RANGER_HOME} \
     && ln -s ${RANGER_HOME}/ranger-${RANGER_VERSION}-admin ${RANGER_HOME}/admin \
     && rm -f /home/ranger/dist/ranger-${RANGER_VERSION}-admin.tar.gz \
     && rm -f /opt/ranger/admin/install.properties \
     && mkdir -p /var/run/ranger /var/log/ranger /usr/share/java/ \
+    && chmod 755 ${RANGER_SCRIPTS}/setup-ranger-shadow-group.sh \
```
kumaab: Adding users/groups have moved to the base image: https://github.com/apache/ranger-tools/blob/main/docker/Dockerfile , please see this script: https://github.com/apache/ranger-tools/blob/0e27092aedb0db01be54300f2936a6244e80ec6c/docker/create_users_and_groups.sh#L1 if a user like `shadow` is really required to be added.

dev-support/ranger-docker/README.md
```diff
@@ -79,6 +79,10 @@ cd dev-support/ranger-docker
 # To enable file based sync source for usersync do:
 export ENABLE_FILE_SYNC_SOURCE=true
 
+# Unix authentication service is disabled by default in docker (ENABLE_UNIX_AUTH=false).
```
kumaab: The comment may be removed, it's not relevant in docker deployments.

kumaab (commented):
Thanks @fimugdha for the patch, added a few comments.

ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
```diff
@@ -195,13 +195,15 @@ public class UserGroupSyncConfig {
     /* SSL Configs */
     private static final String SSL_KEYSTORE_FILE_TYPE_PARAM       = "ranger.keystore.file.type";
     private static final String SSL_TRUSTSTORE_FILE_TYPE_PARAM     = "ranger.truststore.file.type";
-    private static final String SSL_KEYSTORE_PATH_PARAM            = "ranger.usersync.keystore.file";
-    private static final String SSL_KEYSTORE_PATH_PASSWORD_PARAM   = "ranger.usersync.keystore.password";
+    private static final String SSL_KEYSTORE_PATH_PARAM            = "ranger.usersync.service.https.attrib.keystore.file";
```
pradeepagrawal8184, line 198: it seems setup.py still using "ranger.usersync.keystore.file" and ranger-ugsync-template.xml does not have entry of  ranger.usersync.service.https.attrib.keystore.file ; am i missing something here.

unixauthservice/conf.dist/ranger-ugsync-default.xml
```diff
@@ -93,4 +93,12 @@
 		<name>ranger.usersync.dest.ranger.session.cookie.name</name>
 		<value>RANGERADMINSESSIONID</value>
 	</property>
+	<property>
+		<name>ranger.usersync.unix.auth.enabled</name>
+		<value>false</value>
+	</property>
+	<property>
+		<name>ranger.usersync.unix.backend</name>
+		<value>nss</value>
```
pradeepagrawal8184: what was the previous default value for this property?

unixauthservice/src/main/java/org/apache/ranger/authentication/server/RangerUserSyncStarter.java
```diff
@@ -43,8 +43,7 @@ public void init() {
     static class UnixAuthenticationServiceThread extends Thread {
         @Override
         public void run() {
-            String[] params = {"-enableUnixAuth"};
```
pradeepagrawal8184, line 46: address this in ranger_usersync.py file as well

unixauthservice/scripts/install.properties
```diff
@@ -61,10 +66,12 @@ hadoop_conf=/etc/hadoop/conf
 #
 CRED_KEYSTORE_FILENAME=/etc/ranger/usersync/conf/rangerusersync.jceks
 
-# SSL Authentication
-AUTH_SSL_ENABLED=false
+# SSL Authentication configuration to be used when unix authentication is enabled.
+AUTH_SSL_ENABLED=true
```
pradeepagrawal8184: i think we should this also as false when ENABLE_UNIX_AUTH is false by default.

unixauthservice/scripts/install.properties
```diff
@@ -28,6 +28,11 @@ POLICY_MGR_URL =
 # defaults to unix
 SYNC_SOURCE = unix
 
+# Enable unix auth service, only when sync source is set to unix along with unix backend is set to passwd.
+# Also, Unix authentication is enabled in Ranger Admin.
+# defaults to false
+ENABLE_UNIX_AUTH = false
+
```
pradeepagrawal8184, line 35: what will happen to upgrade case by default. probably they have to edit this line if they want to continue with unixauth.

unixauthservice/src/main/java/org/apache/ranger/authentication/UnixAuthenticationService.java
pradeepagrawal8184: Can you test this once when unix auth is disabled and usersync is running in HA once.

dev-support/ranger-docker/Dockerfile.ranger
fimugdha: @kumaab, Created this pull-request on ranger-tools repository : https://github.com/apache/ranger-tools/pull/16. Could you please review ? It will be needed for me to test this ranger side changes.

dev-support/ranger-docker/README.md
fimugdha: Okay, This comment will get removed from the latest patch.

ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
fimugdha: @pradeepagrawal8184, the "ranger.usersync.keystore.file" will be used in case of unix authentication is enabled. 

The "ranger.usersync.service.https.attrib.keystore.file" new config changes where not added as a part of this PR - https://github.com/apache/ranger/pull/1103/changes#diff-20795b8427231f8ec96f91b20a1bae2804a6f23327f44fe6156ee8677e9109f5.

Currently SSL enable is not used on default docker setup. 

Do you think I need to handle the change of adding this new config into the usersync `install.properties` and in `ranger-ugsync-template.xml` files in this pull-request ?

unixauthservice/conf.dist/ranger-ugsync-default.xml
fimugdha: The default value for this property in the code is `passwd`. Refer : [DEFAULT_UGSYNC_UNIX_BACKEND](https://github.com/apache/ranger/blob/master/ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java#L150) & [getUnixBackend()](https://github.com/apache/ranger/blob/master/ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java#L290-L298)

unixauthservice/src/main/java/org/apache/ranger/authentication/server/RangerUserSyncStarter.java
fimugdha: Sure, This change will get handled in the latest patch.

unixauthservice/scripts/install.properties
fimugdha: Yes. How upgrade scenario for such use-cases are getting handled on apache side ?

unixauthservice/src/main/java/org/apache/ranger/authentication/UnixAuthenticationService.java
fimugdha: @pradeepagrawal8184, Sure, I will test it and keep you updated.

unixauthservice/scripts/install.properties
fimugdha: okay, this value will get updated in the latest patch.

unixauthservice/conf.dist/ranger-ugsync-default.xml
pradeepagrawal8184: will there be any issue in upgrade case ?

unixauthservice/scripts/install.properties
pradeepagrawal8184: Yes, if he usage this updated file as is for upgrade then he needs to know the new default behaviours. mainly when he is already using unix auth before the upgrade.

unixauthservice/conf.dist/ranger-ugsync-default.xml
fimugdha: removed the config from the file

unixauthservice/scripts/install.properties
fimugdha: Discussed with @pradeepagrawal8184 for this review, post upgrade user need to enable this flag if using unix authentication.

ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
fimugdha: Discussed with @pradeepagrawal8184, changes for ranger.usersync.service.https.attrib.keystore.file in `install.properties`, `ranger-ugsync-template.xml` & `usersync setup.py` will be handled in the ticket fix for using common embedded server.

fimugdha: cc @Sanket-Shelar

unixauthservice/src/main/java/org/apache/ranger/authentication/UnixAuthenticationService.java
fimugdha: @pradeepagrawal8184 , Thanks for this review. Earlier the Unix Auth used to be enabled always using the argument `-enableUnixAuth`, the main thread remains alive inside `startService()` and the finally block never runs, therefore the curator leader election kept running. With this PR fix it caused a regression on leader selection when Unix Auth is disabled.

Updated the fix and tested this fix on a HA setup cluster with both unix auth enabled & disabled.

unixauthservice/src/main/java/org/apache/ranger/authentication/UnixAuthenticationService.java
```diff
@@ -105,31 +106,43 @@ public UnixAuthenticationService() {
...
+            unixBackend = config.getUnixBackend();
+
             LOG.info("Starting User Sync Service!");
             startUnixUserGroupSyncProcess();
             Thread.sleep(5000);
-            if (enableUnixAuth) {
+            if (enableUnixAuth && "passwd".equalsIgnoreCase(unixBackend)) {
                 LOG.info("Enabling Unix Auth Service!");
                 init();
+                unixAuthServiceStarted = true;
```
pradeepagrawal8184, line 127: unixAuthServiceStarted is set before startService() — If startService() throws before entering its accept loop, finally will still stop the HA latch while user sync keeps running. Unlikely in practice, and similar to pre-existing behavior when auth was always enabled, but moving the flag to after successful listener startup would be slightly safer.

fimugdha: @pradeepagrawal8184 , Agreed that setting unixAuthServiceStarted before startService() means a bind/SSL failure would still stop the HA curator leader latch in finally while UserGroupSync keeps running. That matches the pre-existing behavior when -enableUnixAuth was always passed and finally unconditionally stopped curator leader latch on any exit from run().

The primary regression this change fixes is when Unix Auth is disabled (the new default): the main thread returns after starting UserGroupSync, and we must not stop leader election in that case. Keeping the flag before startService() preserves the prior enabled-auth shutdown semantics on listener exit or startService() failure.

I kept the current placement for minimal change. Extensive HA testing has already been completed with the current placement.

spolavarpau1 (approved):

PR #1174, opened 2026-08-24
RANGER-5753: LDAP/Active Directory authentication does not validate local Ranger account status

vyommani:
When Ranger Admin is configured for LDAP or Active Directory authentication
(`ranger.authentication.method=LDAP` or `ACTIVE_DIRECTORY`), `RangerAuthenticationProvider`
did not check the local Ranger account's status after a successful directory bind. This meant
a user's local account status could be out of sync with their actual ability to sign in when
directory-based authentication was used, unlike local/JDBC authentication, which already
respects account status via the `STATUS` column read by the JDBC user-details query.

This change adds a status check, `UserMgr#isUserDisabled`, and calls it from
`RangerAuthenticationProvider#authenticate` right after each successful LDAP, LDAP-bind,
AD-bind, AD, PAM, and UNIX authentication branch, before the method returns — so a disabled
local account is rejected consistently across all supported authentication methods.

## How was this patch tested?
added new unit tests.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
@@ -1177,7 +1177,7 @@ public class UserMgr {
         return roleList;
     }
 
-    public boolean isUserDisabled(String loginId) {
+    public boolean isUserNotActive(String loginId) {
         if (loginId == null || loginId.trim().isEmpty()) {
             return false;
         }
--- a/security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
@@ -1177,6 +1177,14 @@ public class UserMgr {
         return roleList;
     }
 
+    public boolean isUserDisabled(String loginId) {
+        if (loginId == null || loginId.trim().isEmpty()) {
+            return false;
+        }
+        XXPortalUser xXPortalUser = daoManager.getXXPortalUser().findByLoginId(loginId);
+        return xXPortalUser != null && xXPortalUser.getStatus() == RangerCommonEnums.STATUS_DISABLED;
+    }
+
     @Transactional(readOnly = false, propagation = Propagation.REQUIRED)
     public XXPortalUser updateOldUserName(String userLoginId, String newUserName, String currentPassword) {
         if (userLoginId == null || newUserName == null || userLoginId.trim().isEmpty() || newUserName.trim().isEmpty()) {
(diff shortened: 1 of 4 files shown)


security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
```diff
@@ -1177,6 +1177,14 @@ public Collection<String> getRolesByLoginId(String loginId) {
         return roleList;
     }
 
+    public boolean isUserDisabled(String loginId) {
+        if (loginId == null || loginId.trim().isEmpty()) {
+            return false;
+        }
+        XXPortalUser xXPortalUser = daoManager.getXXPortalUser().findByLoginId(loginId);
+        return xXPortalUser != null && xXPortalUser.getStatus() == RangerCommonEnums.STATUS_DISABLED;
```
pradeepagrawal8184: what about changing this to : return xXPortalUser != null && xXPortalUser.getStatus() != RangerCommonEnums.ACT_STATUS_ACTIVE;

vyommani: addressed.

security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
```diff
@@ -1177,6 +1177,14 @@ public Collection<String> getRolesByLoginId(String loginId) {
         return roleList;
     }
 
+    public boolean isUserDisabled(String loginId) {
```
mneethiraj: `isUserDisabled()` method checks if the user status is "not active", which is different from disabled status. I suggest naming the method to `isUserActive()` and update all its references to handle the return value.

security-admin/src/main/java/org/apache/ranger/security/handler/RangerAuthenticationProvider.java
```diff
@@ -701,6 +714,13 @@ private Authentication getJDBCAuthentication(Authentication authentication, Stri
         return authentication;
     }
 
+    private void checkAccountNotDisabled(String userName) {
```
mneethiraj: I suggest renaming method `checkAccountNotDisabled()` to `ensureUserIsActive()`. Also, update the messages to replace "disabled" with "not active".

security-admin/src/main/java/org/apache/ranger/biz/UserMgr.java
mneethiraj: @vyommani  - I take back this comment. I will update here after further review.

mneethiraj: The method implementation checks if user is not active i.e., not explicitly if the user is disabled - note an user can be in one of other statuses like deactivated. So, I suggest:
- renaming `isUserDisabled()` to `isUserNotActive()`
- renaming `checkAccountNotDisabled() to `blockNotActiveUser()`

vyommani: addressed.

mneethiraj (approved):

PR #1176, opened 2026-08-24
RANGER-5756: Make J10056 skip automatic reactivation of disabled external users

pradeepagrawal8184:

J10056 previously set all external users in ACT_STATUS_DISABLED to active during upgrade. 
Disabled is a valid administrative state, and the patch cannot reliably tell which rows should be changed. 
Leave matching accounts unchanged and log them for operator review instead.

## How was this patch tested?


vyommani (approved):
LGTM

PR #1184, opened 2026-08-25
RANGER-5763: Fix NPE in HBase plugin snapshot hooks when table descriptor is null

shmilygkd:

HBase 2.6.3/2.6.4 call the preSnapshot coprocessor hook with a null TableDescriptor when the table does not exist: HBASE-29361 moved the snapshot hooks ahead of the table-existence check and passes the descriptor from getTableDescriptors().get() without a null check. RangerAuthorizationCoprocessor dereferences the descriptor in preSnapshot/preCloneSnapshot/preRestoreSnapshot, so taking a snapshot of a non-existent table throws an NPE inside the coprocessor, and with hbase.coprocessor.abortonerror defaulting to true the HMaster aborts.

HBase corrected the ordering in HBASE-29955 (2.6.5); this change makes the plugin tolerate a null descriptor: the permission check is skipped when the descriptor is null, and the operation is rejected by HBase itself.

## How was this patch tested?

Unit tests added to RangerAuthorizationCoprocessorTest: the snapshot hooks skip the permission check when the table descriptor is null and run it when the descriptor is present. The failure mode was observed on a cluster running HBase 2.6.3 with the ranger hbase plugin.

### DIFF
--- a/hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
+++ b/hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
@@ -319,16 +319,31 @@ public class RangerAuthorizationCoprocessor implements AccessControlService.Inte
 
     @Override
     public void preSnapshot(ObserverContext<MasterCoprocessorEnvironment> ctx, SnapshotDescription snapshot, TableDescriptor hTableDescriptor) throws IOException {
+        if (hTableDescriptor == null) {
+            // HBase can call this hook with a null table descriptor when the table does not exist (HBASE-29361); the operation fails in HBase anyway
+            return;
+        }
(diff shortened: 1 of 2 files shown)


shmilygkd:
ping @pradeepagrawal8184 @ramackri: the CI here hasn't started, it's waiting on workflow approval. Could one of you take a look when you get a chance? Thanks.

shmilygkd:
ping @mneethiraj. Could someone approve the CI run here? It's been in `action_required` since Aug 25, so the checks never ran.

The patch is 3 null guards on the HBase snapshot hooks: on 2.6.3/2.6.4, snapshotting a non-existent table passes a null TableDescriptor into preSnapshot and it NPEs, which aborts the HMaster. Upstream fixed the call order in 2.6.5 but the plugin still dereferences it. JIRA: RANGER-5763

PR #1185, opened 2026-08-25
RANGER-5764:Enhancing RangerUserStore to include or exclude user group mappings 

abhiishek26:

1>Currently, when RangerUserStoreCache refreshes the user store from the database, it always attempts to load user-to-group mappings. In large deployments with a high number of users and groups, loading these mappings on every refresh can significantly increase  overhead . 

2>I have made it to flag based approach , if flag is set to true then the mappings to the plugins will be provided by the UserStore download apis else not . 

3>The major cause of performance bottleneck was final Map<String, Set<String>> userGroups    = xUserMgr.getUserGroups();
which has been changed to  
final Map<String, Set<String>> userGroups       = downloadUserGroupsEnabled ? xUserMgr.getUserGroups() : new HashMap<>();
so based on the requirements we can download or skip user-group mappings in UserStore for Plugins 

## How was this patch tested?

1>For performance testing this patch was tested for 1 million users and .5 million groups where every 2 user was mapped to one group creating .5 million mappings 
2>mvn clean test and mvn clean install were performed locally

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/common/RangerUserStoreCache.java
+++ b/security-admin/src/main/java/org/apache/ranger/common/RangerUserStoreCache.java
@@ -27,6 +27,7 @@ import org.apache.ranger.plugin.util.RangerUserStore;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
@@ -42,6 +43,7 @@ public class RangerUserStoreCache {
 
     private final int             waitTimeInSeconds;
     private final boolean         dedupStrings;
+    private final boolean         includeUserGroupMappings;
     private final ReentrantLock   lock = new ReentrantLock();
     private       RangerUserStore rangerUserStore;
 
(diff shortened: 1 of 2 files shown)


vikaskr22:
@abhiishek26 , please check if you have used the correct JIRA ID in the commit.

security-admin/src/main/resources/conf.dist/ranger-admin-site.xml
```diff
@@ -379,6 +379,16 @@
 		<name>ranger.downloadpolicy.session.log.enabled</name>
 		<value>false</value>
 	</property>
+	<!-- User-store download Properties starts-->
+	<property>
+		<name>ranger.admin.userstore.download.usergroups.enabled</name>
```
mneethiraj: I suggest renaming the property to `ranger.admin.userstore.include.usergroups.mappings`.

spolavarpau1:
@abhiishek26 , can you also update the title of the commit with correct jira number? (https://issues.apache.org/jira/browse/RANGER-5764)

abhiishek26:
> @abhiishek26 , please check if you have used the correct JIRA ID in the commit.

Thanks Vikas , Sorry it was a mismatch

security-admin/src/main/resources/conf.dist/ranger-admin-site.xml
abhiishek26: Thanks @mneethiraj , I have updated it

abhiishek26:
> @abhiishek26 , can you also update the title of the commit with correct jira number? (https://issues.apache.org/jira/browse/RANGER-5764)

Thanks @spolavarpau1 , I have updated it

security-admin/src/main/java/org/apache/ranger/common/RangerUserStoreCache.java
```diff
@@ -42,13 +43,15 @@ public class RangerUserStoreCache {
 
     private final int             waitTimeInSeconds;
     private final boolean         dedupStrings;
+    private final boolean         downloadUserGroupsEnabled;
```
mneethiraj: `downloadUserGroupsEnabled` => `includeUserGroupMappings`

abhiishek26: its updated .

PR #1188, opened 2026-08-26
RANGER-5734: Update EmbeddedServer to use custom TrustStore

vikaskr22:

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java class reads and uses custom truststore and updates the defaultSSLContext.

But Ranger-Admin/KMS uses the connector approach and Tomcat's connector doesn't use the keyStore/TrustStore from the defaultContext. Here, code to set keyStore into Tomcat's connector is already available but similar code to set custom trustStore was missing. Hence , it was falling back to JVM's default cacerts.

As part of this PR, code has been added to use custom user provided  trustStore. 

## How was this patch tested?

-mvn build has passed
- Existing UTs has passed
- Took patch, applied and deployed on internal cluster and verified if custom trustStore is being used or not. For this, I enabled mTLS between KMS & Admin by setting _ranger.service.https.attrib.clientAuth=true_ and found it working.

### DIFF
--- a/embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
+++ b/embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
@@ -51,6 +51,7 @@ import java.security.SecureRandom;
 import java.security.UnrecoverableKeyException;
 import java.security.cert.CertificateException;
 import java.util.Date;
+import java.util.Enumeration;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -164,8 +165,12 @@ public class EmbeddedServer {
             ssl.setScheme("https");
             ssl.setAttribute("SSLEnabled", "true");
             ssl.setAttribute("sslProtocol", EmbeddedServerUtil.getConfig("ranger.service.https.attrib.ssl.protocol", DEFAULT_SSL_PROTOCOL));
-            ssl.setAttribute("keystoreType", EmbeddedServerUtil.getConfig("ranger.keystore.file.type", RANGER_KEYSTORE_FILE_TYPE_DEFAULT));
-            ssl.setAttribute("truststoreType", EmbeddedServerUtil.getConfig("ranger.truststore.file.type", RANGER_TRUSTSTORE_FILE_TYPE_DEFAULT));
+
+            String keystoreType    = EmbeddedServerUtil.getConfig("ranger.keystore.file.type", RANGER_KEYSTORE_FILE_TYPE_DEFAULT);
+            ssl.setAttribute("keystoreType", keystoreType);
+
+            String truststoreType    = EmbeddedServerUtil.getConfig("ranger.truststore.file.type", RANGER_TRUSTSTORE_FILE_TYPE_DEFAULT);
+            ssl.setAttribute("truststoreType", truststoreType);
 
             String clientAuth = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.clientAuth", "false");
 
@@ -176,6 +181,8 @@ public class EmbeddedServer {
             ssl.setAttribute("clientAuth", clientAuth);
 
             String providerPath    = EmbeddedServerUtil.getConfig("ranger.credential.provider.path");
+
+            // Resolve KeyStore & related properties
             String credentialAlias = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.keystore.credential.alias", "keyStoreCredentialAlias");
             String keystorePass    = null;
 
@@ -187,20 +194,8 @@ public class EmbeddedServer {
                 keystorePass = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.keystore.pass");
             }
 
-            String truststoreAlias = EmbeddedServerUtil.getConfig("ranger.truststore.alias");
-            String truststorePass  = null;
-
-            if (providerPath != null && truststoreAlias != null) {
-                truststorePass = CredentialReader.getDecryptedString(providerPath.trim(), truststoreAlias.trim(), EmbeddedServerUtil.getConfig("ranger.truststore.file.type", RANGER_TRUSTSTORE_FILE_TYPE_DEFAULT));
-
-                if (StringUtils.isBlank(truststorePass) || "none".equalsIgnoreCase(truststorePass.trim())) {
-                    truststorePass = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.pass");
-                }
-            }
-
             String keystoreFile    = getKeystoreFile();
             String keyAlias        = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.keystore.keyalias", "rangeradmin");
-            String keystoreType    = EmbeddedServerUtil.getConfig("ranger.keystore.file.type", RANGER_KEYSTORE_FILE_TYPE_DEFAULT);
             String validationError = validateHttpsKeystore(keystoreFile, keystorePass, keyAlias, keystoreType);
 
             if (validationError != null) {
@@ -211,12 +206,30 @@ public class EmbeddedServer {
             ssl.setAttribute("keystorePass", keystorePass);
             ssl.setAttribute("keystoreFile", keystoreFile);
 
-            String trustStoreFile = EmbeddedServerUtil.getConfig("ranger.truststore.file");
+            // Resolve TrustStore & related properties
+            String truststoreCredsAlias = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.credential.alias");
+            String truststorePass  = null;
+
+            if (providerPath != null && truststoreCredsAlias != null) {
+                truststorePass = CredentialReader.getDecryptedString(providerPath.trim(), truststoreCredsAlias.trim(), EmbeddedServerUtil.getConfig("ranger.truststore.file.type", RANGER_TRUSTSTORE_FILE_TYPE_DEFAULT));
+            }
+
+            if (StringUtils.isBlank(truststorePass) || "none".equalsIgnoreCase(truststorePass.trim())) {
+                truststorePass = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.pass");
+            }
+
+            String trustStoreFile = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.file");
+
             if (StringUtils.isNotBlank(trustStoreFile) && StringUtils.isNotBlank(truststorePass)) {
+                validationError = validateHttpsTruststore(trustStoreFile, truststorePass, truststoreType);
+
+                if (validationError != null) {
+                    LOG.warning("HTTPS configuration validation for trustStore failed: " + validationError + " TLS handshaking may fail if mTLS is enabled.");
+                }
                 ssl.setAttribute("truststorePass", truststorePass);
                 ssl.setAttribute("truststoreFile", trustStoreFile);
             } else {
-                LOG.info("TrustStore is not set, TrustStoreFile is " +  trustStoreFile + " and is TruststorePass empty " + StringUtils.isBlank(truststorePass));
+                LOG.info("Truststore not configured for HTTPS connector. File=" + trustStoreFile + ", and  is trustStorePassword empty " + StringUtils.isBlank(truststorePass));
             }
 
             String enabledProtocols        = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.ssl.enabled.protocols", DEFAULT_ENABLED_PROTOCOLS);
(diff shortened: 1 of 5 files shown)


embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -187,6 +187,17 @@ public void start() {
...
             }
 
+            String truststoreAlias = EmbeddedServerUtil.getConfig("ranger.truststore.alias");
+            String truststorePass  = null;
+
+            if (providerPath != null && truststoreAlias != null) {
+                truststorePass = CredentialReader.getDecryptedString(providerPath.trim(), truststoreAlias.trim(), EmbeddedServerUtil.getConfig("ranger.truststore.file.type", RANGER_TRUSTSTORE_FILE_TYPE_DEFAULT));
+
+                if (StringUtils.isBlank(truststorePass) || "none".equalsIgnoreCase(truststorePass.trim())) {
+                    truststorePass = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.pass");
```
pradeepagrawal8184: The fallback uses a config key that does not exist anywhere else in the repo:
Ranger’s documented and setup-script-populated property is ranger.truststore.password (see ranger-admin-default-site.xml and security-admin/scripts/setup.sh). A typical install will have:

ranger.truststore.file
ranger.truststore.alias
ranger.truststore.password (plain text in site XML)
With the new code, if credential-store decryption fails or returns "none", the fallback reads a non-existent property, so truststorePass stays blank and the truststore is never applied — even when ranger.truststore.password is configured.

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -187,6 +187,17 @@ public void start() {
                 keystorePass = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.keystore.pass");
             }
 
+            String truststoreAlias = EmbeddedServerUtil.getConfig("ranger.truststore.alias");
+            String truststorePass  = null;
+
+            if (providerPath != null && truststoreAlias != null) {
```
pradeepagrawal8184: Truststore password resolution nests the plain-text fallback inside the providerPath && truststoreAlias block. If either is missing, truststorePass is never resolved from plain text. That’s inconsistent with keystore behavior and can block valid configs.

Suggested pattern (aligned with keystore + existing Ranger properties):


`if (providerPath != null && truststoreAlias != null) {
    truststorePass = CredentialReader.getDecryptedString(...);
}
if (StringUtils.isBlank(truststorePass) || "none".equalsIgnoreCase(truststorePass.trim())) {
    truststorePass = EmbeddedServerUtil.getConfig("ranger.truststore.password");
}`

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -200,6 +211,14 @@ public void start() {
             ssl.setAttribute("keystorePass", keystorePass);
             ssl.setAttribute("keystoreFile", keystoreFile);
 
+            String trustStoreFile = EmbeddedServerUtil.getConfig("ranger.truststore.file");
```
pradeepagrawal8184: Keystore gets upfront validation via validateHttpsKeystore() with a clear LOG.severe on failure. Truststore has no equivalent — a bad path or password only surfaces as a vague info log: 
a validateHttpsTruststore() (or reuse of existing loading logic) would give operators actionable errors, similar to keystore.

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -200,6 +211,14 @@ public void start() {
...
+            if (StringUtils.isNotBlank(trustStoreFile) && StringUtils.isNotBlank(truststorePass)) {
+                ssl.setAttribute("truststorePass", truststorePass);
+                ssl.setAttribute("truststoreFile", trustStoreFile);
+            } else {
+                LOG.info("TrustStore is not set, TrustStoreFile is " +  trustStoreFile + " and is TruststorePass empty " + StringUtils.isBlank(truststorePass));
```
pradeepagrawal8184: Prefer something like: "Truststore not configured for HTTPS connector (file={}, password={})" at DEBUG, or only log when clientAuth requires it.

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
vikaskr22: @pradeepagrawal8184 ,  Isn't  getTrustManagers() method doing the same thing ? If there is any such misconfigurations, wouldn't it be caught there and logged ?

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
vikaskr22: Agree. I am fixing this.

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
vikaskr22: @pradeepagrawal8184 , I have tried to follow the pattern of getTrustManager() method along with introduction of a new property named **ranger.service.https.attrib.truststore.pass** .

First it tries to read password from CredentialStore and if it's empty, then fallbacks to above property.  

Should we add above property in the ranger-admin-default-site.xml as well ? And should we also need to update the setup.sh script to populate this new property ?

pradeepagrawal8184: yes

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
pradeepagrawal8184: then find why do we have getKeyManagers

vikaskr22:
@pradeepagrawal8184 , I have tried to address the above review comments.  Testing is still in-progress but request you to please review it if any other aspects are missing. Thanks.

kms/scripts/setup.sh
```diff
@@ -1112,6 +1114,37 @@ update_properties() {
...
+    			updatePropertyToFilePy $propertyName $newPropertyValue $to_file_kms_site
+
+    			if [ "${keystore}" != "" ]
+    			then
+    				propertyName=ranger.service.https.attrib.truststore.pass
+    				newPropertyValue="_"
+    				updatePropertyToFilePy $propertyName $newPropertyValue $to_file_kms_site
+    				$PYTHON_COMMAND_INVOKER ranger_credential_helper.py -l "cred/lib/*" -f "$keystore" -k "$policymgr_https_truststore_credential_alias" -v "$ranger_kms_https_truststore_password" -c 1
+    			else
+    				propertyName=ranger.service.https.attrib.truststore.pass
```
pradeepagrawal8184, line 1135: When keystore is set, password is written as _, credential helper runs, then a second block may overwrite with plaintext if the keystore file is missing. security-admin’s version is cleaner (fallback only inside the keystore != "" branch).

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -200,6 +206,32 @@ public void start() {
...
+            if (StringUtils.isNotBlank(trustStoreFile) && StringUtils.isNotBlank(truststorePass)) {
+                validationError = validateHttpsTruststore(trustStoreFile, truststorePass, truststoreType);
+
+                if (validationError != null) {
+                    LOG.warning("HTTPS configuration validation for trustStore failed: " + validationError + " TLS handshaking may fail if mTLS is enabled.");
+                }
+                ssl.setAttribute("truststorePass", truststorePass);
+                ssl.setAttribute("truststoreFile", trustStoreFile);
+            } else {
+                LOG.info("Truststore not configured for HTTPS connector. File=" + trustStoreFile + ", and  is trustStorePassword empty " + StringUtils.isBlank(truststorePass));
```
pradeepagrawal8184, line 232: and  is => there is extra space. 
may be change that to :
Truststore not configured for HTTPS connector. file={}, passwordEmpty={}".

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -200,6 +206,32 @@ public void start() {
...
+
+            String trustStoreFile = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.file");
+
+            if (StringUtils.isNotBlank(trustStoreFile) && StringUtils.isNotBlank(truststorePass)) {
+                validationError = validateHttpsTruststore(trustStoreFile, truststorePass, truststoreType);
+
+                if (validationError != null) {
+                    LOG.warning("HTTPS configuration validation for trustStore failed: " + validationError + " TLS handshaking may fail if mTLS is enabled.");
+                }
+                ssl.setAttribute("truststorePass", truststorePass);
```
pradeepagrawal8184, line 229: if there is validation failure then probably no need to set.

pradeepagrawal8184 (changes requested):

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
vikaskr22: @pradeepagrawal8184 , this is the same pattern that we use for keystore. We validate and log but proceed further. I believe idea is to not block the user from using any keystore/truststore .  

For example: For trustStore verification, we check the trustStore size, it should not be zero. Similarly we check if it contains any TrustedCertificateEntry.   Without this, trustStore is still valid, though empty. And the communication without mTLS will continue to work. That's why i am simply logging at warn.

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
vikaskr22: @pradeepagrawal8184 , output of this is written to catalina.out for Admin and to stdout.log for KMS. **This class EmbeddedServer uses JUL, not SLF4J.**  You can see the imports. With JUL, placeholder pattern **{}** is not supported, probably that's why everywhere string concatenation is used.

kms/scripts/setup.sh
vikaskr22: I agree, security-admin approach is cleaner than KMS. 

On overwrite if keystore is missing, isn't this expected  to write the password in the config file ? Both KMS & Admin will have  same outcome, through a different  IF-Else branching. Am I missing anything ?

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
vikaskr22: Fixed now.

embeddedwebserver/src/main/java/org/apache/ranger/server/tomcat/EmbeddedServer.java
```diff
@@ -200,6 +206,32 @@ public void start() {
             ssl.setAttribute("keystorePass", keystorePass);
             ssl.setAttribute("keystoreFile", keystoreFile);
 
+            // Resolve TrustStore & related properties
+            String truststoreCredsAlias = EmbeddedServerUtil.getConfig("ranger.service.https.attrib.truststore.credential.alias");
+            String truststorePass  = null;
+
+            if (providerPath != null && truststoreCredsAlias != null) {
+                truststorePass = CredentialReader.getDecryptedString(providerPath.trim(), truststoreCredsAlias.trim(), EmbeddedServerUtil.getConfig("ranger.truststore.file.type", RANGER_TRUSTSTORE_FILE_TYPE_DEFAULT));
```
pradeepagrawal8184, line 213: Wrong store type passed to CredentialReader — Truststore password decryption uses ranger.truststore.file.type, but the third argument to CredentialReader.getDecryptedString() selects how to open the credential provider file, not the SSL truststore. The keystore password lookup correctly uses ranger.keystore.file.type. If those types differ (e.g. credential store is bcfks, SSL truststore is jks), password decryption can fail silently and fall back to plaintext. Fix: use keystoreType (already resolved at line 169) instead of ranger.truststore.file.type.

vikaskr22: @pradeepagrawal8184 , Generally keyStoreType/trustStoreType and credentialStore type remains same. There is not any separate property for credentialStoreType. It uses the same keyStore type value for credStore as well.

Whatever value is being used for credStoreType, it should be same every time credStore is being read. Hence I have updated the code to use keyStoreType as suggested. 

Please review one more time. Thanks.

pradeepagrawal8184: Can you find other code references for the same ?

vikaskr22: In the same file, getKeyManagers() method and getTrustManagers() method, both are reading the same credentialStore as they are reading the same property "ranger.credential.provider.path" . 

But see the storeType that is being used at two places, in getKeyManagers() it uses keyStoreType (ranger.keystore.file.type) to read from credStore and in getTrustStoreManagers(), it uses trustStoreType(ranger.truststore.file.type).

But in our case,  I have updated the code as suggested to use the same storeType, that is, keyStoreType while reading from the credentialStore. Is something missing there ?

PR #1195, opened 2026-08-28
RANGER-5772: replace use of elasticsearch libraries with opensearch libraries in OpenSearchAuditDestination

mneethiraj:

- updated OpenSearchAuditDestination to replace use of elasticsearch libraries with opensearch libraries.
- updated docker setup to use Opensearch 3.8.0

## How was this patch tested?

Brought up Ranger admin and audit services in docker setup with opensaerch as audit index store. Verified that audits generated by Kafka plugin are accessible in Ranger admin.


dev-support/ranger-docker/docker-compose.ranger-audit-service.yml
```diff
@@ -134,8 +134,8 @@ services:
     hostname: ranger-opensearch.rangernw
     volumes:
       - ./scripts/opensearch/opensearch.yml:/usr/share/opensearch/config/opensearch.yml
-      - opensearch-data:/usr/share/opensearch/data
```
kumaab, line 137: Let's keep the `named volumes` here instead of switching to ./-relative `bind mounts` ?

1. The switch from named volumes (storage that Docker creates and manages internally) to ./-relative bind mounts is breaking CI — ranger-audit-ingestor exits(1) on startup in the services-docker-build job: https://github.com/apache/ranger/actions/runs/33148952833/job/98883370423?pr=1195
 
   Root cause: the `ingestor` image runs as non-root (USER `ranger`) and writes to `/var/log/ranger/audit-ingestor` and `/var/log/ranger/audit-ingestor/audit`. 
 
   With named volumes, Docker initializes the volume from the image and preserves the `ranger:ranger` ownership set in the Dockerfile, so the process can write. With bind mounts, Docker creates the host dirs as `root:root` on a Linux host and does not copy image ownership, so ranger can't write to its log/spool dir and the container dies immediately. 
 
   This doesn't reproduce on Docker Desktop for macOS/Windows because its file-sharing layer masks ownership — which is why it passes locally but fails on the Linux CI runner.

2. `docker compose down -v` no longer gives a clean slate. Named volumes are removed by down -v; bind-mounted host dirs are not. So the OpenSearch index, ingestor spool, and logs persist across teardowns and have to be manually `rm -rf`'d for a true fresh start. For an audit pipeline this also risks misleading test runs (replayed spool events / stale indexed docs from a previous run).

mneethiraj: Thanks @kumaab for the details. With named volumes, startup of ranger-opensearch container fails in my laptop with the following error. Hence I changed to bind mounts. Perhaps I am missing something in my local?

```
$ docker logs -f ranger-opensearch
Enabling OpenSearch Security Plugin
Enabling execution of install_demo_configuration.sh for OpenSearch Security Plugin
OpenSearch 2.12.0 onwards, the OpenSearch Security Plugin a change that requires an initial password for 'admin' user.
Please define an environment variable 'OPENSEARCH_INITIAL_ADMIN_PASSWORD' with a strong password string.
If a password is not provided, the setup will quit.
 For more details, please visit: https://opensearch.org/docs/latest/install-and-configure/install-opensearch/docker/
### OpenSearch Security Demo Installer
### ** Warning: Do not use on production or public reachable systems **
OpenSearch install type: rpm/deb on Linux 6.12.76-linuxkit aarch64
OpenSearch config dir: /usr/share/opensearch/config/
OpenSearch config file: /usr/share/opensearch/config/opensearch.yml
OpenSearch bin dir: /usr/share/opensearch/bin/
OpenSearch plugins dir: /usr/share/opensearch/plugins/
OpenSearch lib dir: /usr/share/opensearch/lib/
Detected OpenSearch Version: 3.8.0
Detected OpenSearch Security Version: 3.8.0.0
/usr/share/opensearch/config/opensearch.yml seems to be already configured for Security. Quit.
Enabling execution of OPENSEARCH_HOME/bin/opensearch-performance-analyzer/performance-analyzer-agent-cli for OpenSearch Performance Analyzer Plugin
./opensearch-docker-entrypoint.sh: line 60: /usr/share/opensearch/logs/PerformanceAnalyzer.log: Permission denied
Exception in thread "main" java.lang.RuntimeException: starting java failed with [1]
output:
[0.000s][error][logging] Error opening log file 'logs/gc.log': Permission denied
[0.000s][error][logging] Initialization of output 'file=logs/gc.log' using options 'filecount=32,filesize=64m' failed.
error:
Could not rename log file 'logs/gc.log' to 'logs/gc.log.03' (Permission denied).
Invalid -Xlog option '-Xlog:gc*,gc+age=trace,safepoint:file=logs/gc.log:utctime,pid,tags:filecount=32,filesize=64m', see error log for details.
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
	at org.opensearch.tools.launchers.JvmErgonomics.flagsFinal(JvmErgonomics.java:125)
	at org.opensearch.tools.launchers.JvmErgonomics.finalJvmOptions(JvmErgonomics.java:87)
	at org.opensearch.tools.launchers.JvmErgonomics.choose(JvmErgonomics.java:70)
	at org.opensearch.tools.launchers.JvmOptionsParser.jvmOptions(JvmOptionsParser.java:150)
	at org.opensearch.tools.launchers.JvmOptionsParser.main(JvmOptionsParser.java:108)
```

kumaab: Sure, looks like a stale volume issue:
>./opensearch-docker-entrypoint.sh: line 60: /usr/share/opensearch/logs/PerformanceAnalyzer.log: Permission denied

Named volumes persist across `docker compose down` so it's possible an old state is sticking around.
Please try a fresh setup and also delete named volumes explicitly by:
~~~
docker volume rm ranger-docker_opensearch-data ranger-docker_opensearch-logs
~~~

kumaab (approved):
LGTM

PR #1199, opened 2026-08-31
RANGER-5776: Prevent DB2HSM migration tool from overwriting existing keys in HSM

ChinmayHegde24:
Currently, when the master key is migrated from the KMS DB to Luna HSM using the migration tool **(DB2HSMMKUtil)**, it silently overwrites any existing key present in the HSM. This overwrite should never happen, as it could lead to loss of data if the HSM already holds the active master key.

The fix is added to check the presence of alias in HSM before writing to the HSM. If the alias is already present, the write is skipped and a warning is logged.
This brings the behaviour in sync with how **HSM2DBMKUtil** works.
(Note: This is also missing in **RangerSafenetKeySecure** side so added fix here too)


vikaskr22 (approved):
Looks good to me.

dhavalshah9131 (approved):

PR #1201, opened 2026-08-31
RANGER-5750:Handle authorization of hbase:acl table in RangerAuthorizationCoprocessor in HBase Ranger plugin

rameeshm:
RANGER-5750:Handle authorization of hbase:acl table in RangerAuthorizationCoprocessor in HBase Ranger plugin

 Currently hbase:acl table is treated as special table and this should be authorized through Ranger plugin when users are not super users.

## How was this patch tested?

Patch was tested in Docker setup by running the unit test and manual test in Hbase shell.

### DIFF
--- a/hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
+++ b/hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
@@ -1390,7 +1390,7 @@ public class RangerAuthorizationCoprocessor implements AccessControlService.Inte
             LOG.warn("canSkipAccessCheck: exiting{}", "Unexpeceted: User is null: access denied, not audited!");
 
             throw new AccessDeniedException("No user associated with request (" + operation + ") for action: " + access + "on table:" + table);
-        } else if (isAccessForMetadataRead(access, table)) {
+        } else if (isAccessForMetadataRead(access, table, user)) {
             LOG.debug("canSkipAccessCheck: true: metadata read access always allowed, not audited");
 
             result = true;
@@ -1434,8 +1434,25 @@ public class RangerAuthorizationCoprocessor implements AccessControlService.Inte
 
     /* ---- EndpointObserver implementation ---- */
 
-    boolean isAccessForMetadataRead(String access, String table) {
+    boolean isAccessForMetadataRead(String access, String table, User user) {
         if (authUtils.isReadAccess(access) && isSpecialTable(table)) {
+            if (StringUtils.equals(table, "hbase:acl")) {
+                boolean isSystemOrSuperUser = false;
+                try {
+                    User currentUser = User.getCurrent();
+                    if (currentUser != null && user != null) {
+                        isSystemOrSuperUser = Objects.equals(currentUser.getShortName(), user.getShortName());
+                    }
+                } catch (IOException e) {
+                    LOG.warn("Unable to obtain the current user", e);
+                }
+                if (!isSystemOrSuperUser && user != null) {
+                    isSystemOrSuperUser = userUtils.isSuperUser(user);
+                }
+                if (!isSystemOrSuperUser) {
+                    return false;
+                }
+            }
             LOG.debug("isAccessForMetadataRead: Metadata tables read: access allowed!");
 
             return true;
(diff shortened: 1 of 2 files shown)


hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
```diff
@@ -1434,8 +1434,25 @@ boolean canSkipAccessCheck(User user, final String operation, String access, fin
...
+                    LOG.warn("Unable to obtain the current user", e);
+                }
+                if (!isSystemOrSuperUser && user != null) {
+                    isSystemOrSuperUser = userUtils.isSuperUser(user);
+                }
+                if (!isSystemOrSuperUser) {
+                    return false;
+                }
+            }
             LOG.debug("isAccessForMetadataRead: Metadata tables read: access allowed!");
```
kumaab: debug could be updated to also log when the access is not allowed.

kumaab (approved):
LGTM

hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
```diff
@@ -1434,13 +1434,34 @@ boolean canSkipAccessCheck(User user, final String operation, String access, fin
 
     /* ---- EndpointObserver implementation ---- */
 
-    boolean isAccessForMetadataRead(String access, String table) {
+    boolean isAccessForMetadataRead(String access, String table, User user) {
         if (authUtils.isReadAccess(access) && isSpecialTable(table)) {
-            LOG.debug("isAccessForMetadataRead: Metadata tables read: access allowed!");
+            if (StringUtils.equals(table, "hbase:acl")) {
```
vyommani, line 1439: This duplicates the system-or-super-user predicate already in requireSystemOrSuperUser(). Can we extract it
boolean isSystemOrSuperUser(User user) {
    if (user == null) return false;
    if (userUtils.isSuperUser(user)) return true;
    try {
        User loginUser = User.getCurrent();
        return loginUser != null && Objects.equals(loginUser.getShortName(), user.getShortName());
    } catch (IOException excp) {
        LOG.warn("isSystemOrSuperUser: unable to obtain login user; treating [{}] as non-system", user.getShortName(), excp);
        return false;
    }
}
and have both callers use it? Then this method reduces to if (ACL_TABLE_NAME.equals(table) && !isSystemOrSuperUser(user)) return false;. Also use PermissionStorage.ACL_TABLE_NAME.getNameAsString() instead of the literal

hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
```diff
@@ -1434,13 +1434,34 @@ boolean canSkipAccessCheck(User user, final String operation, String access, fin
...
+                        }
+                    } catch (IOException e) {
+                        LOG.warn("Unable to obtain the current user", e);
+                    }
+                }
+                if (!isSystemOrSuperUser) {
+                    return false;
+                }
+            }
+            LOG.debug("isAccessForMetadataRead: Metadata tables read: access allowed for user: {}!", (user != null ? user.getShortName() : ""));
```
vyommani, line 1458: This now fires for every access to every non-special table (each Get/Put/Scan on user tables), and the message says "Metadata tables read: not access allowed" even when the table isn't a metadata table. Please move it inside the isSpecialTable branch or drop it.

hbase-agent/src/test/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessorTest.java
```diff
@@ -391,12 +391,27 @@ public void test18_requirePermission_region_allowsOrDenies() throws Exception {
     }
 
     @Test
-    public void test19_isSpecialTable_and_metadataRead() {
+    public void test19_isSpecialTable_and_metadataRead() throws Exception {
```
vyommani, line 394: The case this change exists for isn't asserted: a non-system, non-super user reading hbase:acl → false. Please add that, plus a regression guard that an ordinary user reading hbase:meta is still true.

hbase-agent/src/main/java/org/apache/ranger/authorization/hbase/RangerAuthorizationCoprocessor.java
rameeshm: @vyommani  for this I had to do a small refactoring on isSystemOrSuperUser. Please review this change

vyommani (approved):

PR #1203, opened 2026-09-01
RANGER-5414: Remove unused (dead code) RangerKMSStartUp and related c…

vikaskr22:
…lasses

A dead class kms/src/main/java/org/apache/ranger/kms/biz/RangerKMSStartUp.java exists in the repo. Removing this class along with other related classes like corresponding Test class and one change in RangerMasterKey. Due to this class only, RangerMasterKey is required to have the default constructor, that is of no use.

## How was this patch tested?

- mvn build passed
- UT passed
- Already tested in internal cluster by removing this class.


dhavalshah9131 (approved):
I confirmed that deleted class files are not being referred or in use anymore.

PR #1208, opened 2026-09-03
RANGER-5773: Harden encryption of stored service configuration secrets

vyommani:

Strengthens how Ranger Admin protects certain encrypted configuration values (e.g. service
connection passwords) at rest, and migrates existing data to the improved approach. New writes
get the improved protection immediately; a one-time patch upgrades data written before this
change.

## How was this patch tested?

Ran end-to-end against a real running instance (`dev-support/ranger-docker`), not just unit
tests, using genuine legacy-format data:

- **Migration**: all genuinely-legacy rows converted; each row's stored value differs from its
  pre-migration value (confirms a real decrypt/re-encrypt, not a relabel); zero failures.
- **HA mismatch**: pointed one node at a different key than the one data was migrated under —
  fails closed with an actionable error instead of a silent bad-password auth failure; reverting
  the key restores correct decryption.
- **Idempotency**: re-running the migration patch against already-migrated data is a no-op, no
  failures.
- Full `mvn test` / project build verification is still pending — this pass was done in an
  environment without Maven Central access, so verification so far is against the modified
  classes compiled and run directly, not through the project's own build.

### Open questions for reviewers (flagged in-code, not silently decided)

1. Migration patch runs single-pass/single-transaction (matches existing precedent in this
   codebase) — acceptable at very large scale, or should it batch?
2. The new consistency checker WARNs and lets startup continue on a key mismatch rather than
   failing it outright, to avoid turning a legitimate transient state (e.g. an in-progress key
   rotation) into an outage. Right trade-off, or should a hard mismatch block startup?
3. The new read-path branch relies on the new decrypt call throwing reliably on a wrong key,
   rather than repeating the existing byte-exact re-encrypt-and-compare self-check the legacy
   branches use. Is that reliance justified as-is?

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/patch/PatchServicePasswordV2Migration_J10067.java
+++ b/security-admin/src/main/java/org/apache/ranger/patch/PatchServicePasswordV2Migration_J10067.java
@@ -1,0 +1,238 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.ranger.patch;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.ranger.biz.ServiceDBStore;
+import org.apache.ranger.db.RangerDaoManager;
+import org.apache.ranger.entity.XXService;
+import org.apache.ranger.entity.XXServiceConfigMap;
+import org.apache.ranger.plugin.util.PasswordUtils;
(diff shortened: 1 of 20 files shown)


security-admin/src/main/java/org/apache/ranger/patch/PatchServicePasswordV2Migration_J10067.java
```diff
@@ -0,0 +1,238 @@
...
+ * as its own per-row failure with the same root cause.
+ * - A row is only treated as "legacy-format, needs migrating" when {@code
+ * PasswordUtils.isLegacyFormat()} recognizes it (first field names a real crypto algorithm),
+ * not merely because its stored value contains a comma. The old, looser
+ * {@code storedValue.contains(",")} heuristic could misclassify a plaintext password that
+ * happens to contain a comma as "already encrypted" and attempt to decrypt/re-encrypt it —
+ * rows that don't match either recognized format are counted as notEncrypted and left alone.
+ */
+@Component
+public class PatchServicePasswordV2Migration_J10067 extends BaseLoader {
```
pradeepagrawal8184: Please change the patch number to J10070; others previous numbers are in progress by other developers.

pradeepagrawal8184 (changes requested):

security-admin/src/main/java/org/apache/ranger/patch/PatchServicePasswordV2Migration_J10067.java
pradeepagrawal8184: Also make an entry in core schema file of all DB's like : https://github.com/apache/ranger/blob/master/security-admin/db/mysql/optimized/current/ranger_core_db_mysql.sql#L2009

vyommani:
@mneethiraj can you please review this PR ?

PR #1209, opened 2026-09-03
RANGER-3917 : Ranger UI should have the provision to display the audit metrics for each active plugin

dhavaljrajpara:
Ranger UI should have the provision to display the audit metrics for each active plugin

## How was this patch tested?
1.Applied the patch locally.
2.Verified that the build completed successfully.
3.Checked that the development server started and worked properly.
4.Performed UI-level sanity testing on the policy modules.
5)Tested matric tab related changes and also validated all graph related data populated properly or not.


security-admin/src/main/webapp/react-webapp/src/views/AuditEvent/Metrics/MetricsLogs.jsx
```diff
@@ -0,0 +1,407 @@
...
+
+  const closeMetricGraphModal = () => {
+    dispatch({
+      type: ACTIONS.SHOW_METRIC_GRAPH_MODAL,
+      showMetricGraphModal: false,
+      metricGraphData: state.metricGraphData
+    });
+  };
+
+  const columns = React.useMemo(
```
fimugdha, line 158: @dhavaljrajpara  `Metrics Details` column is missing. Any reason why this column is getting skipped ?

dhavaljrajpara: Yes,
In the new implementation, they removed the Metrics Details data from the API response because the data is now fetched from Solr, which doesn't contain that field. Dinesh suggested removing that column as a result.

fimugdha (approved):

PR #1218, opened 2026-09-08
RANGER-5696: Added deltaSyncServerType config to optimize delta sync …

krishnamuttevi:
### Summary: Add configuration property to optimize AD/LDAP delta sync search filters

**Description:**
Currently, Ranger Usersync uses an OR conditional query containing both modifyTimeStamp and uSNChanged attributes to detect changes during delta syncs. According to customer feedback, executing a search query with both attributes combined negatively affects Active Directory (AD) performance. At present, this behavior is hardcoded by design and cannot be configured to use only the environment-specific attribute (such as uSNChanged for AD).

**Changes:**
Introduced a new configuration property (e.g., ranger.usersync.ldap.deltasync.servertype) to allow administrators to specify the directory server type and optimize the search query:

**When set to ad:** _The delta sync query uses only the uSNChanged attribute._

**When set to ldap:** _The delta sync query uses only the modifyTimestamp attribute._

**When empty or unset:** _Falls back to the existing OR-combined filter to ensure backward compatibility._

**Impact:**
Improves Active Directory search performance and reduces query overhead by eliminating complex OR conditions and avoiding unsupported attribute queries. There is no behavioral change or impact on existing deployments unless the new property is explicitly configured.

**Testing:**

- _Verified delta sync works correctly against Active Directory when the property is configured as ad._

- _Verified delta sync works correctly against OpenLDAP when the property is configured as ldap._

- _Verified legacy backward compatibility behaves as expected (using the OR conditional) when the property is not configured._

- _Tested the performance by synchronizing ~1.14 million users. With the optimized ad configuration, the sync process completed 1 minute faster compared to the default OR-combined filter._

### DIFF
--- a/ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
+++ b/ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
@@ -122,7 +122,6 @@ public class LdapUserGroupBuilder implements UserGroupSource {
     private String         groupNameAttribute;
     private String         groupCloudIdAttribute;
     private String         currentSyncSource;
-    private String         deltaSyncServerType;
     private String[]       userSearchBase;
     private String[]       groupSearchBase;
     private Set<String>    groupNameSet;
@@ -377,7 +376,6 @@ public class LdapUserGroupBuilder implements UserGroupSource {
         currentSyncSource           = config.getCurrentSyncSource();
         userSearchEnabled           = config.isUserSearchEnabled();
         groupSearchEnabled          = config.isGroupSearchEnabled();
-        deltaSyncServerType         = config.getDeltaSyncServerType();
         ldapUrl                     = config.getLdapUrl();
         ldapBindDn                  = config.getLdapBindDn();
         ldapBindPassword            = config.getLdapBindPassword();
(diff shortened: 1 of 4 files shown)


ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
```diff
@@ -805,6 +806,18 @@ public boolean isGroupSearchFirstEnabled() {
         return groupSearchFirstEnabled;
     }
 
+    public String getDeltaSyncServerType() {
```
spolavarpau1: I don't think we need this method. Instead have this logic in LdapUserGroupBuilder.java

ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
```diff
@@ -513,11 +515,19 @@ private long getUsers(boolean computeDeletes) throws Throwable {
             }
 
             if (config.isDeltaSyncEnabled()) {
-                extendedUserSearchFilter = "(objectclass=" + userObjectClass + ")(|(uSNChanged>=" + deltaSyncUserTime + ")(modifyTimestamp>=" + deltaSyncUserTimeStamp + "Z))";
+                if ("ad".equalsIgnoreCase(deltaSyncServerType)) {
```
spolavarpau1: Add a private method that returns extendedUserSearchFilter value and move the below logic there.

ugsync/src/main/java/org/apache/ranger/unixusersync/config/UserGroupSyncConfig.java
krishnamuttevi: Thanks for suggestion, I have removed the function and added the logic in LdapUserGroupBuilder.java

ugsync/src/main/java/org/apache/ranger/ldapusersync/process/LdapUserGroupBuilder.java
krishnamuttevi: I have updated the logic and added the private method getDeltaSyncFilter()

spolavarpau1 (approved):

PR #1220, opened 2026-09-09
RANGER-5745 - Ranger: Support for wildcard in table name on SQL row f…

rgupta94:
We need a wildcard support for table names in Hive Row Level Filter policies (e.g. *), so one policy can cover multiple tables in a database instead of one policy per table. An upgrade patch updates the Hive service definition on existing Ranger installations to allow this in the Admin UI and enforce it at query time.

### DIFF
--- a/agents-common/src/main/resources/service-defs/ranger-servicedef-hive.json
+++ b/agents-common/src/main/resources/service-defs/ranger-servicedef-hive.json
@@ -194,11 +194,30 @@
   },
   "rowFilterDef": {
     "accessTypes": [
-      { "name": "select" }
+      {
+        "name": "select"
+      }
     ],
     "resources": [
-      { "name": "database", "matcherOptions": { "wildCard": "false" }, "lookupSupported": true, "mandatory": true, "uiHint": "{ \"singleValue\":true }" },
-      { "name": "table",    "matcherOptions": { "wildCard": "true" }, "lookupSupported": true, "mandatory": true, "uiHint": "" }
+      {
+        "name": "database",
+        "matcherOptions": {
+          "wildCard": "false"
+        },
+        "lookupSupported": true,
+        "mandatory": true,
+        "uiHint": "{ \"singleValue\":true }"
+      },
+      {
+        "name": "table",
+        "matcherOptions": {
+          "wildCard": "true"
+        },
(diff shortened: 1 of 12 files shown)


agents-common/src/main/resources/service-defs/ranger-servicedef-hive.json
```diff
@@ -198,7 +198,7 @@
     ],
     "resources": [
       { "name": "database", "matcherOptions": { "wildCard": "false" }, "lookupSupported": true, "mandatory": true, "uiHint": "{ \"singleValue\":true }" },
-      { "name": "table",    "matcherOptions": { "wildCard": "false" }, "lookupSupported": true, "mandatory": true, "uiHint": "{ \"singleValue\":true }" }
+      { "name": "table",    "matcherOptions": { "wildCard": "true" }, "lookupSupported": true, "mandatory": true, "uiHint": "" }
```
rameeshm: @rgupta94  Ranger has a functionality of not to allow insert/update of table which has rowfilter / column masking in it and this will present the users from insert / update of the table in the database. Please check it out. There is a check in isBlockAccessIfRowfilterColumnMaskSpecified in RangerHiveAuthorizer which cause this issue. May be we should document this say that  "applying a wildcard row filter will trigger RangerHivePlugin.blockUpdateIfRowfilterColumnMaskSpecified, effectively blocking all UPDATE/ALTER commands for that user across the entire database. Users should be advised to apply these wildcard policies strictly to Read-Only roles or disable the property in ranger-hive-security.xml." Also add some test cases for this in Hive Plugin. CC @mneethiraj

rgupta94: updated.

agents-common/src/main/resources/service-defs/ranger-servicedef-hive.json
```diff
@@ -194,11 +194,30 @@
   },
   "rowFilterDef": {
     "accessTypes": [
-      { "name": "select" }
+      {
```
mneethiraj, line 197: I suggest retaining existing formatting - for conistency.

rgupta94:
This change is no longer needed,

PR #1221, opened 2026-09-09
RANGER-5790: [Ranger][UI] Atlas resource label not updated in DB during upgrade 'Type Catagory' typo persists

RakeshGuptaDev:

The fix for the typo "Type Catagory" → "Type Category" introduced in https://issues.apache.org/jira/browse/RANGER-4805 (ranger-servicedef-atlas.json) is applied only during fresh installations.

During an upgrade, the Ranger UI continues to display the old typo because resource labels are retrieved from the x_resource_def database table, which retains the pre-upgrade values.

## How was this patch tested?

- Verified that the updated resource labels and descriptions from the service definition JSON are correctly synchronized to the x_resource_def table during an upgrade.
- Verified that the Ranger UI displays the updated "Type Category" label after upgrading an existing Ranger installation.


dev-support/checkstyle-suppressions.xml
```diff
@@ -93,4 +93,5 @@
   <suppress files="PatchForAtlasPolicyUpdateForEntityRead_J10064" checks="TypeName"/>
   <suppress files="PatchForOzoneServiceDefPolicyConditionUpdate_J10066" checks="TypeName"/>
   <suppress files="PatchForOzoneServiceDefAssumeRoleUpdate_J10065" checks="TypeName"/>
+  <suppress files="PatchForAtlasResourceLabelUpdate_J10070" checks="TypeName"/>
```
vyommani: i am using J10070 please use 71.

RakeshGuptaDev: Updated to J10071.

PR #1227, opened 2026-09-11
RANGER-5795: Support pluggable token supplier in Ranger REST client to obtain and refresh bearer tokens dynamically

kumaab:

Introduce new config:
- `ranger.plugin.<service>.policy.rest.client.jwt.supplier` to supply a custom Token Supplier
- defaulting to `DefaultTokenSupplier` if unavailable and `JWT_SOURCE` is configured.

The configured class must implement the JDK's `java.util.function.Supplier<String>`, so custom token suppliers can be plain lambdas/suppliers with no dependency on Ranger libraries. If the property is set, the class is instantiated via reflection.

Refactoring occurrences to generic TokenSupplier and using JWT as one default implementation via `DefaultTokenSupplier`

## How was this patch tested?

- added additional unit tests


nixonrodrigues (approved):

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPluginContext.java
```diff
@@ -196,4 +198,61 @@ void cleanResourceMatchers() {
...
+                    /* prefer a constructor that accepts the Ranger Configuration */
+                    Constructor<?> ctor = clz.getDeclaredConstructor(Configuration.class);
+
+                    ret = (Supplier<String>) ctor.newInstance(config);
+                } catch (NoSuchMethodException excp) {
+                    /* fall back to the no-argument constructor */
+                    ret = (Supplier<String>) clz.getDeclaredConstructor().newInstance();
+                }
+            } else {
+                LOG.error("{}={}: class does not implement {}. Requests may be sent without a bearer token", providerProp, clzName, Supplier.class.getName());
```
mneethiraj: Instead of continuing after this error, I suggest throwing `IllegalArgumentException` - as this is a misconfiguration.

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPluginContext.java
```diff
@@ -196,4 +198,61 @@ void cleanResourceMatchers() {
...
+            }
+        } catch (Throwable excp) {
+            LOG.error("{}={}: failed to instantiate token supplier. Requests may be sent without a bearer token", providerProp, clzName, excp);
```
mneethiraj: Instead of continuing after this error, I suggest throwing `IllegalArgumentException` - as this is a misconfiguration.

agents-common/src/main/java/org/apache/ranger/plugin/policyengine/RangerPluginContext.java
```diff
@@ -196,4 +198,61 @@ void cleanResourceMatchers() {
 
         LOG.debug("<== cleanResourceMatchers()");
     }
+
+    private static Supplier<String> getTokenSupplier(String propertyPrefix, RangerPluginConfig config) {
+        String           providerProp = propertyPrefix + DefaultTokenSupplier.JWT_SUPPLIER;
+        String           clzName      = config.get(providerProp);
+        Supplier<String> ret          = null;
+
+        if (StringUtils.isNotBlank(clzName) && !DefaultTokenSupplier.class.getName().equals(clzName)) {
```
mneethiraj: I suggest replacing 205 to 221 with the following - for simplification:

```
final Supplier<String> ret;

if (StringUtils.isNotBlank(clzName)) {
  LOG.info("Using Token supplier [{}], config: [{}]", clzName, providerProp);

  if (DefaultTokenSupplier.class.getName().equals(clzName)) {
    ret = new DefaultTokenSupplier(propertyPrefix, config);
  } else {
    ret = getCustomTokenSupplier(clzName, config, providerProp);
  }
} else if (isJwtSourceConfigured(propertyPrefix, config)) {
  ret = new DefaultTokenSupplier(propertyPrefix, config);
} else {
  LOG.debug("No token supplier configured, config: [{}]", providerProp);

  ret = null;
}
```

mneethiraj (approved):

PR #1229, opened 2026-09-11
RANGER-5787 : Refactoring the security zone API

dhavalshah9131:
Improvement and refactoring the security zone API

## How was this patch tested?
Successfully validated the Security Zone API.
Build succeeded with unit tests

## Note
Existing security zone which already has set of services including the service type KMS

- Admin user can remove kms service via PUT operation. i.e. update
- Admin user cannot delete the security zone which has the kms service.

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
+++ b/security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
@@ -185,7 +185,7 @@ public class SecurityZoneREST {
         } else {
             securityZone.setId(zoneId);
         }
-
+        blockAdminFromKMSServiceOnUpdate(securityZone, zoneId);
         RangerSecurityZone ret;
 
         try {
@@ -219,6 +219,8 @@ public class SecurityZoneREST {
         try {
             ensureAdminAccess();
 
+            blockAdminFromKMSServiceOnDelete(zoneName, null);
+
             RangerSecurityZoneValidator validator = validatorFactory.getSecurityZoneValidator(svcStore, securityZoneStore);
 
             validator.validate(zoneName, RangerValidator.Action.DELETE);
@@ -250,6 +252,8 @@ public class SecurityZoneREST {
 
         try {
             ensureAdminAccess();
+
+            blockAdminFromKMSServiceOnDelete(null, zoneId);
 
             RangerSecurityZoneValidator validator = validatorFactory.getSecurityZoneValidator(svcStore, securityZoneStore);
 
@@ -706,6 +710,50 @@ public class SecurityZoneREST {
                     }
                 }
             }
+        }
+    }
+
+    private void blockAdminFromKMSServiceOnUpdate(RangerSecurityZone submittedZone, Long zoneId) {
+        if (bizUtil.isAdmin()) {
+            blockAdminFromKMSService(submittedZone);
+
+            if (zoneId != null) {
+                RangerSecurityZone existingZone = null;
+
+                try {
+                    existingZone = securityZoneStore.getSecurityZone(zoneId);
+                } catch (WebApplicationException excp) {
+                    throw excp;
+                } catch (Exception ex) {
+                    LOG.error("Unable to get Security Zone with id : {}", zoneId, ex);
+
+                    throw restErrorUtil.createRESTException(ex.getMessage());
+                }
+
+                blockAdminFromKMSService(existingZone);
+            }
+        }
+    }
+
+    private void blockAdminFromKMSServiceOnDelete(String zoneName, Long zoneId) {
+        if (bizUtil.isAdmin()) {
+            RangerSecurityZone existingZone = null;
+
+            try {
+                if (zoneId != null) {
+                    existingZone = securityZoneStore.getSecurityZone(zoneId);
+                } else if (zoneName != null) {
+                    existingZone = securityZoneStore.getSecurityZoneByName(zoneName);
+                }
+            } catch (WebApplicationException excp) {
+                throw excp;
+            } catch (Exception ex) {
+                LOG.error("Unable to get Security Zone with id : {}, name : {}", zoneId, zoneName, ex);
+
+                throw restErrorUtil.createRESTException(ex.getMessage());
+            }
+
+            blockAdminFromKMSService(existingZone);
         }
     }
 
--- a/security-admin/src/test/java/org/apache/ranger/rest/TestSecurityZoneREST.java
+++ b/security-admin/src/test/java/org/apache/ranger/rest/TestSecurityZoneREST.java
@@ -39,6 +39,7 @@ import org.apache.ranger.plugin.model.RangerSecurityZoneV2;
 import org.apache.ranger.plugin.model.RangerSecurityZoneV2.RangerSecurityZoneChangeRequest;
 import org.apache.ranger.plugin.model.validation.RangerSecurityZoneValidator;
 import org.apache.ranger.plugin.model.validation.RangerValidator;
+import org.apache.ranger.plugin.store.EmbeddedServiceDefsUtil;
 import org.apache.ranger.plugin.store.PList;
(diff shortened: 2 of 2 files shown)


security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
mneethiraj: It is unnecessary to retrieve entire `XXService` and `XXServiceDef` objects just to verify if the service-type is KMS. I suggest optimizing by using a DB query:

jpa_named_queries.xml:
```
	<named-query name="XXService.getServiceType">
		<query>select svcDef.name from XXService obj, XXServiceDef svcDef where obj.name = :svcName and obj.type = svcDef.id</query>
	</named-query>
```

XXServiceDao.java:
```
public String getServiceType(String svcName) {
  try {
    return getEntityManager().createNamedQuery("XXService.getServiceType", String.class)
                    .setParameter("svcName", svcName)
                    .getSingleResult()
  } catch (NoResultException ignored) {
    // ignore
  }

  return null;
}
```

SecurityZoneREST.java:
```
XXServiceDao svcDao = daoManager.getXXService();

for (String serviceName : serviceMap.keySet()) {
  String serviceType = svcDao.getServiceType(serviceName);

  if (EmbeddedServiceDefsUtil.EMBEDDED_SERVICEDEF_KMS_NAME.equals(serviceType)) {
    throw restErrorUtil.createRESTException("Not allowed to update security zone containing KMS service", MessageEnums.OPER_NOT_ALLOWED_FOR_ENTITY);
  }
}
```

dhavalshah9131: @mneethiraj ,
I have updated the PR. 

Thanks

security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
```diff
@@ -695,20 +697,60 @@ private void blockAdminFromKMSService(RangerSecurityZone securityZone) {
...
                 }
             }
         }
     }
 
+    private void blockAdminFromKMSServiceOnUpdate(RangerSecurityZone submittedZone, Long zoneId) {
+        if (bizUtil.isAdmin()) {
+            blockAdminFromKMSService(submittedZone);
+
+            if (zoneId != null) {
```
vyommani, line 714: The existing-zone check here blocks the only way to fix a zone that already contains a KMS service (e.g. one created on an older build before this check existed). Delete is blocked below, and ROLE_KEY_ADMIN can't operate on zones, so an admin who submits this zone with the KMS service removed is still rejected — the zone can't be edited or deleted through the API.

The check also adds nothing: if the submitted zone still contains the KMS service, blockAdminFromKMSService(submittedZone) catches it; if it doesn't, that's the removal we want to allow. Suggest checking only the submitted zone on update, dropping the securityZoneStore.getSecurityZone(zoneId) read, and adding a test that an admin can remove a KMS service from an existing zone (PUT without KMS → DELETE should then succeed).

security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
```diff
@@ -695,20 +697,60 @@ private void blockAdminFromKMSService(RangerSecurityZone securityZone) {
...
-                    XXService xService = daoManager.getXXService().findByName(serviceName);
-
-                    if (xService != null) {
-                        XXServiceDef xServiceDef = daoManager.getXXServiceDef().getById(xService.getType());
+                    String serviceType = daoManager.getXXServiceDef().findServiceDefTypeByServiceName(serviceName);
 
-                        if (EmbeddedServiceDefsUtil.KMS_IMPL_CLASS_NAME.equals(xServiceDef.getImplclassname())) {
-                            throw restErrorUtil.createRESTException("KMS Services/Service-Defs are not accessible for Zone operations", MessageEnums.OPER_NOT_ALLOWED_FOR_ENTITY);
-                        }
+                    if (EmbeddedServiceDefsUtil.EMBEDDED_SERVICEDEF_KMS_NAME.equals(serviceType)) {
```
vyommani, line 702: This changes what "KMS service" means for the gate. Every other sys-admin/key-admin boundary in security-admin keys on implclassname == KMS_IMPL_CLASS_NAME  ServiceREST (getServices filtering, policy CRUD, validateConfig) and RangerBizUtil (the "System Admin cannot create/update/delete KMS Service-Def" check). A key admin can legitimately create a service-def with a different name and the KMS impl class; services under it are key-admin-only everywhere else but would pass this gate for a sys admin.

If the motivation was the two-step lookup / latent NPE on a null xServiceDef, a findServiceDefImplClassByServiceName named query keeps the single round-trip without changing semantics. If name-based is intentional, worth stating in the JIRA since it's a behavior change in a refactor. @mneethiraj  for a second opinion.

security-admin/src/test/java/org/apache/ranger/rest/TestSecurityZoneREST.java
vyommani, line 1: Coverage gaps: (1) the old create-path mocks were removed but there's no test asserting create is still blocked for a KMS service; (2) delete-by-id / by-name where the KMS service sits alongside non-KMS services (current tests use a single-entry map); (3) the V2 RangerSecurityZoneChangeRequest path adding a KMS service; (4) findServiceDefTypeByServiceName returning null for an unknown service name.

security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
vyommani: Walk through it with a concrete scenario.

The scenario. A cluster running Ranger 2.9 has zone Z1, and at some point an admin used PUT /zones/{id} (which had no KMS check) to add dev_kms to it. So the DB now has Z1.services = {dev_hive, dev_kms}. Now the cluster upgrades to a build with this PR.

What happens under the PR's code. The admin realises dev_kms shouldn't be there and tries to remove it: PUT /zones/Z1 with body services = {dev_hive} (no KMS). blockAdminFromKMSServiceOnUpdate runs:

blockAdminFromKMSService(submittedZone) — submitted has only dev_hive, passes.
existingZone = securityZoneStore.getSecurityZone(zoneId) — loads the DB copy, which still has dev_kms.
blockAdminFromKMSService(existingZone) — finds dev_kms, throws "KMS Services/Service-Defs are not accessible for Zone operations".

So the removal is rejected. The admin then tries DELETE /zones/Z1 — blockAdminFromKMSServiceOnDelete loads the same DB copy, finds dev_kms, throws. Nobody else can help: ensureAdminAccess() requires ROLE_SYS_ADMIN, so a key admin can't update or delete zones at all. Z1 is now permanently frozen — can't be edited, can't be deleted — and dev_kms stays zoned, which is exactly the state the fix is meant to prevent. The only way out is editing x_security_zone in the DB by hand.

Why the existing-zone check adds nothing. Consider the two possible updates an admin can send for Z1:

Submitted zone still contains dev_kms → step 1 already throws. Step 3 never runs.
Submitted zone doesn't contain dev_kms → this is the admin removing it. We want this to succeed. Step 3 is the only thing stopping it.

So step 3 never blocks anything step 1 didn't already block; its only effect is to prevent removal.

security-admin/src/main/java/org/apache/ranger/rest/SecurityZoneREST.java
dhavalshah9131: Hi @vyommani ,
I have added the check of implclassname.

security-admin/src/test/java/org/apache/ranger/rest/TestSecurityZoneREST.java
dhavalshah9131: Added the test cases.

mneethiraj (approved):

vyommani (approved):
LGTM

PR #1230, opened 2026-09-11
RANGER-5792: Align to best practices for User rest API

vikaskr22:

Best practice improvement for User rest layer.

## How was this patch tested?

- mvn build passed
- UT passed
- Tested on internal cluster

### DIFF
--- a/security-admin/src/main/java/org/apache/ranger/biz/RangerBizUtil.java
+++ b/security-admin/src/main/java/org/apache/ranger/biz/RangerBizUtil.java
@@ -72,7 +72,6 @@ import javax.servlet.http.HttpServletResponse;
 import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


security-admin/src/main/java/org/apache/ranger/biz/RangerBizUtil.java
```diff
@@ -1061,22 +1060,23 @@ public boolean checkUserAccessible(VXUser vXUser) {
             return true;
         }
 
-        boolean            isAccessible = true;
-        Collection<String> roleList     = userMgr.getRolesByLoginId(vXUser.getName());
+        boolean      isAccessible      = true;
+        List<String> payloadRoles      = StringUtil.transformElements(vXUser.getUserRoleList(), s -> s.trim().toUpperCase());
```
mneethiraj: Instead of creating new lists with upper case role names, consider adding following method and calling it from here:

```
package org.apache.ranger.authorization.utils;

public class StringUtil {
  ...
  public static boolean containsIgnoreCase(Collection<String> values, String str) {
    if (values != null && str != null) {
      return values.stream().anyMatch(s -> str.equalsIgnoreCase(s));
    }

    return false;
  }
  ...
}
```

vikaskr22: Thanks @mneethiraj  for the review. I have incorporated the change and tested on internal cluster and found it working. Please review it one more time. Thanks.

mneethiraj (approved):

